PR #11, opened 2015-06-09
moved monitoring tools into a proper python package namespace structure. updated pylint file to be in sync with openshift-ansible. other style fixes.

twiest:


kwoodson:
:+1:

PR #16, opened 2015-06-16
Add fixed-up Zabbix plotting tools

jgkennedy:
Ready to test
Should pass pylint


openshift_tools/reporting/zabbixplot.py
```diff
@@ -52,41 +69,47 @@ def get_event_data(self):
...
+        # TODO: This will ignore events that start before the data
+        if event_value != "0":
+            # Add downtime, assuming this may be the last event
+            deltatime = current_period_end - event_clock
+            down = True
+        elif event_value == "0" and down:
+            # Remove downtime, compensating for recovered downtime
+            deltatime = current_period_end - event_clock
+            down = False
+        return down, deltatime
```
twiest: "down" is not in scope. You should initialize "down" to a value at the beginning of the method so that it's in scope here.

openshift_tools/reporting/zabbixplot.py
```diff
@@ -52,41 +69,47 @@ def get_event_data(self):
...
+        deltatime = 0
+        # TODO: This will ignore events that start before the data
+        if event_value != "0":
+            # Add downtime, assuming this may be the last event
+            deltatime = current_period_end - event_clock
+            down = True
+        elif event_value == "0" and down:
+            # Remove downtime, compensating for recovered downtime
+            deltatime = current_period_end - event_clock
+            down = False
```
twiest: What should be done in the third case?

You're not handling it here, do you want it to fall through and just return the default values?

If so, then this is fine. I'm just making sure it's doing what you expect.

openshift_tools/reporting/zabbixplot.py
```diff
@@ -52,41 +69,47 @@ def get_event_data(self):
...
-        self.timestamp_start = int(self.time_start.strftime("%s"))
-        self.timestamp_end = int(self.time_end.strftime("%s"))
-        self.mtime_start = mdates.date2num(self.time_start)
-        self.mtime_end = mdates.date2num(self.time_end)
+        self.timespan = ZabbixTimespan(days)
 
-        self.zapi = ZabbixAPI(self.server)
-        self.zapi.login(self.username, self.password)
+        self.zapi = ZabbixAPI(server)
+        self.zapi.login(username, password)
```
twiest: Consider passing in zapi as a param to the ctor instead of constructing this object yourself.

That way, if zapi changes it's way of being initialized, this class won't need to change at all.

This would also reduce the number of arguments passed in.

twiest:
I left a few comments, but overall this is looking very good.

I _really_ like how small your scripts are. That's EXACTLY the pattern we want. Do the heavy lifting in a library, have the scripts to some cli arg parsing and then call the library.

Good Job!

Let me know when you've addressed my comments, and I'll re-review this.

openshift_tools/reporting/zabbixplot.py
jgkennedy: It's in the parameters.

openshift_tools/reporting/zabbixplot.py
jgkennedy: It may be possible for an event to fire that says the trigger is still up. Added an explicit else block to clarify.

jgkennedy:
Feedback issues fixed.

openshift_tools/reporting/zabbixplot.py
```diff
@@ -48,45 +64,52 @@ def get_event_data(self):
...
+            down = True
+        elif event_value == "0" and down:
+            # Remove downtime, compensating for recovered downtime
+            deltatime = current_period_end - event_clock
+            down = False
+        else:
+            # Unchanged status of "up", and we're not recovering from downtime
+            # Last event covered this time, so return a delta of zero
+            deltatime = 0
+        return down, deltatime
```
twiest: It seems that deltatime is now not in scope since it's not being initialized at the top of the function.

Am I wrong?

jgkennedy: It's being initialized in all 3 cases.

openshift_tools/reporting/zabbixplot.py
```diff
@@ -48,45 +64,52 @@ def get_event_data(self):
...
-
-        host_func = lambda event: event["hosts"][0]["hostid"]
+    @staticmethod
+    def _calculate_event_downtime(event_value, event_clock, current_period_end, down):
+        """Calculate the amount of downtime or uptime caused by an event"""
+        # TODO: This will ignore events that start before the data
+        if event_value != "0":
+            # Add downtime, assuming this may be the last event
+            deltatime = current_period_end - event_clock
+            down = True
```
twiest: Consider using early exits for these three cases.

So something like this (untested):
```python
    @staticmethod
    def _calculate_event_downtime(event_value, event_clock, current_period_end, down):
        """Calculate the amount of downtime or uptime caused by an event"""
        # TODO: This will ignore events that start before the data
        if event_value != "0":
            # Add downtime, assuming this may be the last event
            return True, current_period_end - event_clock

        if event_value == "0" and down:
            # Remove downtime, compensating for recovered downtime
            return False, current_period_end - event_clock

        # Unchanged status of "up", and we're not recovering from downtime
        # Last event covered this time, so return a delta of zero
        return down, 0
```

openshift_tools/reporting/zabbixplot.py
```diff
@@ -48,45 +64,52 @@ def get_event_data(self):
...
-        self.time_start = datetime.datetime.now() - datetime.timedelta(days=days)
-        self.time_end = datetime.datetime.now() - datetime.timedelta(days=0)
-        self.timestamp_start = int(self.time_start.strftime("%s"))
-        self.timestamp_end = int(self.time_end.strftime("%s"))
-        self.mtime_start = mdates.date2num(self.time_start)
-        self.mtime_end = mdates.date2num(self.time_end)
-
-        self.zapi = ZabbixAPI(self.server)
-        self.zapi.login(self.username, self.password)
+        self.timespan = ZabbixTimespan(days)
```
twiest: Consider passing in the timespan instead of creating the object here (same as you're doing with zapi).

twiest:
Looks great! :+1:

jgkennedy:
Made a last second fix to the function call. OK to merge?

Removed --num-hosts argument

and
```
-    plot.plot_aggregate(args.periods, args.num_hosts)
+    plot.plot_aggregate(args.periods or args.days)
```

twiest:
@jgkennedy yep, :+1:   go ahead and merge it when you're ready.

PR #40, opened 2015-07-07
added zagg-client docker image

mwoodson:


docker/oso-rhel7-zagg-client/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM oso-rhel7-ops-base:latest
+
+RUN echo "root:redhat" | chpasswd
```
twiest: is this needed?

twiest:
:+1:

PR #50, opened 2015-07-19
Changed oso-rhel7-zagg-client container to use ansible to configure itself on startup. Also changed it to use a vars file that can be populated in a secrets volume. Updated oso-rhel7-ops-base to have cron and setup the ansible config by default.

twiest:


kwoodson:
:+1:

PR #68, opened 2015-07-29
removed ansible zabbix config from zaio. Instead, use the latest from openshift-ansible (this makes it operate like a normal zabbix instance, and ensures we're using the latest configs).

twiest:


kwoodson:
:+1:

PR #71, opened 2015-07-31
Removed htpasswd as we're now generating it on container startup. Changed build.sh scripts to display the time it took to build the image.

twiest:


kwoodson:
:+1:

PR #77, opened 2015-08-07
zbxapi as a library

kwoodson:


openshift_tools/monitoring/zbxapi.py
```diff
@@ -0,0 +1,256 @@
...
+        'Usermacro': ['create', 'createglobal', 'delete', 'deleteglobal', 'get', 'update', 'updateglobal'],
+        'Usermedia': ['get'],
+    }
+
+    def __init__(self, data=None):
+        if not data:
+            data = {}
+        self.server = data.get('server', None)
+        self.username = data.get('username', None)
+        self.password = data.get('password', None)
```
twiest: I'd prefer a NamedTuple called "ZabbixConnection" or something like that to doing a dictionary.

I hate passing in a single dictionary for all params, that makes me have to parse the function to find out what's supposed to be in the dictionary in the first place.

Also, it increases the likelihood of not finding typos.

openshift_tools/monitoring/zbxapi.py
```diff
@@ -0,0 +1,256 @@
...
+        if any([value == None for value in [self.server, self.username, self.password]]):
+            raise ZabbixAPIError('Please specify zabbix server url, username, and password.')
+
+        self.verbose = data.get('verbose', False)
+        self.use_ssl = data.has_key('use_ssl')
```
twiest: I would move these back to being normal params if you change the other to a NamedTuple.

openshift_tools/monitoring/zbxapi.py
```diff
@@ -0,0 +1,256 @@
...
+    setattr(ZabbixAPI, _class_name, ZabbixAPI.meta(_class_name, _method_names))
+
+#if __name__ == '__main__':
+#    data = {
+#        'server': 'http://localhost/zabbix/api_jsonrpc.php',
+#        'username': '',
+#        'password': '',
+#    }
+#    zbx = ZabbixAPI(data)
+#    print zbx.get_content('user', 'get', {})
```
twiest: Can these comments be removed?  Seems like dead code.

Or maybe they should be moved to the top as part of the module doc string to show how it can be used.

twiest:
:+1:

twiest:
:+1:

PR #86, opened 2015-08-14
added default reading to zagg sender

mwoodson:


openshift_tools/monitoring/zagg_sender.py
```diff
@@ -27,22 +27,60 @@
...
+        if not zagg_connection:
+            zagg_connection = self.get_default_zagg_connecton()
+
+        self.host = host
+        self.zaggclient = ZaggClient(zagg_connection=zagg_connection)
+
+    def parse_config(self):
+        """ parse default config file """
+
+        if not self.config:
```
kwoodson: This should probably check whether self.config_file exists as well.  If it does not exist it will throw an exception.

kwoodson:
:+1:

PR #89, opened 2015-08-19
added zagg client check doc

mwoodson:


docs/zagg_client_checks.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+
+Currently there are two different methods that can be used to send data to Zagg.
+
+. *ZaggSender python libraries*:  This is the *PREFERRED* way to write checks. ZaggSender is a python interface used to send data to Zagg
+
+. *ops-zagg-client*:  This is a command line tool that can be used to send data to Zagg.
+
+Both of these clients read their default settings from the config file: `/etc/openshift_tools/zagg_client.yaml`.
+
+More information will be provided below on each of these methods
```
twiest: no need for this line, I'm assuming people won't stop here and say "wtf, where is the rest" :)

docs/zagg_client_checks.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+
+==== ZaggSender - Python API
+
+This is a python API used to create and send data to Zagg.  This is similar to the `zabbix_sender` ruby library used in Openshift V2.
```
twiest: Since this is a public repo, I'd prefer not to reference the ruby library.

docs/zagg_client_checks.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+
+ZaggSender is maintained in the https://github.com/openshift/openshift-tools/tree/master/openshift_tools/monitoring[openshift-tools] repository.  It is installed via the `python-openshift-tools-monitoring` RPM.
+
+ZaggSender reads it's default settings from the config file: `/etc/openshift_tools/zagg_client.yaml`.
+
+The basic bash script would follow the pattern:
```
twiest: I don't see a bash script. Did you mean "basic python script" ?

docs/zagg_client_checks.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+#!/usr/bin/env python
+
+import urllib
+import yaml
+from openshift_tools.monitoring.zagg_sender import ZaggSender
+
+resp = urllib.urlopen('http://www.google.com')
+google_response = resp.getcode()
+
+zs = ZaggSender()
```
twiest: Does this now default to grabbing the creds?

I know you say that above, I just didn't know this change was in.

I think I prefer the ctor as it was before, and adding a static method to create the object from the config file. 

Something like this:
```
zs = ZaggSender.from_config()
```

But anyway, if this is already done, then I guess we can just go with it.

docs/zagg_client_checks.asciidoc
```diff
@@ -0,0 +1,117 @@
...
+#!/bin/bash
+
+<do checks and gather data>
+
+ops-zagg-client -k sample_key1 -o sample_value1
+ops-zagg-client -k sample_key2 -o sample_value2
+----
+
+Here is an example script of capturing the response code of google.com and sending it to Zagg as Zabbix item "web.google.status":
+.ops-zagg-client example script
```
twiest: formatting problem here, doesn't display like I think you intend it to.

twiest:
:+1:

PR #90, opened 2015-08-19
added descriptions of the containers, removed oso-rhel7-mysql as we don't use it.

twiest:


mwoodson:
:+1: seems reasonable

PR #91, opened 2015-08-20
Changed oso-rhel7-zagg-client to set it's hostname in the config file based off of an environment variable.

twiest:


mwoodson:
:+1:

PR #102, opened 2015-08-31
added cron-send-docker-metrics.py

twiest:


docker/oso-rhel7-zagg-client/monitoring_scripts/cron-send-docker-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+    except DockerException as ex:
+        print "\nERROR talking to docker: %s\n" % ex.message
+        keys = {
+            'docker.ping': 0,  # Docker is down
+        }
+        exit_code = 10
+
+    zs.add_zabbix_keys(keys)
+
+    print "Sending these metrics:"
```
kwoodson: @twiest,

Do we want these to always be verbose?

kwoodson:
:+1:  Not blocking on comments.

docker/oso-rhel7-zagg-client/monitoring_scripts/cron-send-docker-metrics.py
twiest: I debated this, and for this quick hack I determined yes. For a real version, I'll probably add -v :)

PR #112, opened 2015-09-14
fixed oso-rhel7-ops-base build script to work with the latest versions of docker that mount tmpfs on top of /run/secrets

twiest:


kwoodson:
:+1:

PR #119, opened 2015-09-16
added http to https redirect

stenwt:


stenwt:
router needs to get changed, not httpd in the container.

PR #123, opened 2015-09-17
[WIP] flock and timeout cron-send-create-app

mmahut:
Make sure we do not run this check twice as it will fail on both instance resulting in an alert.

In the future we should monitor the execution time of this check.


mmahut:
@mwoodson can you have a look?

mmahut:
ok to test

mmahut:
needs rebasing

PR #129, opened 2015-09-22
Monitor the openshift-node service only on node clients

mmahut:
We do not require monitoring openshift-node on the master hosts.

### DIFF
--- a/docker/oso-rhel7-zagg-client/root/config.yml
+++ b/docker/oso-rhel7-zagg-client/root/config.yml
@@ -34,6 +34,8 @@
     - name: run docker storage space checks every 30 minutes
       minute: "*/30"
       job: /usr/local/bin/cron-send-docker-metrics.py
+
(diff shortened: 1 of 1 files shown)


mmahut:
@twiest Thomas, do you know if we require openshift-node to be running on masters?

twiest:
@mmahut Yes, all masters are nodes as well right now (and probably in the future too as we want our monitoring containers to all run inside of OpenShift).

twiest:
@mmahut just to be clear, this PR removes a vital check from our masters, so I'd prefer it to be closed and not merged in.

Since all masters are nodes as well, I can't think of a check that I'd want to run only on the "pure" nodes that I wouldn't also want run on the masters. So, at least for now, I don't think we need to separate the node crons out into their own thing.

Thanks for pinging me before merging this in.

mmahut:
@twiest Thank you for clarification. Closing this.

PR #133, opened 2015-09-25
[wip] Get accurate count of total user pods in a cluster

BlueShells:
someone please help me check if I missed something ,I can get the through a mock script from the zagg-client ,but after I rebuild the zagg-client ,the cron job has been added ,but the script can't be found in the /usr/bin/ ,so that means the line the openshift-tools-scripts.spec haven't work,I don't know why @twiest


twiest:
@BlueShells That's because the docker build pulls the rpms from the yum repo and an RPM with your script hasn't been built yet.

For now, just test by manually putting the script in place, or sym link.

scripts/monitoring/cron-send-user-running-pod-count.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+
+COUNT=$(oadm manage-node --list-pods --selector=type=compute | grep Running | wc -l)
+echo
+echo "Number of User Running pod account : $COUNT"
```
mmahut: Can we get output from this script only on failure or -v/--verbose is passed?

That way we do not get spammed from cron.

twiest: Please disregard this comment. With ops-runner, we'll handle the output. We're not using cron e-mails to detect failures in v3, we're using zabbix instead.

docker/oso-rhel7-zagg-client/root/config.yml
```diff
@@ -45,7 +45,9 @@
     - name: send openshift-master process count
       minute: "*/2"
       job: /usr/bin/cron-send-process-count '^/usr/bin/openshift start master' openshift.master.process.count
-
+    - name: send cluster user running pod total count
+      minute: "*/2"
+      job: /usr/bin/cron-send-user-running-pod-count.sh
```
twiest: Please use ops runner, like this (untested):
```
      job: ops-runner -n csurpc.openshift.master.user.pod.running.count /usr/bin/cron-send-user-running-pod-count.sh
```

scripts/monitoring/cron-send-user-running-pod-count.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+
+COUNT=$(oadm manage-node --list-pods --selector=type=compute | grep Running | wc -l)
```
twiest: I don't think this is going to work. Based on what we saw with Max's PR yesterday.

I think this needs to be:
```
COUNT=$(KUBECONFIG=/etc/openshift/master/admin.kubeconfig oadm manage-node --list-pods --selector=type=compute | grep Running | wc -l  ; exit ${PIPESTATUS[0]})
```


The KUBECONFIG will make sure we have the correct auth, the PIPESTATUS will make sure we exit with the correct exit code.

scripts/monitoring/cron-send-user-running-pod-count.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
twiest: I suggest you change this to this:
```
#!/bin/bash -e
```

That way any errors will exit the script.

Then, ops-runner can alert us to the failure. Otherwise, you may exit with an exit code of 0, even though the script failed.

scripts/openshift-tools-scripts.spec
```diff
@@ -23,6 +23,7 @@ cp -p monitoring/ops-zagg-client.py %{buildroot}/usr/bin/ops-zagg-client
 cp -p monitoring/ops-zagg-processor.py %{buildroot}/usr/bin/ops-zagg-processor
 cp -p monitoring/ops-zagg-heartbeater.py %{buildroot}/usr/bin/ops-zagg-heartbeater
 cp -p monitoring/cron-send-process-count.sh %{buildroot}/usr/bin/cron-send-process-count
+cp -p monitoring/cron-send-user-running-pod-count.sh %{buildroot}/usr/bin/cron-send-user-running-pod-count.sh
```
twiest: This should be:
```
cp -p monitoring/cron-send-user-running-pod-count.sh %{buildroot}/usr/bin/cron-send-user-running-pod-count
```

docker/oso-rhel7-zagg-client/root/config.yml
```diff
@@ -66,7 +66,9 @@
     - name: send openshift-master pods count
       hour: "*/2"
       job: ops-runner -n cspodc.openshift.master.pod.count /usr/bin/cron-send-pod-count
-
+    - name: send cluster user running pod total count
+      minute: "*/2"
+      job: ops-runner -n csurpc.openshift.master.user.pod.running.count /usr/bin/cron-send-user-running-pod-count.sh
```
twiest: Change to: /usr/bin/cron-send-user-running-pod-count

twiest:
:+1:

PR #141, opened 2015-09-30
PCP Metric sampler

kwoodson:


docker/oso-rhel7-zagg-client/root/config.yml
```diff
@@ -25,6 +26,9 @@
     - name: run pcp checks every 5 minutes
       minute: "*/5"
       job: /usr/bin/ops-zagg-client --send-pcp
+    - name: run pcp sampler every 5 minutes
+      minute: "*/5"
+      job: /usr/local/bin/pcp_sample.py -m kernel.all.cpu.idle,kernel.all.cpu.nice,kernel.all.cpu.steal,kernel.all.cpu.sys,kernel.all.cpu.user,kernel.all.cpu.wait.total
```
twiest: To follow convention with the other senders, pcp_sample.py should be named cron-send-pcp-sampled-metrics or something similar.

scripts/monitoring/pcp_sample.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/python
```
twiest: This should be put in openshift-tools/scripts/monitoring and packaged in an RPM.

The other sends need to be packaged as well, they were just quick and dirty, but we definitely don't want to add to the tech debt here.

twiest: This needs to be added to the spec file so that it gets packaged and installed.

docker/oso-rhel7-zagg-client/root/config.yml
twiest: Once pcp_sample.py is packaged, it shouldn't be in /usr/local/bin, so that'll need to be fixed.

openshift_tools/monitoring/pminfo.py
```diff
@@ -115,6 +118,36 @@ def connect(self):
...
+          start
+          end
+          number of cpus
+          interval
+    '''
+    num_cpu = multiprocessing.cpu_count()
+    jiffy = os.sysconf(os.sysconf_names['SC_CLK_TCK'])
+    return ((end - start) * 100 / (float(jiffy) * interval)) / (num_cpu * 10)
+
+def get_sampled_data(metrics, interval):
```
twiest: I think we probably want a count here as well.

Basically, I want to say "give me 10 samples, at 1 second intervals" and have this function take 10 seconds to give me a list of 10 values.

Basically, if I sample for 60 seconds, I don't just want 2 values over those 60 seconds, I want multiple, so I can then apply an average, median, stddev, etc to it.

scripts/monitoring/pcp_sample.py
```diff
@@ -0,0 +1,71 @@
...
+import sys
+import argparse
+from openshift_tools.monitoring import pminfo
+from openshift_tools.monitoring.zagg_sender import ZaggSender
+
+def parse_args():
+    '''Parse the arguments for this script'''
+    parser = argparse.ArgumentParser(description="Tool to sample pcp metrics")
+    parser.add_argument('-m', '--metrics', action="store",
+                        help="Comma delimited string of metrics to send to zabbix")
```
twiest: I'd prefer this to be able to take multiple -m arguments.

This looks nicer and makes more sense to me than a comma separated list.

Otherwise the comma separated list becomes a wall of text to my eyes.

This is how ansible works with -e, docker works with -v, -p, etc. I think that's a more established pattern.

```
/usr/local/bin/pcp_sample.py -m kernel.all.cpu.idle -m kernel.all.cpu.nice -m kernel.all.cpu.steal -m kernel.all.cpu.sys -m kernel.all.cpu.user -m kernel.all.cpu.wait.total
```

scripts/monitoring/pcp_sample.py
```diff
@@ -0,0 +1,71 @@
...
+    parser.add_argument('-i', '--interval', default=10,
+                        action="store", help="Sample pcp metrics for $i amount of time. Default: 10")
+    parser.add_argument('-d', '--debug', default=False,
+                        action="store_true", help="debug mode")
+    parser.add_argument('-v', '--verbose', default=False,
+                        action="store_true", help="Verbose?")
+    parser.add_argument('-t', '--test', default=False,
+                        action="store_true", help="Run the script but don't send to zabbix")
```
twiest: I'd add a -c or --count so we can say how many samples to take.

openshift_tools/monitoring/pminfo.py
```diff
@@ -115,6 +118,39 @@ def connect(self):
...
+    '''
+    if not metrics:
+        return {}
+
+    sampled_results = []
+    for i in range(count):
+        sampled_results.append(get_metrics(metrics))
+        if i == count-1:
+            break
+        time.sleep(interval)
```
twiest: ok, I'm torn on this, so I definitely want your opinion.

Should interval be the time between checks, or the total time they want sampled.

In other words, if I say I want an interval of 60 seconds and a count of 10, should that take 1 minute or 10 minutes?

I think I'm leaning towards the interval being the entire time, which would make this line simply:
```
time.sleep(interval / count)
```

That seems to make the most sense.

Thoughts?

openshift_tools/monitoring/pminfo.py
```diff
@@ -115,6 +118,39 @@ def connect(self):
...
+
+    # Set the number of cpus for this machine
```
twiest: this comment seems out of place.

scripts/monitoring/cron-send-pcp-sampled-metrics.py
```diff
@@ -0,0 +1,74 @@
...
+import sys
+import argparse
+from openshift_tools.monitoring import pminfo
+from openshift_tools.monitoring.zagg_sender import ZaggSender
+
+def parse_args():
+    '''Parse the arguments for this script'''
+    parser = argparse.ArgumentParser(description="Tool to sample pcp metrics")
+    parser.add_argument('-m', '--metrics', action="store",
+                        help="Comma delimited string of metrics to send to zabbix")
```
twiest: I'd prefer this to be able to take multiple -m arguments.

This looks nicer and makes more sense to me than a comma separated list.

Otherwise the comma separated list becomes a wall of text to my eyes.

This is how ansible works with -e, docker works with -v, -p, etc. I think that's a more established pattern.

```
/usr/local/bin/pcp_sample.py -m kernel.all.cpu.idle -m kernel.all.cpu.nice -m kernel.all.cpu.steal -m kernel.all.cpu.sys -m kernel.all.cpu.user -m kernel.all.cpu.wait.total
```

openshift_tools/monitoring/pminfo.py
kwoodson: @twiest,

I think to stay consistent with sar:
       -i interval
              Select data records at seconds as close as possible to the number specified by the interval parameter.

From the man page:
The  accounting  system, based  on  the  values in the count and interval parameters, writes information the specified number of times spaced at the specified intervals in seconds.

scripts/monitoring/cron-send-pcp-sampled-metrics.py
kwoodson: This has been added.

docker/oso-rhel7-zagg-client/root/config.yml
```diff
@@ -25,6 +26,9 @@
     - name: run pcp checks every 5 minutes
       minute: "*/5"
       job: /usr/bin/ops-zagg-client --send-pcp
+    - name: run pcp sampler every 5 minutes
+      minute: "*/5"
+      job: /usr/bin/cron-send-pcp-sampled-metrics.py -m kernel.all.cpu.idle,kernel.all.cpu.nice,kernel.all.cpu.steal,kernel.all.cpu.sys,kernel.all.cpu.user,kernel.all.cpu.wait.total,kernel.all.cpu.irq.hard,kernel.all.cpu.irq.soft
```
twiest: This should be changed to use multiple -m params.

scripts/monitoring/cron-send-pcp-sampled-metrics.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/python
+'''
+  Sample pcp for cpu statistics over an interval
+
+  Example:
+  ./pcp_sample.py -m 'kernel.all.cpu.idle,kernel.all.cpu.nice,kernel.all.cpu.steal,kernel.all.cpu.sys,\
```
twiest: This example should be changed to use multiple -m params.

scripts/monitoring/cron-send-pcp-sampled-metrics.py
twiest: Ah, ok, the comment should be fixed then.

twiest:
Looks great!  :+1:

PR #144, opened 2015-10-01
Projectcounter

gburges:


scripts/monitoring/cron-send-project-count.sh
```diff
@@ -0,0 +1,13 @@
+!# /bin/bash
+
+#get the number of projects except which are ours
+projects=$(oca get projects | egrep -Ev 'NAME|default|openshift|openshift-infra' | wc -l)
+
+echo
+echo "Number of projects: $projects "
```
mmahut: Please make the script silent unless we run it with -v or it fails - so we do not pollute the cronjob mails.

scripts/monitoring/cron-send-project-count.sh
```diff
@@ -0,0 +1,13 @@
+!# /bin/bash
```
kwoodson: No space between shabang and binary path.

kwoodson:
Need entry for the cron in the config.yml and also need the corresponding item/trigger in openshift-ansible.

4-6 hours for the cron seems sufficient for our needs at the moment.  Trigger should be informational and report if there are no projects.

scripts/monitoring/cron-send-project-count.sh
twiest: In v3, we're not relying on cron e-mails to detect failures in our scripts, we're using ops-runner instead.

So, please have your script output data that makes sense by default, and we'll add things to ops-runner to make it nicer / quieter if we want.

PR #149, opened 2015-10-05
broke heartbeats up into full and quick. Full does a host registration, quick just sends the item data.

twiest:


docker/oso-rhel7-zagg-client/root/config.yml
```diff
@@ -30,9 +30,13 @@
       minute: "*/5"
       job: ops-runner -n cspsm.kernal.all.cpu /usr/bin/cron-send-pcp-sampled-metrics -m kernel.all.cpu.idle -m kernel.all.cpu.nice -m kernel.all.cpu.steal -m kernel.all.cpu.sys -m kernel.all.cpu.user -m kernel.all.cpu.wait.total -m kernel.all.cpu.irq.hard -m kernel.all.cpu.irq.soft
 
-    - name: run register every 15 minutes
-      minute: "*/15"
-      job: ops-runner -n ozc.send.heartbeat /usr/bin/ops-zagg-client --send-heartbeat
+    - name: Do a full heartbeat
+      minute: "10"
```
kwoodson: There a reason we want to do this more often than what we had previously set? 15 min down to 10.  I would think this would happen less often.

kwoodson:
Aside from the comment, since it is not blocking. :+1:

docker/oso-rhel7-zagg-client/root/config.yml
twiest: The 10 means do it 10 minutes after the hour. So this happens once an hour, exactly at 10 minutes after the hour.

PR #150, opened 2015-10-05
Sturpin 2015 09 30 count pods

stenwt:


scripts/openshift-tools-scripts.spec
```diff
@@ -58,6 +59,9 @@ OpenShift Tools Monitoring Scripts
 - added ops-runner. It sends the exit code of the command to zabbix.
   (twiest@redhat.com)
 
+* Thu Oct 01 2015 Sten Turpin <sten@redhat.com> 0.0.14-1
+- Adding a count of running pods per cluster (sten@redhat.com)
```
twiest: These should not be part of this PR, they should be done by tito when we do the rpm build / push.

twiest:
:+1:

PR #152, opened 2015-10-07
Gburges projectcounter

gburges:


gburges:
@kwoodson please check it

docker/oso-rhel7-zagg-client/root/config.yml
```diff
@@ -59,7 +59,9 @@
     - name: send openshift-master process count
       minute: "*/2"
       job: ops-runner -n cspc.openshift.master.process.count /usr/bin/cron-send-process-count '^/usr/bin/openshift start master' openshift.master.process.count
-
+    - name: send openshift-master project count every 4 hours
++     hour: "*/4"
++     job: /usr/local/bin/cron-send-project-count.sh
```
kwoodson: This needs to match the name that is going to be in the rpm spec file.  The name you entered on line 29 of the .spec file is this:
/usr/bin/cron-send-project-count

On a side note, you can also use the ops-runner.
ops-runner -n <key> /usr/bin/cron-send-project-count

docker/oso-rhel7-zagg-client/root/config.yml
```diff
@@ -59,7 +59,9 @@
     - name: send openshift-master process count
       minute: "*/2"
       job: ops-runner -n cspc.openshift.master.process.count /usr/bin/cron-send-process-count '^/usr/bin/openshift start master' openshift.master.process.count
-
+    - name: send openshift-master project count every 4 hours
++     hour: "*/4"
++     job: ops-runner -n /usr/bin/cron-send-project-count
```
kwoodson: This still requires a key.  The -n <key> should be passed and then the script name.

ops-runner -n <key> /usr/bin/cron-send-project-count

PR #163, opened 2015-10-11
projectcount done

gburges:


scripts/monitoring/cron-send-project-count.sh
```diff
@@ -0,0 +1,12 @@
+!#/bin/bash
+
+#get the number of projects except which are ours
+projects=$(oca get projects | egrep -Ev 'NAME|default|openshift|openshift-infra' | wc -l)
```
mmahut: Is this working from the subshell?

Can you error out the return code if this command fails?

scripts/monitoring/cron-send-project-count.sh
```diff
@@ -0,0 +1,12 @@
+!#/bin/bash
```
mmahut: Should be #!.

kwoodson:
:+1:

PR #165, opened 2015-10-12
Continer ID in bash prompt and Zabbix registration name

joelddiaz:
@twiest Change base container bash prompt to show the container ID on the bash prompt.

Update zagg-client ansible playbook to use the container ID when registering to Zabbix. Force include .bashrc on start.sh script to make container ID available.


twiest:
:+1:

PR #167, opened 2015-10-14
[WIP] Adding etc health monitoring for every kubelet

mmahut:
Monitoring etcd via the API pings through the kubelets.
Related https://github.com/openshift/openshift-ansible/pull/693

### DIFF
--- a/scripts/monitoring/cron-send-etcd-status.sh
+++ b/scripts/monitoring/cron-send-etcd-status.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 # setting up the certificate and the key
 CERT=`ls /etc/openshift/node/system*node*.crt`
 KEY=`ls /etc/openshift/node/system*node*.key`
--- a/scripts/monitoring/cron-send-etcd-status.sh
+++ b/scripts/monitoring/cron-send-etcd-status.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# setting up the certificate and the key
+CERT=`ls /etc/openshift/node/system*node*.crt`
+KEY=`ls /etc/openshift/node/system*node*.key`
+
+# querying the health ping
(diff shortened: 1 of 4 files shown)


mmahut:
@kwoodson  @twiest  I need the /etc/openshift/node to be mounted in the monitoring container to access the private key that is needed to reach the API, can I bind mount it? Or is there a nicer way to archive this?

twiest:
@mmahut let's talk offline. I sent you an e-mail.

scripts/monitoring/cron-send-etcd-status.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# setting up the certificate and the key
+CERT=`ls /etc/openshift/node/system*node*.crt`
+KEY=`ls /etc/openshift/node/system*node*.key`
+
+# querying the health ping
+wget -q -O- --no-check-certificate --certificate=${CERT} --private-key=${KEY} https://0.0.0.0:10250/healthz/ping |grep -c "ok" >/dev/null
```
twiest: This seems like it's monitoring the kubelet itself, not etcd, which is fine and needed, but this check says it's to monitor etcd specifically.

Am I missing something?

scripts/monitoring/cron-send-etcd-status.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
twiest: You should do a '#!/bin/bash -e'  so that it exits on error and ops-runner will alert us to the problem. Otherwise this script could fail and still exit with an exit code of 0.

scripts/monitoring/cron-send-etcd-status.sh
mmahut: You are right, it should be port 4001, not 10250. Going to rebase.

twiest: It's more than that, though. You're running this on all of the nodes. And you're binding in the node's cert. You're also hitting 0.0.0.0, which is "all local interfaces."

That's why I said this seems to be monitoring the kubelet and not etcd.

PR #173, opened 2015-10-29
Add pcp/pmcd status check.

joelddiaz:
@twiest @kwoodson Will kill the main "pcp start" process which will cause the container to exit.

### DIFF
--- a/docker/oso-f22-host-monitoring/Dockerfile
+++ b/docker/oso-f22-host-monitoring/Dockerfile
@@ -38,6 +38,11 @@ ADD pmlogger_control /etc/pcp/pmlogger/control
 ADD pmdas/ /var/lib/pcp/pmdas/
 
 #
+# script to watch health of pmcd
+ADD check-pmcd-status.sh /usr/local/bin/check-pmcd-status.sh
+ADD container-start.sh /usr/local/bin/start.sh
+
+#
 # Expose pmcd's main port on the host interfaces.
 EXPOSE 44321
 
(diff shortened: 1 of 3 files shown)


docker/oso-f22-host-monitoring/Dockerfile
```diff
@@ -38,6 +38,11 @@ ADD pmlogger_control /etc/pcp/pmlogger/control
 ADD pmdas/ /var/lib/pcp/pmdas/
 
 #
+# script to watch health of pmcd
+ADD check-pmcd-status.sh /usr/local/bin/check-pmcd-status.sh
+RUN chmod +x /usr/local/bin/check-pmcd-status.sh
```
twiest: This shouldn't be needed as long as it's executable in the filesystem at least AFAIK.

docker/oso-f22-host-monitoring/Dockerfile
```diff
@@ -59,4 +64,4 @@ LABEL RUN docker run -d --privileged --net=host --pid=host --ipc=host -v /sys:/s
 # and can be overridden by passing additional parameters to docker run.
 ENV PATH /usr/share/pcp/lib:/usr/libexec/pcp/bin:$PATH
 #ENTRYPOINT ["pmcd"]
-CMD ["/usr/share/pcp/lib/pcp", "start"]
+CMD ["/bin/sh", "-c", "/usr/local/bin/check-pmcd-status.sh & /usr/share/pcp/lib/pcp start"]
```
twiest: I'd prefer having a start.sh script. Any reason not to have one?

joelddiaz:
@twiest latest version removes full paths in scripts

twiest:
:+1:

PR #174, opened 2015-11-02
etcd status monitoring

mmahut:
etcd status check retrieves the store status from etcd API and collect interesting items which can be plotted into graphs.

We need to bind mount the etcd client certs to access the API.

### DIFF
--- a/scripts/monitoring/cron-send-etcd-status.py
+++ b/scripts/monitoring/cron-send-etcd-status.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+'''
+   Command to send status of etcd to zabbix
+'''
+
+#This is not a module, but pylint thinks it is.  This is a command.
+#pylint: disable=invalid-name
+
+from openshift_tools.monitoring.zagg_sender import ZaggSender
+import json
+import requests
+
+def main():
+    ''' Get data from etcd API
+    '''
+
+    API_HOST = 'https://localhost:4001/v2/stats/store'
+    SSL_CLIENT_CERT = '/etc/openshift/master/master.etcd-client.crt'
+    SSL_CLIENT_KEY = '/etc/openshift/master/master.etcd-client.key'
+    zs = ZaggSender()
+
+    # Fetch the store statics from API
+    try:
+        request = requests.get(API_HOST, cert=(SSL_CLIENT_CERT, SSL_CLIENT_KEY), verify=False)
+        content = json.loads(request.content)
+        etcd_ping = 1
+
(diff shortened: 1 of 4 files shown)


mmahut:
Tested locally and all appear to be working fine.

@twiest ready for review, please take a look.

twiest:
@mmahut please rebase onto the latest master.

scripts/monitoring/cron-send-etcd-status.py
```diff
@@ -0,0 +1,43 @@
...
+def main():
+    ''' Get data from etcd API
+    '''
+
+    API_HOST = 'https://localhost:4001/v2/stats/store'
+    SSL_CLIENT_CERT = '/etc/openshift/master/master.etcd-client.crt'
+    SSL_CLIENT_KEY = '/etc/openshift/master/master.etcd-client.key'
+
+    # Fetch the store statics from API
+    request = requests.get(API_HOST, cert=(SSL_CLIENT_CERT, SSL_CLIENT_KEY), verify=False)
```
twiest: I would suggest adding another zabbix item called "openshift.master.etcd.ping" and when you're able to ping etcd, then send a 1. When you're not able to ping etcd, send a 0.

Currently, what happens here when etcd is down?

I would also place this in a try / catch, just to make sure.

That's the kind of logic that I do in the docker info gathering script.

See here:
https://github.com/openshift/openshift-tools/blob/master/docker/oso-rhel7-zagg-client/monitoring_scripts/cron-send-docker-metrics.py#L156-L180

twiest:
@mmahut good job on this, it's looking very good.

scripts/monitoring/cron-send-etcd-status.py
mmahut: Right now the ops-runner will fail - but you are right, I will catch that exception and send it as an item, so we can alert high on it if it really fails.

twiest: Yeah, so all ops-runner failures are considered bugs in the script. ops-runner is just there to alert us to the bugs. We never want to rely on ops-runner for valid failure cases.

mmahut: Fixed and rebased.
```
[CTR][root@e02d0d932ad6 ~]$ /usr/bin/cron-send-etcd-status
ERROR talking to etcd API: ('Connection aborted.', error(111, 'Connection refused'))
```

mmahut:
@twiest PTAL

twiest:
@mmahut Looks great!  :+1:

PR #175, opened 2015-11-02
add scripts to check and report pcp ping state

joelddiaz:
and update zagg-client playbooks to add cron job for checking pcp ping state


twiest:
:+1:

PR #176, opened 2015-11-03
Make sure cron-send-project-count.sh is executable

mmahut:
@gburges all is fine, just one last detail - the script needs to be executable. Can you review this change?


mmahut:
Additionally, /sbin/bash doesn't exist.

gburges:
+1

PR #179, opened 2015-11-03
Run etcd check every two minutes

mmahut:
Adjusting the cron runs. I made a typo initially, put hour instead of minutes.


mwoodson:
:+1:

PR #183, opened 2015-11-06
Load the etcd port as it can be different when using external etcd

mmahut:
When not using the build-in etcd component, the port and the url of the etcd api may be different. We need to load it from the configuration file to be sure.


mmahut:
@twiest one scenario I did not account for is when we are running both build-in and external etcd in an environment. The following change will load the correct url from the configuration of OpenShift's master. Can you review it, please?

zhiwliu:
:+1:

gburges:
+1

PR #185, opened 2015-11-11
Add scripts to report S3 bucket stats from master nodes

joelddiaz:


docker/oso-rhel7-zagg-client/Dockerfile
```diff
@@ -16,6 +16,8 @@ RUN yum clean metadata && \
     yum -y update && \
     yum clean all
 
+RUN pip install awscli
```
twiest: use the python-awscli rpm.

joelddiaz:
@twiest verified python-awscli from li.repo works with cron-send-s3-metrics.py. Changed to pull python-awscli during yum installation instead of via pip.

scripts/monitoring/cron-send-s3-metrics.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python
+'''
+  Collect and send AWS S3 bucket stats
+'''
+# vim: expandtab:tabstop=4:shiftwidth=4
```
twiest: I believe the vim stuff has to be within the first 5 lines, so having it exactly at line 5 is a bit scary. What if, for instance, you decide to make the module doc string longer and that pushes this down a few lines?

Please consider moving this line above the doc string comment.

joelddiaz: Addressed in current PR

joelddiaz:
@twiest pylint looks to be satisfied now. AWS and OC parts are pulled out into their own classes.

openshift_tools/monitoring/awsutil.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self.access_key = access_key
+        self.secret_access_key = secret_access_key
+        self.verbose = verbose
+
+    def run_cmd(self, cmd, verbose=False):
+        ''' Actually call out aws tool (while keeping AWS creds minimally exposed) '''
+
+        os.environ["AWS_ACCESS_KEY_ID"] = self.access_key
+        os.environ["AWS_SECRET_ACCESS_KEY"] = self.secret_access_key
```
twiest: I'm not a big fan of stomping on this processes environment so that the child will have the values.

Consider using the env param to check_output instead.

Something like:
```
results = subprocess.check_output(cmd, env={'AWS_ACCESS_KEY_ID': self.access_key, 'AWS_SECRET_ACCESS_KEY': self.secret_access_key})
```

openshift_tools/monitoring/awsutil.py
```diff
@@ -0,0 +1,103 @@
...
+import subprocess
+
+class AWSUtil(object):
+    ''' Routines for interacting with awscli '''
+
+    def __init__(self, access_key, secret_access_key, verbose=False):
+        ''' Save auth details for later usage '''
+
+        self.access_key = access_key
+        self.secret_access_key = secret_access_key
```
twiest: I think we should default to using the environment variables if the client of this library didn't pass in these values.

That seems to be the pattern that libraries that work with AWS use.

openshift_tools/monitoring/awsutil.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if verbose or self.verbose:
+            print "Running command: " + str(cmd)
+        try:
+            results = subprocess.check_output(cmd)
+        except OSError:
+            raise Exception("Error running command: {}".format(str(cmd)))
+
+        os.environ["AWS_ACCESS_KEY_ID"] = ""
+        os.environ["AWS_SECRET_ACCESS_KEY"] = ""
```
twiest: This shouldn't be needed if we pass the env into subprocess.

openshift_tools/monitoring/ocutil.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python
+'''
+  Interface to OpenShift oc command
+'''
+# vim: expandtab:tabstop=4:shiftwidth=4
```
twiest: move to line 2

openshift_tools/monitoring/awsutil.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python
+'''
+    Interface to awscli
+'''
+# vim: expandtab:tabstop=4:shiftwidth=4
```
twiest: move to line 2

openshift_tools/monitoring/ocutil.py
```diff
@@ -0,0 +1,68 @@
...
+
+        if namespace == None:
+            self.namespace = 'default'
+        else:
+            self.namespace = namespace
+
+        self.verbose = verbose
+
+        # Mimic what the 'oca' command does
+        os.environ["KUBECONFIG"] = "/etc/openshift/master/admin.kubeconfig"
```
twiest: Same as above, pass this into the env of the subprocess.

joelddiaz:
@twiest re-tamed pylint. Changed the way environment variables are handled (now defaults to using pre-existing environment if none passed in). Also expose AWS creds only when running the 'aws' tool.

openshift_tools/monitoring/ocutil.py
```diff
@@ -0,0 +1,67 @@
...
+        '''
+
+        if namespace == None:
+            self.namespace = 'default'
+        else:
+            self.namespace = namespace
+
+        self.verbose = verbose
+
+    def run_cmd(self, cmd, verbose=False):
```
twiest: This is private, please rename to _run_cmd.

Consider removing verbose.

openshift_tools/monitoring/awsutil.py
```diff
@@ -0,0 +1,109 @@
...
+            self.access_key = access_key
+
+        if secret_access_key == None:
+            self.secret_access_key = os.environ["AWS_SECRET_ACCESS_KEY"]
+        else:
+            self.secret_access_key = secret_access_key
+
+        self.verbose = verbose
+
+    def run_cmd(self, cmd, verbose=False):
```
twiest: This is private, please rename to _run_cmd.

Consider removing verbose.

scripts/monitoring/cron-send-s3-metrics.py
```diff
@@ -0,0 +1,96 @@
...
```
twiest: move to line 2

joelddiaz:
@twiest privatized to _run_cmd(), and moved the vim settings upwards

twiest:
:+1:

PR #187, opened 2015-11-13
Docker cron timing

joelddiaz:
create dokerutil library
bring cron-send-docker-metrics.py into openshift-tools-scripts
removing timing function from cron-send-docker-metrics.py
create cron-send-docker-timer.py to provide timing info


openshift_tools/timeout.py
```diff
@@ -1,50 +1,77 @@
...
+#         pylint does not like the name 'timeout'
+# Status: permanently disabled
+# pylint: disable=invalid-name
+class timeout(object):
+    ''' anonymous block for setting arbritrary timeouts '''
+
+    # Reason: disable pylint too-few-public-methods
+    # Status: permanently disabled
+    # pylint: disable=too-few-public-methods
+    def __init__(self, seconds=1, error_message='Timeout'):
```
twiest: Why default seconds?

I think they should be forced to pass that in.

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -0,0 +1,53 @@
...
+from openshift_tools.monitoring.dockerutil import DockerUtil
+import json
+
+if __name__ == "__main__":
+    keys = None
+    zs = ZaggSender()
+    try:
+        cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
+        du = DockerUtil(cli)
+        du_dds = du.get_disk_usage()
```
twiest: Does this still timeout?

I don't see where it's timing out now.

Should we use a "with timeout(15):" here?

twiest:
:+1:

twiest:
:+1:

PR #206, opened 2015-12-01
making ugly quick fix to work around /run issue

mwoodson:


mwoodson:
@twiest @kwoodson @joelddiaz  watched and approved this

PR #207, opened 2015-12-02
Fix order for when pathname is generated.

joelddiaz:
Need to wait until zagg-server-vars is loaded before building file path string


kwoodson:
:+1:

PR #221, opened 2015-12-11
Changed ops-zagg-processor to send metrics on the number of metrics in the queue, heart beats in the queue and the number of errors while processing.

twiest:


kwoodson:
:+1:

PR #225, opened 2015-12-14
Added iputils to ops-base. Added the ability to pause a docker build if an env var is set for all of our containers.

twiest:


joelddiaz:
:+1:

PR #227, opened 2015-12-15
openshift-tools-script-monitoring creates /etc/openshift_tools

joelddiaz:
make the mkdir more tollerant of this


kwoodson:
:+1:

PR #229, opened 2015-12-16
updated the config for zagg client to accomodate ansible-tower

mwoodson:


docker/oso-rhel7-zagg-client/root/ansible-tower_default_vars
```diff
@@ -0,0 +1,7 @@
+---
+# These are defaults variables, override using zagg_client_vars.yml vars file
```
twiest: I don't like the mixing of - and _ in the filename here. Also, this needs to have .yml as the extension.

So, I recommend ansible_tower_default_vars.yml

docker/oso-rhel7-zagg-client/root/ansible-tower_default_vars
```diff
@@ -0,0 +1,7 @@
+---
```
twiest: This file needs to have a .yml extension.

kwoodson:
:+1:

PR #236, opened 2016-01-04
raise timeout when doing DNS tests

joelddiaz:
have seen getent take up to 80 sec to complete
raise timeout to allow it to finish


twiest:
:+1:

PR #237, opened 2016-01-06
merged zagg client and pcp into oso-rhel7-host-monitoring container

mwoodson:


mwoodson:
This failed because we are checking in vendor libraries that are a clone from another repo.

kwoodson:
:+1:

PR #245, opened 2016-01-19
log non-zero exits/error messages to /var/log/ops-runner.log

joelddiaz:


scripts/monitoring/ops-runner.sh
```diff
@@ -20,8 +20,21 @@ function print_usage() {
   echo
 }
 
+function set_date() {
+  # yyyy-mm-dd hh:mm:ss
+  DATE=`date +"%Y-%m-%d %T " | tr -d "\n"`
+}
```
twiest: Why have a whole function for this? You only use it in one place.

scripts/monitoring/ops-runner.sh
```diff
@@ -20,8 +20,21 @@ function print_usage() {
...
+
+function log() {
+  set_date
+  # Replace newlines with '~' so we can do a single-line log entry
+  NO_NEWLINES=$(echo "$1" | tr '\n' '~')
+  printf "%s Output: %s While running: %s\n" "$DATE" "$NO_NEWLINES" "$FULL_COMMAND" >> /var/log/ops-runner.log
```
twiest: Use syslog format for date, hostname, program name

scripts/monitoring/ops-runner.sh
```diff
@@ -52,6 +67,7 @@ while getopts ":n:c:s:h" opt; do
       ;;
     \?)
       echo "Invalid option: -$OPTARG" >&2
+      log "Invalid option: -$OPTARG"
```
twiest: Don't log on invalid input.

scripts/monitoring/ops-runner.sh
```diff
@@ -20,8 +20,21 @@ function print_usage() {
...
+  set_date
+  # Replace newlines with '~' so we can do a single-line log entry
+  NO_NEWLINES=$(echo "$1" | tr '\n' '~')
+  printf "%s Output: %s While running: %s\n" "$DATE" "$NO_NEWLINES" "$FULL_COMMAND" >> /var/log/ops-runner.log
+}
+
 function die() {
-  echo "$1"
+  echo "$1" >&2
+  log "$1"
```
twiest: Not needed here.

scripts/monitoring/ops-runner.sh
```diff
@@ -74,19 +90,28 @@ if [ ! -z "$SLEEP" ] ; then
    sleep $(( $RANDOM % $SLEEP  + 1 ))s
 fi
 
+TEMP_FILE=$(mktemp)
```
twiest: Always to a prefix so that we know what created the temp file.

scripts/monitoring/ops-runner.sh
```diff
@@ -74,19 +90,28 @@ if [ ! -z "$SLEEP" ] ; then
...
-  "$@"
-  EXITCODE=$?
+  "$@" | tee $TEMP_FILE
+  EXITCODE=${PIPESTATUS[0]}
 fi
 set -e
 
+if [ "$EXITCODE" -ne "0" ] ; then
+  log "`cat $TEMP_FILE`"
+  log "Exit code: ${EXITCODE}"
```
twiest: Combine into a single log entry.

scripts/monitoring/ops-runner.sh
```diff
@@ -74,19 +90,28 @@ if [ ! -z "$SLEEP" ] ; then
...
+fi
+rm $TEMP_FILE
```
twiest: We need to trap EXIT to clean this up.

scripts/monitoring/ops-runner.sh
```diff
@@ -74,19 +90,28 @@ if [ ! -z "$SLEEP" ] ; then
...
+
+echo "Exit code: ${EXITCODE}"
+
 # Create our dynamic item for this command
 ops-zagg-client --discovery-key disc.ops.runner --macro-string '#OSO_COMMAND' --macro-names "$NAME"
```
twiest: We should exit with the same exit code as the program we called.

```exit $EXITCODE```

**Comment by kwoodson** [comment]
ok to test

File: scripts/monitoring/ops-runner.sh [hunk 9]
```diff
@@ -31,9 +47,11 @@ if [ $# -eq 0 -o "$1" == "--help" ] ; then
   exit
 fi
 
+FULL_COMMAND="$0 $*"
```
twiest: This isn't used.

scripts/monitoring/ops-runner.sh
```diff
@@ -31,9 +47,11 @@ if [ $# -eq 0 -o "$1" == "--help" ] ; then
...
 NAME=""
 COMMAND=""
 SLEEP=""
+TEMP_FILE=""
```
twiest: Why does this need to be here?

You set the value a few lines down, and you don't reference it in between the two.

scripts/monitoring/ops-runner.sh
```diff
@@ -74,21 +92,31 @@ if [ ! -z "$SLEEP" ] ; then
    sleep $(( $RANDOM % $SLEEP  + 1 ))s
 fi
 
+TEMP_FILE=$(mktemp ops-runner-${NAME}-XXXXXXXXXX)
```
twiest: I would move this to above the cleanup function so that it's always in scope (in case an exit comes before this is defined, the exit won't error).

twiest:
:+1:

PR #263, opened 2016-01-27
add script to detect and remove stray OVS rules/flows

joelddiaz:


scripts/monitoring/cron-fix-ovs-rules.py
```diff
@@ -0,0 +1,107 @@
...
+        except (OSError, CalledProcessError):
+            print "ovs-ofctl not found or misbehaving"
+            exit(1)
+
+        self.ports = None
+        self.rules = None
+
+    def get_port_list(self):
+        ''' Returns of active/good ports '''
+        if self.ports == None:
```
twiest: Consider doing early exit: 

```python
if self.ports:
   return self.ports
```

scripts/monitoring/cron-fix-ovs-rules.py
```diff
@@ -0,0 +1,107 @@
...
+                    port_list.append(match.group(1))
+            self.ports = port_list
+
+        return self.ports
+
+    def get_rule_list(self):
+        ''' Returns list of active rules we would care about. We care
+            about the case where the assigned cookie is equal to the
+            output port '''
+        if self.rules == None:
```
twiest: Consider doing early exit

scripts/openshift-tools-scripts.spec
```diff
@@ -53,6 +54,20 @@ mkdir -p %{buildroot}/var/run/zagg/data
 # ----------------------------------------------------------------------------------
 # openshift-tools-scripts-monitoring-pcp subpackage
```
twiest: fix name

twiest:
:+1:

twiest:
:+1:

PR #275, opened 2016-02-01
report stray OVS rules found before and after cleanup

joelddiaz:


scripts/monitoring/cron-fix-ovs-rules.py
```diff
@@ -93,20 +98,41 @@ def remove_rules(self, rule_list):
...
 
-if __name__ == "__main__":
-    ovs_fixer = OVS()
-
-    # Dev says rules before ports since OpenShift will set up ports, then rules
-    rules = ovs_fixer.get_rule_list()
-    ports = ovs_fixer.get_port_list()
+ZBX_KEY = "openshift.node.ovs.stray.rules"
 
+def find_bad_rules(ports, rules):
```
twiest: Why isn't this part of the class?

twiest:
:+1:

PR #276, opened 2016-02-01
add cron-ovs-fix to cronjob

joelddiaz:


docker/oso-rhel7-host-monitoring/root/config.yml
```diff
@@ -91,6 +91,10 @@
       minute: "0"
       job: ops-runner -n csos.openshift.master.ovs.status /usr/bin/cron-send-ovs-status
 
+    - name: fix and report on stray OVS rules
+      minute: "*/5"
+      job: ops-runner -n cfor.openshift.node.ovs.stray.rules /usr/bin/cron-fix-ovs-rules
```
twiest: I think we want to flock on just about every cron job.

Consider adding a -f to ops-runner.

twiest:
:+1:

PR #281, opened 2016-02-03
initial import of remote healing tool

joelddiaz:


joelddiaz:
@joelsmith : Script has changed a bit since last time we spoke about this. General flow is still going to be a bunch of 'if(trigger): action' statements. Go ahead and give this a security review.

joelsmith:
The key that will be used to run this script must be configured with a command="/path/to/the/script" option. With that force command in place, arguments will not be passed to the script as parameters. Instead, no matter what command (+ args) the user tries to run with the key, sshd will run the script (with no args). Because of this, the script should not parse arguments (that aren't there) but should instead look at and safely parse the SSH_ORIGINAL_COMMAND environment variable.

See https://github.com/openshift/puppet-openshift-online/blob/master/modules/oo_openshift/files/node/utils/verify_rhc-move_rsync_command for an example of such a script.

joelsmith:
Sorry, there's another option, and perhaps it's what you had in mind. You can keep the script as is, but have a validation script wrapper that runs as the force command. It could do the validation and parsing of SSH_ORIGINAL_COMMAND and then invoke this script as written. That has the advantages of making this script easy to use for a human. It has the disadvantage of having to keep another script in sync.

If you want to keep this script easy to use for humans and have it do the SSH_ORIGINAL_COMMAND validation, perhaps the force command could be "path/to/script --ssh-validate", and if the script sees that the ssh-validate option is present, then it parses SSH_ORIGINAL_COMMAND safely and sets the arg options as if they had been passed in on the command line.

PR #284, opened 2016-02-05
2nd attempt remote healing tool

joelddiaz:
No longer depend on sudo. Will now work with authorized_keys forcing the command with the command line parameters being made available through SSH_ORIGINAL_COMMAND and when run manually. Add strict hostname validation.

### DIFF
--- a/scripts/remote-heal/remote-healer.py
+++ b/scripts/remote-heal/remote-healer.py
@@ -1,0 +1,170 @@
+#!/usr/bin/python
+# vim: expandtab:tabstop=4:shiftwidth=4
+'''
+    Tool to process and take action on incoming zabbix triggers.
+'''
+# Disabling invalid-name because pylint doesn't like the naming conention we have.
+# pylint: disable=invalid-name
+
+import argparse
+import ConfigParser
+import logging
+import os
+import re
+import shlex
+import subprocess
+import sys
+
+class RemoteHealer(object):
+    ''' Class to process zabbix triggers and take appropriate actions '''
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


joelddiaz:
@joelsmith here's a version with the added hostname validation and accepting regular command line params and SSH_ORIGINAL_COMMAND

scripts/remote-heal/remote-healer.py
```diff
@@ -0,0 +1,162 @@
...
+                                         'from command line or ' +
+                                         'SSH_ORIGINAL_COMMAND and take ' +
+                                         'appropriate healing actions')
+        parser.add_argument("--host", required=True)
+        parser.add_argument("--trigger", required=True)
+        parser.add_argument("--trigger-val", required=True)
+        parser.add_argument("--verbose", action="store_true", help='Print to stdout')
+        parser.add_argument("--debug", action="store_true", help='Log more details')
+
+        if os.getenv("SSH_ORIGINAL_COMMAND"):
```
joelsmith: I'm not a big fan of this approach since I can't do "ssh root@tower 'do-this --host foo; do-that --host bar; remote-healer --host baz". Since SSH_ORIGINAL_COMMAND would be set, it would go down the env var parser code path, and it would surprise me when I had my heal action run on host foo instead of host baz.  I prefer to have the force command use an explicit option that tells the script to look at the SSH_ORIGINAL_COMMAND, and have it otherwise use normal argv parsing. That being said, it's not really a security thing, and my example is pretty contrived. I just don't think it would be much fun to figure out why things weren't working in a situation like that.

joelsmith:
:+1:

PR #289, opened 2016-02-08
create /etc/origin softlink if necessary

joelddiaz:


kwoodson:
:+1:

joelddiaz:
@stenwt this should fix the missing /etc/origin directories

PR #295, opened 2016-02-09
force create /etc/origin symlink

joelddiaz:
Startup playbook checks for existence of /etc/origin, and then creates a softlink if it isn't there.
Ansible errors out if you create a soft link to a non-existent file.
There is a non-existent /etc/openshift when running host-monitoring locally.
Use 'force: yes' to let Ansible create a soflink to a non-existent file w/o erroring.


kwoodson:
lgtm :+1:

PR #306, opened 2016-02-18
deal with secure and non-secure registries

joelddiaz:


scripts/monitoring/cron-send-registry-checks.py
```diff
@@ -128,6 +129,10 @@ def healthy_registry(url):
             https://github.com/docker/distribution/blob/master/docs/deploying.md#running-a-domain-registry
         '''
 
+        url = '{}://{}:{}/'.format(self.docker_protocol, ip_addr, port)
+        if not secure:
+            url = re.sub('^https', 'http', url)
```
twiest: Consider not using a regex here.

```python
proto = self.docker_protocol
if not secure:
    proto = 'http'

url = '{}://{}:{}/'.format(proto, ip_addr, port)
```

twiest:
:+1:

PR #317, opened 2016-02-29
add retry logic and set zagg_client to 2 retries

joelddiaz:
add custom patch to deal with exceptions in the exception path of urllib3


kwoodson:
:+1:  i like this!

PR #326, opened 2016-03-07
python-openshift-tools-ansible needs to pull in openshift-ansible-zabbix

joelddiaz:
this allows removal of the embedded libzabbix in the Dockerfile for zagg-web container


kwoodson:
lgtm :+1:

PR #329, opened 2016-03-10
use base python from RHEL for SNI support

joelddiaz:
starting with python-2.7.5-34.el7.src.rpm (https://rhn.redhat.com/errata/RHSA-2015-2101.html) RHEL's python has native SNI support


kwoodson:
:+1:

PR #337, opened 2016-03-11
zagg-server doesn't need monitoring-openshift

joelddiaz:
this dependency is pulling in atomic-openshift into zagg-web (where it isn't needed/used)


kwoodson:
:+1:

PR #341, opened 2016-03-15
change deps to binary path requirements

joelddiaz:
need this because upstream origin RPM package names (origin and origin-clients) are different than enterprise RPM names (atomic-openshift and atomic-openshift-clients)


kwoodson:
:+1:

PR #342, opened 2016-03-15
Initial checkin of vendored OpenShift v3 installer.

twiest:


mwoodson:
:+1:

twiest:
The bot is failing because of pylint checks on files in the installer. I'm not going to change them because we are consuming them as is.

PR #347, opened 2016-03-16
initial centos7 container for zabbix-server

joelddiaz:
main change from previous container building:
1) set a default hostname that is clearly the result of not setting HOSTNAME
2) create a dummy/default RSA key


kwoodson:
:+1:

PR #365, opened 2016-03-23
[wip]add two item pv.capacity total and available capacity and dynamic pv count of each type

BlueShells:
add two item pv.capacity total and available capacity and dynamic pv count of each type


kwoodson:
:+1:

PR #383, opened 2016-04-07
Moved oadm_router to tools

kwoodson:


kwoodson:
@joelddiaz :+1: in #382

PR #384, opened 2016-04-08
initial spec for building libzabbix RPM out of openshift-tools

joelddiaz:


joelddiaz:
@twiest @kwoodson let me know what you think of this (specifically the naming)

ansible/openshift-tools-ansible.spec
```diff
@@ -0,0 +1,49 @@
+# %commit is intended to be set by tito custom builders provided
+# in the .tito/lib directory. The values in this spec file will not be kept up to date.
+%{!?commit:
+%global commit c64d09e528ca433832c6b6e6f5c7734a9cc8ee6f
+}
```
twiest: Does this actually work?

Do we need to set something in tito for it to work?

twiest:
@joelddiaz naming is fine by me. Rest looks good. Just had the one comment / question.

joelddiaz:
@twiest: deleted the %{!?commit... stuff out of the spec. tito built the RPM fine without it.

kwoodson:
:+1:

PR #385, opened 2016-04-11
initial spec for building libzabbix RPM out of openshift-tools

joelddiaz:


joelddiaz:
@kwoodson :+1: #384 in stg

PR #386, opened 2016-04-11
copy filters, inventory, docs and test from openshift-ansible

joelddiaz:
generate openshift-tools-ansible-inventory rpm

PR #389, opened 2016-04-12
copy filters, inventory, docs and test from openshift-ansible

joelddiaz:


joelddiaz:
@kwoodson :+1: #386 in stg

PR #399, opened 2016-04-12
router fixes for when a svc or a dc are missing.  oc utility fixes

kwoodson:


kwoodson:
@mwoodson :+1: in #398

PR #403, opened 2016-04-12
fixed order of host monitoring selinux volume mount

mwoodson:


mwoodson:
@dak1n1 approved #401 in stg

PR #404, opened 2016-04-12
router image parameter fixed.

kwoodson:


kwoodson:
@mwoodson :+1:  in #400

kwoodson:
Ok to test

PR #414, opened 2016-04-18
copy host/inventory tools from openshift-ansible/bin

joelddiaz:
and generate equivalent RPMs
clean up pylint


joelddiaz:
@kwoodson :+1: #413 in stg

PR #418, opened 2016-04-18
chmod +x for all the executable files in scripts-inventory-clients RPM

joelddiaz:
and pylint fixes


joelddiaz:
@kwoodson :+1: #417 in stg

PR #419, opened 2016-04-18
Inventory clients role

joelddiaz:
copied from openshift-tools/openshift/installer/atomic-openshift-3.1/roles/openshift_ansible_inventory and modified to use openshift-tools\* RPMs


twiest:
:+1:

PR #422, opened 2016-04-18
Added vendor-github.yml and vendored the latest openshift-ansible byo installer from github.

twiest:


twiest:
:+1:  in STG pr #421

PR #425, opened 2016-04-18
depend on base python-openshift-tools (for __init__.py)

joelddiaz:


joelddiaz:
@kwoodson :+1: #424 in stg

PR #427, opened 2016-04-19
push python-2.7.5-34 (SNI) dep into python-openshift-tools-web only

joelddiaz:


joelddiaz:
@kwoodson :+1: #426 in stg

PR #429, opened 2016-04-19
fix comment path

joelddiaz:


joelddiaz:
@kwoodson :+1: #428 in stg

PR #436, opened 2016-04-21
depend on ansible1.9 RPM from EPEL

joelddiaz:


joelddiaz:
@kwoodson :+1: #435 in stg

PR #437, opened 2016-04-21
no need for copr-openshift-ansible repo

joelddiaz:


joelddiaz:
@kwoodson :+1: #434 in stg

PR #438, opened 2016-04-22
remove openshift-ansible COPR repo

joelddiaz:
and depend on RPM 'ansible1.9' from EPEL


joelddiaz:
ok to test

joelddiaz:
ok to test

kwoodson:
:+1:

PR #430, opened 2016-04-20
Few oc utility updates and fixes. oadm_router attempt

kwoodson:

PR #439, opened 2016-04-22
Few oc utility updates and fixes. oadm_router attempt

kwoodson:


kwoodson:
@mwoodson :+1:  in #430

PR #440, opened 2016-04-22
remove openshift-ansible COPR repo

joelddiaz:
and depend on RPM 'ansible1.9' from EPEL


joelddiaz:
@kwoodson :+1: #438 in stg

PR #447, opened 2016-04-22
Fix kubeconfig copy inside of ansible startup

kwoodson:


kwoodson:
@mwoodson :+1:  in #445

PR #450, opened 2016-04-25
Adding a check for kubeconfig so it succeeds on nodes.

kwoodson:


kwoodson:
@joelddiaz :+1: in #449

PR #455, opened 2016-04-27
oadm updates to fix registry bug for serviceip

kwoodson:


kwoodson:
@mwoodson :+1:  in #454

PR #460, opened 2016-04-28
update symlink name to be more descriptive

joelddiaz:


joelddiaz:
@twiest :+1: #459 in stg

PR #461, opened 2016-04-28
renamed build-local-setup.sh to build-local-setup-centos7.sh and fixe…

twiest:
…d it to work with our latest way to build. Also fixed the local_development_monitoring.adoc to be in line with our current way of developing.


joelddiaz:
:+1:

PR #464, opened 2016-04-28
Route fixes and feature to accept paths and certs

kwoodson:


kwoodson:
@mwoodson :+1: in #463

PR #471, opened 2016-05-02
Add ability to force refresh on the registry

kwoodson:


kwoodson:
@mwoodson  :+1: in #470

PR #472, opened 2016-05-02
added new docker storage setup options

mwoodson:


kwoodson:
:+1:

ansible/roles/docker_storage_setup/templates/docker-storage-setup.j2
```diff
@@ -4,4 +4,5 @@
 # For more details refer to "man docker-storage-setup"
 DEVS={{ dss_docker_device }}
 VG=docker_vg
-
+DATA_SIZE=99%VG
+AUTO_EXTEND_POOL=no
```
detiber, line 8: Why don't you want to enable auto extending of the pool? wouldn't that make it easier for adding additional PVs to the VG.

I'm not sure the DATA_SIZE var is applied after initial configuration.

PR #473, opened 2016-05-02
added new docker storage setup options

mwoodson:


mwoodson:
@kwoodson approved #472 in stg

PR #476, opened 2016-05-03
Added cacert to router.

kwoodson:


kwoodson:
@mwoodson :+1: in #475

PR #477, opened 2016-05-03
set dm.basesize to 3G

joelddiaz:
EXTRA_DOCKER_STORAGE_OPTIONS is a no-op on 3.1 clusters because docker 1.8's /usr/bin/docker-storage-setup doesn't read/use it.
On 3.2 (docker 1.9) it will not allow container images sizes beyond 3G


twiest:
:+1:

PR #479, opened 2016-05-03
Added cat_fact module.

twiest:


detiber:
@twiest trying to see what this buys above command/register.

twiest:
@detiber The command module is WAY too loose and unstructured. We prefer to use modules specific to the job than the command module. Don't get me wrong, command is great in a pinch, but it's definitely not our goto module.

Also, I have plans to add features to this like "only cat if the file is < 10 megs" and other such things. Since this is going into memory, it's important to have stuff like that. This is just pass 1.

joelddiaz:
:+1:

PR #480, opened 2016-05-04
Added cat_fact module.

twiest:


twiest:
@joelddiaz +1d in STG pr #479

PR #490, opened 2016-05-05
added fstab_mount_options

mwoodson:


mwoodson:
@kwoodson approved #489 in stg

PR #492, opened 2016-05-05
added a function to prevent running outside bastion hosts

stenwt:


stenwt:
unchanged from #491 reviewed by @joelsmith

PR #494, opened 2016-05-06
add group module and extend user module to manage group membership

joelddiaz:


joelddiaz:
@kwoodson :+1: #493 in stg

PR #496, opened 2016-05-09
add extra group tests and update label tests

joelddiaz:


joelddiaz:
@kwoodson :+1: #495 in stg

PR #502, opened 2016-05-10
3.2 updates for router and registry

kwoodson:


kwoodson:
@mwoodson :+1: in #501

PR #505, opened 2016-05-11
update logic to avoid delete/create when updating existing node entries

joelddiaz:


joelddiaz:
@kwoodson @rharrison10 :+1: #503 in stg

PR #507, opened 2016-05-11
some tests in the form of an ansible playbook for lib_dyn

joelddiaz:


kwoodson:
:+1:  awesome test cases

PR #508, opened 2016-05-11
some tests in the form of an ansible playbook for lib_dyn

joelddiaz:


joelddiaz:
@kwoodson :+1: #507 in stg

PR #512, opened 2016-05-13
vendor 3.2 installer (3.0.88)

joelddiaz:


joelddiaz:
@twiest :+1: #511 in stg

PR #519, opened 2016-05-16
Updates for oc_secret. path and name required now

kwoodson:


kwoodson:
@mwoodson :+1: in #518

PR #521, opened 2016-05-17
Added error handling when secret fails creation

kwoodson:


kwoodson:
@mwoodson :+1: in #520

PR #522, opened 2016-05-17
Removed pruning from monitoring container.

dak1n1:
And commented out block that was causing config loop failures.

### DIFF
--- a/docker/oso-host-monitoring/centos7/root/config.yml
+++ b/docker/oso-host-monitoring/centos7/root/config.yml
@@ -148,11 +148,6 @@
       minute: "*/5"
       job: ops-runner -f -s 30 -n cskc.openshift.master.kubeconfig.status /usr/bin/cron-send-check-kubeconfig 
 
-    - name: "Clean old builds and deployments"
-      hour: "1"
-      minute: "0"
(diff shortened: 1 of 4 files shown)


dak1n1:
@twiest can I remove this cron job from the monitoring container? Earlier, mwoodson didn't want to +1 this because he said you have plans to keep the pruning inside a container.

But this is redundant, since we already have a role for pruning. We want this to run on a single master, which is exactly what the existing role does: 
https://github.com/openshift/openshift-ansible-ops/blob/prod/ops_roles/oso_auto_prune/tasks/main.yml#L41-55

So can I please clean up this old cruft and then later you guys can make a primary-master container?

twiest:
@dak1n1 I've created a card to move the pruner into the container. We definitely need it running inside the container and not outside.  The card is here:  https://trello.com/c/2pBmzmcy

Is this currently causing issues, or can we wait until that card is implemented (which should hopefully be soon)?

dak1n1:
@twiest it's not necessarily causing issues, but it confused NOC because NOC thought they were looking at the place where all pruning happens (inside the container) and didn't find the most current pruning jobs, which are outside the container. Then they passed that info on to QE who filed a bug with the Online Devs and that's how it came to my attention. https://bugzilla.redhat.com/show_bug.cgi?id=1336670

I just wanted to have all the cron jobs in one location so there's no confusion. But I clarified it with them so we can wait if you prefer. As long as it gets cleaned up eventually, that should be fine.

twiest:
@dak1n1 I absolutely agree that we need all our cron jobs in 1 place.

@joelddiaz has picked up that card, so I think it'll get done soon.

PR #526, opened 2016-05-18
Decode secrets and error handling added.

kwoodson:


kwoodson:
@mwoodson :+1:  in #525

PR #528, opened 2016-05-18
Adding selectors on delete.

kwoodson:


kwoodson:
@mwoodson :+1:  in #527

PR #530, opened 2016-05-18
added openshift_* roles

mwoodson:


mwoodson:
@kwoodson approved #529 in stg

PR #532, opened 2016-05-18
first pass at automating our verification process

stenwt:


gburges:
+1

detiber:
It might be a good thing to trigger a subset of the e2e/conformance tests.

@sdodson do you know the magic invocation they can use to trigger non-destructive e2e tests?

sdodson:
No, @jayunit100 should be able to help determine the correct invocation. That said, I think the e2e tests create their own namespaces so they should be out of the way.

jayunit100:
```
ginkgo.skip=Skipped|Disruptive|Slow|Flaky
```

and 

```
ginkgo.focus=Conformance
```

should do it for you

PR #533, opened 2016-05-18
Adding role for openshift_metrics.

kwoodson:

PR #534, opened 2016-05-18
Adding role for openshift_metrics.

kwoodson:


kwoodson:
@mwoodson :+1:  in #533

PR #538, opened 2016-05-18
shortening the project name in the cron-send-build-app.py

mwoodson:


mwoodson:
@kwoodson approved #537 in stg

PR #540, opened 2016-05-18
fixed a skydns issue

mwoodson:


mwoodson:
@dak1n1 approved #539 in stg

PR #545, opened 2016-05-19
pass environment OSO_MASTER_PRIMARY to monitoring container

joelddiaz:


joelddiaz:
@kwoodson :+1: #544 in stg

PR #546, opened 2016-05-19
add build/deploy/image pruning script

joelddiaz:


joelddiaz:
@kwoodson :+1: #541 in stg

PR #548, opened 2016-05-19
replace pruning cron with new script (from openshift-tools-monitoring)

joelddiaz:
and only run on primary master node


joelddiaz:
@kwoodson :+1: #547 in stg

PR #549, opened 2016-05-23
Moving inventory yaml accounts to dict format.

kwoodson:


ansible/inventory/multi_inventory.py
```diff
@@ -121,7 +118,7 @@ def load_yaml_config(self, conf_file=None):
             config = yaml.safe_load(conf)
 
         # Provide a check for unique account names
-        if len(set([acc['name'] for acc in config['accounts']])) != len(config['accounts']):
+        if len(set(config['accounts'])) != len(config['accounts']):
```
twiest: Can this actually happen now?

Won't YAML remove the duplicate?

twiest:
:+1:

PR #552, opened 2016-05-23
Added update call for list and dict entries.  Also added ability to only return value when key is specified.

kwoodson:


joelddiaz:
:+1:

kwoodson:
ok to test

joelddiaz:
:+1:

PR #557, opened 2016-05-23
Added update call for list and dict entries.  Also added ability to only return value when key is specified.

kwoodson:


kwoodson:
@joelddiaz :+1: in #552

PR #560, opened 2016-05-24
'try' each pruning attempt so an earlier error doesn't stop all pruning

joelddiaz:


scripts/monitoring/cron-openshift-pruner.py
```diff
@@ -183,17 +183,34 @@ def prune_deployments(self):
...
+            print "Error pruning builds"
+            rc = e.returncode
+
+        try:
+            self.prune_images()
+        except subprocess.CalledProcessError as e:
+            print "Error pruning images"
+            rc = e.returncode
+
+        if rc != 0:
```
twiest: if deployments and builds fail, but images succeeds, rc == 0, so no error.

twiest:
:+1:

PR #561, opened 2016-05-24
'try' each pruning attempt so an earlier error doesn't stop all pruning

joelddiaz:


joelddiaz:
@twiest :+1: #560 in stg

PR #571, opened 2016-05-25
Add trigger to alert when failed persistant volumes are present

rharrison10:
I feel this all I need since the item is already being reported.

https://trello.com/c/Kuh0Vjpv

### DIFF
--- a/ansible/roles/os_zabbix/vars/template_openshift_master.yml
+++ b/ansible/roles/os_zabbix/vars/template_openshift_master.yml
@@ -18,7 +18,7 @@ g_template_openshift_master:
     applications:
     - Openshift Master
     key: openshift.master.app.create
-  
+
   - key: openshift.master.app.build.create
     description: "check the app create with a build process"
     value_type: int
@@ -240,7 +240,7 @@ g_template_openshift_master:
   - key: openshift.master.nodesnotready.count
     description: "This check shows how many nodes in a cluster are in NotReady state."
     value_type: int
-    applications: 
+    applications:
     - Openshift Master
 
   - key: openshift.master.nodesnotschedulable.count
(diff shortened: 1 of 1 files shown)


ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -340,6 +340,18 @@ g_template_openshift_master:
     url: 'https://github.com/openshift/ops-sop/blob/master/V3/Alerts/openshift_master.asciidoc'
     priority: info
 
+  - name: 'Failed Persistent Volumes on {HOST.NAME}'
+    expression: '{Template Openshift Master:openshift.master.pv.failed.count.last(#2)}>0'
```
twiest: I would do `.min(#2) > 0` instead of `.last`. I try to not use `.last` as `.min` helps with flapping better.

Unless there's a reason `.last` is better in this specific case.

ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -340,6 +340,18 @@ g_template_openshift_master:
...
+    priority: info
+
+  - name: 'Less than 10 Available Persistent Volumes on {HOST.NAME}'
+    expression: '{Template Openshift Master:openshift.master.pv.available.count.last(#2)}<10'
```
twiest: I would do `.max(#2) < 10` instead of .last. I try to not use `.last`. `.max` helps with flapping better.

Unless there's a reason `.last` is better in this specific case.

ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -340,6 +340,18 @@ g_template_openshift_master:
...
+    priority: info
+
+  - name: 'Less than 5 Available Persistent Volumes on {HOST.NAME}'
+    expression: '{Template Openshift Master:openshift.master.pv.available.count.last(#2)}<5'
```
twiest: I would do `.max(#2) < 5` instead of .last. I try to not use `.last`. `.max` helps with flapping better.

Unless there's a reason `.last` is better in this specific case.

twiest:
:-1:

twiest:
:+1:

PR #574, opened 2016-05-26
add zabbix entries to hold cluster-wide calculated items

joelddiaz:


joelddiaz:
@twiest :+1: #573 in stg

PR #576, opened 2016-05-26
Adding cluster var support.

kwoodson:


twiest:
Wow, really?  That's all the code it took to enable this amazing feature?   

:+1:

PR #575, opened 2016-05-26
memory available and max-mem pod schedulable capacity checks

joelddiaz:

PR #579, opened 2016-05-27
memory available and max-mem pod schedulable capacity checks

joelddiaz:
plus new 'conversions' library
and necessary RPM spec updates to include capacity checks into monitoring-openshift.rpm


joelddiaz:
@twiest :+1: #575 in stg

PR #585, opened 2016-05-30
Fixed a bug for create.

kwoodson:


kwoodson:
@mwoodson :+1: in #584

PR #587, opened 2016-05-31
[wip]Dockerstorage

BlueShells:
Dockerstorage clean auto

### DIFF
--- a/scripts/monitoring/cron-send-docker-metrics.py
+++ b/scripts/monitoring/cron-send-docker-metrics.py
@@ -23,11 +23,11 @@ if __name__ == "__main__":
     zs = ZaggSender()
     try:
         cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
-        cli_storage = Client(base_url='unix://var/run/docker.sock')
+        #cli_storage = Client(base_url='unix://var/run/docker.sock')
         du = DockerUtil(cli)
-        Deadcontainers = cli_storage.containers(quiet=True,filters={'status':'exited'}) #get all the exited container
+        exited_containers = cli.containers(quiet=True,filters={'status':'exited'}) #get all the exited container
+        dead_containers = cli.containers(all=True,quiet=True,filters={'status':'dead'}) #get all the exited container
         du_dds = du.get_disk_usage()
-        print "data_space:",du_dds.data_space_percent_available
         keys = {
             'docker.storage.data.space.used': du_dds.data_space_used,
             'docker.storage.data.space.available': du_dds.data_space_available,
@@ -41,6 +41,8 @@ if __name__ == "__main__":
 
             'docker.storage.is_loopback': int(du_dds.is_loopback),
             'docker.ping': 1, # Docker is up
+            'docker.containers.exited':len(exited_containers),
+            'docker.containers.dead':len(dead_containers)
         }
     except (DockerException, TimeoutException) as ex:
         print "\nERROR talking to docker: %s\n" % ex.message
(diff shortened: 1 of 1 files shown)


scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -22,9 +23,11 @@
     zs = ZaggSender()
     try:
         cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
+        cli_storage = Client(base_url='unix://var/run/docker.sock')
         du = DockerUtil(cli)
+        Deadcontainers = cli_storage.containers(quiet=True,filters={'status':'exited'}) #get all the exited container
```
mwoodson: let's change "Deadcontainers" to "dead_containers"  We don't typically use camelCase for variables names.

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -46,8 +49,36 @@
...
     print "\nDone.\n"
+    
+    #start auto-heal 
+
+    #Deadcontainer = cli_storage.containers(filters={'id':'bf34a7c35b907c1f3047321b3073aa74f16957c2505af3644c2c1e7bb976bd78'})
+    #print '******************************8'
+    #print Deadcontainer[0]['Status']
+    print '***********************************'
+    if int(du_dds.data_space_percent_available) < 50:
+        print 'less than 50'
```
mwoodson: Let's make this more explicit: "Docker has less than 50% storage avaiable.  Attempting to clean up space."

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -22,9 +23,11 @@
     zs = ZaggSender()
     try:
         cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
+        cli_storage = Client(base_url='unix://var/run/docker.sock')
```
twiest: The reason we use `AutoVersionClient` is so that the version of docker can be different than what Dockerpy is hard coded for.

Is there a reason to not use the `AutoVersionClient` that's already here in the `cli` variable?

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -46,8 +49,36 @@
...
+        for container in Deadcontainers:
+            print container['Id']
+            #get the info
+            Deadcontainer = cli_storage.containers(all=True,filters={'id':container['Id']})
+            #print Deadcontainer
+            if len(Deadcontainer) == 1:
+                print '******************************'
+                print Deadcontainer[0]['Status']
+                print '***********************************'
+                if Deadcontainer[0]['Status'].find('Exited') != -1:
```
mwoodson: if 'Exited' in dead_container[0]['Status']:

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -22,9 +23,11 @@
...
         cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
+        cli_storage = Client(base_url='unix://var/run/docker.sock')
         du = DockerUtil(cli)
+        Deadcontainers = cli_storage.containers(quiet=True,filters={'status':'exited'}) #get all the exited container
         du_dds = du.get_disk_usage()
-
+        print "data_space:",du_dds.data_space_percent_available
         keys = {
             'docker.storage.data.space.used': du_dds.data_space_used,
             'docker.storage.data.space.available': du_dds.data_space_available,
```
twiest: Let's add a new key for tracking the number of exited containers. I think that would be something good to have a count of, especially since you're looking to use it for autohealing).

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -46,8 +49,36 @@
...
+                print Deadcontainer[0]['Status']
+                print '***********************************'
+                if Deadcontainer[0]['Status'].find('Exited') != -1:
+                    #do the remove step
+                    print Deadcontainer[0]['Status'].find('Exited')
+                    #cli_storage.remove_container(container=container['Id'])
+            else:
+                print 'container not exist'
+    else:
+        print 'not less',int(du_dds.data_space_percent_available)
```
mwoodson: more explicit.  "Docker storage has more than 50% available. Skipping autoheal to clean up space"

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -46,8 +49,36 @@
...
+            #get the info
+            Deadcontainer = cli_storage.containers(all=True,filters={'id':container['Id']})
+            #print Deadcontainer
+            if len(Deadcontainer) == 1:
+                print '******************************'
+                print Deadcontainer[0]['Status']
+                print '***********************************'
+                if Deadcontainer[0]['Status'].find('Exited') != -1:
+                    #do the remove step
+                    print Deadcontainer[0]['Status'].find('Exited')
```
twiest: I assume this is just for debugging. If so, it should be removed from final PR.

twiest:
@BlueShells @mwoodson   So, we need to be careful with this. The reason OpenShift leaves around containers is because once you remove the container, you lose the docker logs. Once we have logging deployed, this shouldn't be as big of an issue, but right now it is.

Have we investigated the docker cleanup stuff that's built into the openshift-node?

I know @mwoodson looked at it a few OpenShift versions back, but we might want to rely more heavily on that. And maybe change this to cleanup at like 90% (some high number that hopefully shouldn't be hit, unless openshift-node isn't doing it's job).

PR #590, opened 2016-05-31
Renamed cluster_group to clusterid

kwoodson:


kwoodson:
@mwoodson :+1: in #589

PR #593, opened 2016-05-31
Fixing listing capabilities when oo vars are missing

kwoodson:


kwoodson:
@dak1n1 :+1:  in #592

PR #594, opened 2016-05-31
Add tdgc role

wshearn:
Add new role for cloning a repository into a temp directory.

Useful for automating commits to another git repository.


mwoodson:
:+1:

PR #600, opened 2016-06-01
ignore non-ready nodes

joelddiaz:


joelddiaz:
@twiest :+1: #599 in stg

PR #601, opened 2016-06-01
[WIP]Storageclean

BlueShells:
add new keys for dead containers and exited containers , clean dead containers and exited containers when the avalable less than 90%

### DIFF
--- a/scripts/monitoring/cron-send-docker-metrics.py
+++ b/scripts/monitoring/cron-send-docker-metrics.py
@@ -8,7 +8,7 @@
 # Adding the ignore because it does not like the naming of the script
 # to be different than the class name
 # pylint: disable=invalid-name
-
+# pylint: disable=line-too-long
 
 from docker import AutoVersionClient
 from docker.errors import DockerException
@@ -22,9 +22,11 @@ if __name__ == "__main__":
     zs = ZaggSender()
     try:
         cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
+        #cli_storage = Client(base_url='unix://var/run/docker.sock')
         du = DockerUtil(cli)
+        exited_containers = cli.containers(quiet=True, filters={'status':'exited'}) #get all the exited container
+        dead_containers = cli.containers(all=True, quiet=True, filters={'status':'dead'}) #get all the exited container
         du_dds = du.get_disk_usage()
-
         keys = {
             'docker.storage.data.space.used': du_dds.data_space_used,
             'docker.storage.data.space.available': du_dds.data_space_available,
@@ -38,6 +40,8 @@ if __name__ == "__main__":
 
             'docker.storage.is_loopback': int(du_dds.is_loopback),
             'docker.ping': 1, # Docker is up
+            'docker.containers.exited':len(exited_containers),
+            'docker.containers.dead':len(dead_containers)
         }
     except (DockerException, TimeoutException) as ex:
         print "\nERROR talking to docker: %s\n" % ex.message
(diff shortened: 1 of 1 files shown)


scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -8,7 +8,7 @@
 # Adding the ignore because it does not like the naming of the script
 # to be different than the class name
 # pylint: disable=invalid-name
-
+# pylint: disable=line-too-long
```
mwoodson: we can remove this and break up the line below

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -46,8 +50,54 @@
...
+                    try:
+                        cli.remove_container(container=container['Id'])
+                        print 'done this container'
+                    except Exception, e:
+                        print 'something wrong during the remote of the container'
+
+            else:
+                print 'container not exist'
+    else:
+        print 'Docker storage has more than 90% available. Skipping autoheal to clean up space', int(du_dds.data_space_percent_available)
```
mwoodson: ---

```
  print` 'Docker storage has more than 90% available. Skipping autoheal to clean up space', 
       int(du_dds.data_space_percent_available)
```

---

mwoodson:
@twiest.  I watched this work with him on a dev-preview-int.  I'd like another set of eyes on it as well.  One of things we noticed is that the containers would fail to remove for a few different reasons (file system of the container was locked came up a few times), that is why the try/except are in there.

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -22,9 +22,11 @@
     zs = ZaggSender()
     try:
         cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
+        #cli_storage = Client(base_url='unix://var/run/docker.sock')
         du = DockerUtil(cli)
+        exited_containers = cli.containers(quiet=True, filters={'status':'exited'}) #get all the exited container
+        dead_containers = cli.containers(all=True, quiet=True, filters={'status':'dead'}) #get all the exited container
```
twiest, line 31: Nice, I'm glad you got the AutoVersionClient working.

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -46,8 +50,54 @@
...
+            if len(dead_container) == 1:
+                print 'the status of the container is :', dead_container[0]['Status']
+                if dead_container[0]['Status'].find('Dead') != -1:
+                    #do the remove step
+                    print 'status confirmed :', dead_container[0]['Status']
+                    try:
+                        cli.remove_container(container=container['Id'])
+                        print 'done this container'
+                    except Exception, e:
+                        print 'something wrong during the remote of the container'
```
twiest: It seems to me that the the handling of dead containers and exited containers is identical except for the status of 'Dead' or 'Exited'.

I'd put these in a function. Something like this (UNTESTED):

```
def remove_container(self, status):
    # insert the common code here

# ... snip ...

for container in dead_containers:
    self.remove_container('Exited')

# ... snip ...

for container in dead_containers:
    self.remove_container('Dead')
```

Always try to apply the DRY principle:
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

twiest:
@mwoodson ok, sounds good.

@BlueShells I would like us to make sure that the openshift-node container garbage collection stuff is enabled and working BEFORE we push this code.  I want openshift-node handling the day to day cleanup and this to only handle the extreme case.

Feel free to talk to Goulding about priority on that.

PR #602, opened 2016-06-01
detect/enable cluster capacity reporting

joelddiaz:
in the data hierarchy, set g_enable_cluster_capacity_reporting to True for any cluster that should have this reporting enabled


twiest:
:+1:

PR #604, opened 2016-06-01
Fixed a bug with copy and create with contents.

kwoodson:


kwoodson:
@mwoodson :+1:  in #603

PR #608, opened 2016-06-01
integrate build checks to app create loop

stenwt:


kwoodson:
:+1:  give it a whirl!

PR #610, opened 2016-06-01
Fixed return value when updating a list.  Now check before update.

kwoodson:


kwoodson:
@mwoodson :+1: in #609

PR #612, opened 2016-06-01
Content is now idempotent

kwoodson:


kwoodson:
@mwoodson :+1: in #611

PR #621, opened 2016-06-06
Removing multi_inventory.yaml management

kwoodson:


kwoodson:
@mwoodson :+1: in #620

PR #623, opened 2016-06-06
add step to remove /etc/sysconfig/docker-storage file

joelddiaz:
old thin pool settings in the file can keep docker-storage-setup from working properly


joelddiaz:
@kwoodson :+1: #622 in stg

PR #625, opened 2016-06-07
Bug fix for first level entries, added append, and fixed update

kwoodson:


kwoodson:
@mwoodson :+1:  in #624

PR #628, opened 2016-06-07
fix for the s3 docker registry

mwoodson:


mwoodson:
@stenwt approved #627

PR #629, opened 2016-06-07
new aws account setup

mwoodson:


mwoodson:
@gburges approved #619

PR #626, opened 2016-06-07
move monitoring container configuration into passed-in/bound file

joelddiaz:
Rather than pass numerous environment variable to control how the host-monitoring container is started/configured, move environment vars into a yaml file that can be bound in.
This will configure things like the list of cron jobs to instantiate, the clusterid, zagg settings, etc. Logic that determines what gets installed is moved into the oso-host-monitoring role when it generates a monitoring-config.yml file for each node.
Config loop will now install /etc/openshift_tools/monitoring-config.yml, and the systemd file for host-monitoring will bind it in at the same location.

For local host-monitoring testing/development, create a reduced monitoring-config.yml which will not include Node/Master-specific checks (that will ultimately fail on a local environment).

PR #630, opened 2016-06-07
move monitoring container configuration into passed-in/bound file

joelddiaz:
Rather than pass numerous environment variable to control how the host-monitoring container is started/configured, move environment vars into a yaml file that can be bound in.
This will configure things like the list of cron jobs to instantiate, the clusterid, zagg settings, etc. Logic that determines what gets installed is moved into the oso-host-monitoring role when it generates a monitoring-config.yml file for each node.
Config loop will now install /etc/openshift_tools/monitoring-config.yml, and the systemd file for host-monitoring will bind it in at the same location.

For local host-monitoring testing/development, create a reduced monitoring-config.yml which will not include Node/Master-specific checks (that will ultimately fail on a local environment).


joelddiaz:
@twiest :+1: #626 in stg

PR #642, opened 2016-06-08
Added more attach names for the root master and node volumes and fixed

twiest:
bug.


twiest:
Was +1d in STG PR #641

PR #645, opened 2016-06-09
use correct variable name that holds list of crons

joelddiaz:


joelddiaz:
@twiest :+1: #644 in stg

PR #646, opened 2016-06-09
Changed root device detection to be more accurate. Added volume tag a…

twiest:
…nd attachment data to snapshot tags.


joelddiaz:
:+1:

PR #654, opened 2016-06-09
stop passing in environment variable to host-monitoring container

joelddiaz:
only one script uses one of these environment variables (scripts/monitoring/cron-send-create-app.py), and the cron job for it is written in a way where the needed environment variable is explicitely passed in before launching it


kwoodson:
:+1:

PR #657, opened 2016-06-10
allow o+rx on openshift_tools

joelddiaz:


joelddiaz:
@twiest @stenwt :+1: #656 in stg

PR #662, opened 2016-06-13
Fixing a bug in multi_inventory.py when key does not exist

kwoodson:


kwoodson:
@mwoodson :+1:  in #661

PR #668, opened 2016-06-14
Adding latest gce.py

kwoodson:


kwoodson:
@ivanhorvath :+1: in #667

PR #674, opened 2016-06-14
use nodejs-ex instead of nodejs-example for STI test

stenwt:


stenwt:
@kwoodson reviewed #673

PR #681, opened 2016-06-15
added the postfix_amazon ses client role; updated from address on gpg send role

mwoodson:


mwoodson:
@twiest approved #680

PR #683, opened 2016-06-15
removed dump extra .com

mwoodson:


mwoodson:
@twiest approved #682

PR #694, opened 2016-06-16
pushing generated dockerfiles so they have psutil in them

ivanhorvath:


ivanhorvath:
@kwoodson :+1:  in a20e255

PR #700, opened 2016-06-17
changing the cron entries, because they were lacking the names...

ivanhorvath:
… and adding defattr to files section in one of the packages in the script spec file

### DIFF
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -198,15 +198,15 @@ host_monitoring_cron:
 
 - name: send etcd connections count on port 2379
   minute: "*/5"
-  job: ops-runner -f -s 30 -n /usr/bin/cron-send-connection-count.py etcd 2379 openshift.master.etcd.port_2379.connections.established
+  job: ops-runner -f -s 30 -n cscc.openshift.master.etcd.connections.2379 /usr/bin/cron-send-connection-count etcd 2379 openshift.master.etcd.port_2379.connections.established
 
 - name: send etcd connections count on port 2380
   minute: "*/5"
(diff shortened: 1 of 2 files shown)


kwoodson:
:+1:

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -198,15 +198,15 @@ host_monitoring_cron:
...
 
 - name: send etcd connections count on port 2380
   minute: "*/5"
-  job: ops-runner -f -s 30 -n /usr/bin/cron-send-connection-count.py etcd 2380 openshift.master.etcd.port_2380.connections.established
+  job: ops-runner -f -s 30 -n cscc.openshift.master.etcd.connections.2380 /usr/bin/cron-send-connection-count etcd 2380 openshift.master.etcd.port_2380.connections.established
 
 - name: send master api server connections count on port 443
   minute: "*/5"
-  job: ops-runner -f -s 30 -n /usr/bin/cron-send-connection-count.py openshift 443 openshift.master.api.port_443.connections.established
+  job: ops-runner -f -s 30 -n cscc.openshift.master.api.connections.ssl /usr/bin/cron-send-connection-count openshift 443 openshift.master.api.port_443.connections.established
```
twiest: I'd rather have this be `cscc.openshift.master.api.connections.443` to match the pattern.

ssl is generic. We can use ssl on any port. https would be better, but I still prefer the explicit 443.

twiest:
:+1:

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
ivanhorvath: Changed it, pull request is updated, want to give it a thumbs up Thomas?

PR #701, opened 2016-06-17
changing the cron entries, because they were lacking the names and ad…

ivanhorvath:
…ding defattr to files section in one of the packages in the script spec file


ivanhorvath:
@kwoodson @twiest :+1: in 94315e40549107ddeee3c06b1543e604b44542f5

PR #710, opened 2016-06-20
fixed handling urlerror, added timeout

stenwt:


stenwt:
@blentz @rharrison10 reviewed #702

PR #711, opened 2016-06-20
Clarification Updates

brs17:
Specify required Ansible version.
Updated Docker containers list url 
Emphasize importance to setup ZAIO before running oso-zagg-web and oso-host-monitoring.


twiest:
:+1:

PR #720, opened 2016-06-22
Adding instance_states to gce.

kwoodson:


kwoodson:
@mwoodson :+1:  in #669

PR #722, opened 2016-06-22
Changed the thresholds of the established connections as well as critically low memory triggers

twiest:


twiest:
+1d in STG pr #721

PR #725, opened 2016-06-22
lib_gcloud deployment-manager.

kwoodson:


kwoodson:
@mwoodson :+1: in #724

PR #731, opened 2016-06-23
Role to install gcp packages.

kwoodson:


kwoodson:
@mwoodson :+1:  in #730

PR #746, opened 2016-06-23
Migrating filter plugins to openshift-tools ops

kwoodson:

PR #747, opened 2016-06-23
Migrating filter plugins to openshift-tools ops

kwoodson:


kwoodson:
@joelddiaz :+1: in #746

PR #755, opened 2016-06-27
add item to hold FailedScheduling event counts

joelddiaz:

PR #758, opened 2016-06-27
add item to hold FailedScheduling event counts

joelddiaz:


joelddiaz:
double :+1: #755 in stg

PR #759, opened 2016-06-27
openshift event watcher/reporter

joelddiaz:
Use /usr/bin/oc to watch events stream in as json.
Wake up a thread every minute to count and report on the events (via --watch-for parameter).

### DIFF
--- a/scripts/monitoring/cron-event-watcher.py
+++ b/scripts/monitoring/cron-event-watcher.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python
+# vim: expandtab:tabstop=4:shiftwidth=4
+'''
+  Process OpenShift event stream
+'''
+#
+#   Copyright 2016 Red Hat Inc.
+#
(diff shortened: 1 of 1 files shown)


twiest:
:+1:

scripts/monitoring/cron-event-watcher.py
```diff
@@ -0,0 +1,161 @@
...
+import argparse
+from openshift_tools.monitoring.zagg_sender import ZaggSender
+import json
+from Queue import Queue
+import re
+import subprocess
+import threading
+import time
+
+EVENT_PERIOD = 60 # seconds
```
kwoodson: Any reason not to put this into the config file?  Seems like we could potentially want to change the interval without having to rebuild.

joelddiaz:
@kwoodson: updated PR to allow for passing in custom event timeout period

kwoodson:
:+1:

twiest:
:+1:

PR #760, opened 2016-06-27
Added check for nodes without labels

brs17:
Check for and notify on nodes that do not have labels

### DIFF
--- a/scripts/monitoring/cron-send-os-master-metrics.py
+++ b/scripts/monitoring/cron-send-os-master-metrics.py
@@ -382,30 +382,30 @@ class OpenshiftMasterZaggClient(object):
             {'openshift.master.nodesnotschedulable.count' : len(nodes_not_schedulable)})
 
 
-def nodes_not_labeled(self):
-    """ check the nodes in the cluster that are not labeled
-        Note: This check only searches for nodes with no label keys set"""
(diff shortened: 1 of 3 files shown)


scripts/monitoring/cron-send-os-master-metrics.py
```diff
@@ -372,6 +373,34 @@ def nodes_not_ready(self):
...
+
+        print "\nPerforming nodes not labeled check..."
+
+        response = self.ora.get('/api/v1/nodes')
+
+        nodes_not_labeled = []
+        nodes_labeled = []
+        error_found = 1
+
+        for n in response['items']:
```
joelddiaz, line 396: this section can probably be simplified to something like (untested!):

```
for n in response['items']:
  if 'labels' in n['metadata']:
    nodes_labeled.append(n['metadata']['name'])
  else:
    nodes_not_labeled.append(n['metadata']['name'])
```

brs17, line 396: @joelddiaz yeah I was looking to do something like that but was having issues, I don't think I was looking for 'labels' in the right place.

Tested your code and it is spot on!

joelddiaz:
lgtm :+1:

PR #762, opened 2016-06-28
openshift event watcher/reporter

joelddiaz:
Use /usr/bin/oc to watch events stream in as json.
Wake up a thread every minute to count and report on the events (via --watch-for parameter).


joelddiaz:
double :+1: #759 in stg

PR #763, opened 2016-06-28
add flock without erroring on lock already acquired

joelddiaz:


joelddiaz:
double :+1: #761 in stg

PR #768, opened 2016-06-28
deal with new groups being added to a user

joelddiaz:


joelddiaz:
double :+1: #766 in stg

PR #769, opened 2016-06-28
allow passing in auto-logout value for zabbix users

joelddiaz:


joelddiaz:
double :+1: #767 in stg

PR #771, opened 2016-06-28
Added os_zabbix_cluster_stats

twiest:


twiest:
+1d in STG PRs #765 and #770

PR #774, opened 2016-06-29
mark zabbix item as a "%" unit

joelddiaz:


joelddiaz:
double :+1: #773 in stg

PR #775, opened 2016-06-29
Added check to ops-docker-storage-reinitialize.yml for when the docker_vg is being used by another device.

twiest:


robotmaxtron:
:+1:

joelddiaz:
:+1:

PR #778, opened 2016-06-30
report avg cpu and mem allocations for compute nodes across the cluster

joelddiaz:

PR #781, opened 2016-06-30
report avg cpu and mem allocations for compute nodes across the cluster

joelddiaz:


joelddiaz:
double :+1: #778 in stg

PR #783, opened 2016-06-30
Added graphs for mem.users_available and cpu.users_available.

twiest:


twiest:
+1d in STG PRs #779 and #782

PR #789, opened 2016-07-05
Fix for ossh oscp refresh-cache option.

kwoodson:


kwoodson:
@joelddiaz :+1: in #788

PR #791, opened 2016-07-05
added ops_customizations

mwoodson:


mwoodson:
@twiest approved #790

PR #799, opened 2016-07-06
Cpu mem check

ivanhorvath:


ivanhorvath:
@joelddiaz :+1: in  7eda558

PR #802, opened 2016-07-06
Add ops_map_format filter to the ops_filters plugin

joelsmith:

PR #805, opened 2016-07-06
Add ops_map_format filter to the ops_filters plugin

joelsmith:


joelsmith:
Already got a 👍 in #802

PR #807, opened 2016-07-06
Resource builder for gcloud deployment-manager.

kwoodson:


mwoodson:
:+1:

mwoodson:
:+1:

kwoodson:
The last error is too-many-lines but I am unable to turn it off.  This is required to be on the first line of the file.  I attmepted this but it will not execute if that's the case.

PR #815, opened 2016-07-07
add new cluster-wide calculations

joelddiaz:
1) cluster-wide max allocatable memory (for compute nodes)
2) cluster-wide max cpu (for compute nodes) in milicores
3) cluster-wide cpu units scheduled across compute nodes (total in milicores and percentage)
4) cluster-wide cpu unscheduled resources across compute nodes (total in milicores and percentage)
5) cluster-wide mem scheduled across compute nodes (total in bytes and percentage)
6) cluster-wide mem unscheduled resources across compute nodes (total in bytes and percentage)
7) cluster-wide cpu and memory oversubscription (using cpu/mem limits on running pods)

### DIFF
--- a/scripts/monitoring/cron-send-cluster-capacity.py
+++ b/scripts/monitoring/cron-send-cluster-capacity.py
@@ -38,6 +38,7 @@ class OpenshiftClusterCapacity(object):
         self.zagg_sender = None
         self.ora = None
         self.sql_conn = None
+        self.zbx_key_prefix = "openshift.master.cluster.compute_nodes."
 
     def run(self):
         '''  Main function to run the check '''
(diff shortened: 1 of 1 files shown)


detiber:
Just as an fyi, I know that metrics aren't installed across the board yet, but once metrics are available, hawkular contains most (if not all) of this information. https://github.com/openshift/origin-metrics/blob/master/docs/hawkular_metrics.adoc has some information on it. The relevant cluster-level and node-level metrics are stored in the `_system` tenant.

twiest:
:+1:

twiest:
@detiber thanks for the info. We'll look at converting this script once we have metrics installed in more places.

PR #816, opened 2016-07-07
Adding metadata tag for newly provisioned instances to prepare for bootstrap.

kwoodson:


kwoodson:
@mwoodson :+1:  in #812

PR #820, opened 2016-07-07
add new cluster-wide calculations

joelddiaz:
1) cluster-wide max allocatable memory (for compute nodes)
2) cluster-wide max cpu (for compute nodes) in milicores
3) cluster-wide cpu units scheduled across compute nodes (total in milicores and percentage)
4) cluster-wide cpu unscheduled resources across compute nodes (total in milicores and percentage)
5) cluster-wide mem scheduled across compute nodes (total in bytes and percentage)
6) cluster-wide mem unscheduled resources across compute nodes (total in bytes and percentage)
7) cluster-wide cpu and memory oversubscription (using cpu/mem limits on running pods)


joelddiaz:
:+1: #815 in stg

PR #826, opened 2016-07-07
Fixed the gcloud path in the module.

kwoodson:


kwoodson:
@mwoodson :+1: in #825

PR #828, opened 2016-07-07
add --list-clusters to ohi

stenwt:


joelddiaz:
:+1:

blentz:
Syntax looks fine to me, so :+1: from that standpoint.

I'd like @kwoodson or @joelddiaz or someone else that knows ohi to comment on whether the change is in the right place and doing the right things.

PR #832, opened 2016-07-08
add more generous random sleeping for various cron jobs

joelddiaz:


joelddiaz:
@twiest :+1: #831 in stg

PR #834, opened 2016-07-11
Manifest and compute addresses added.

kwoodson:


kwoodson:
@mwoodson :+1: in #834

PR #840, opened 2016-07-11
run event watch to catch/report FailedScheduling events

joelddiaz:


joelddiaz:
:+1: #839 in stg

joelddiaz:
ok to test

PR #844, opened 2016-07-12
Adding projectinfo for metadata and sshkeys.

kwoodson:


kwoodson:
@mwoodson :+1: in #843

PR #846, opened 2016-07-12
Bug fix for logic when comparing ssh keys.

kwoodson:


kwoodson:
@mwoodson :+1: in #845

PR #848, opened 2016-07-12
Updating code to take into account colons in ssh key hosts.

kwoodson:


kwoodson:
@mwoodson :+1: in #847

PR #852, opened 2016-07-12
First attempt at service accounts for gcp.

kwoodson:


kwoodson:
@mwoodson :+1:  in #851

PR #857, opened 2016-07-13
First attempt at dm reconciler.

kwoodson:


kwoodson:
@mwoodson :+1:  in #855

PR #860, opened 2016-07-13
updated gcp cluster creation with reconciler

mwoodson:


mwoodson:
@kwoodson approved #859

PR #868, opened 2016-07-14
refactor to use subprocess.communicate(), capture all pod logs in the project on failure

stenwt:


a13m:
I am definitely :+1: to this change, but I would like to have @kwoodson review, in case there was a reason for not doing it this way originally.  Just to recap, this change is to avoid the deadlock described here in the documentation for Popen.wait():

https://docs.python.org/2/library/subprocess.html#popen-objects

scripts/monitoring/cron-send-create-app.py
```diff
@@ -136,17 +140,16 @@ def oc_cmd(self, cmd):
         proc = subprocess.Popen(cmds, stdout=subprocess.PIPE, stderr=subprocess.PIPE, \
                                 env={'KUBECONFIG': self.kubeconfig, \
                                      'PATH': os.environ["PATH"]})
-        proc.wait()
+        result = proc.communicate()
```
kwoodson: stdout, stderr = proc.communicate()

kwoodson:
:+1:

PR #872, opened 2016-07-14
refactor oocmd to use subprocess.communicate(), output all pod logs in the project on fail

stenwt:


stenwt:
@a13m @kwoodson reviewed #868

PR #875, opened 2016-07-15
Fixed typo with node related zabbix checks(node-not-ready/not-labeled)

brs17:
Added zabbix check for non-schedulable nodes
Added proper sop links in zabbix


stenwt:
:+1:

PR #877, opened 2016-07-15
Service account keys and project policy modules.

kwoodson:


kwoodson:
@mwoodson :+1: in #876

PR #881, opened 2016-07-15
Adding pd pv support for dm for gcp.

kwoodson:


kwoodson:
@mwoodson :+1: in #880

PR #883, opened 2016-07-15
more gcp roles and changes

mwoodson:


mwoodson:
@kwoodson approved #882

PR #885, opened 2016-07-18
delete the cert before upload

BlueShells:
delte the cert before upload


dak1n1:
:+1:  since this should resolve a recurring issue

PR #892, opened 2016-07-20
container build env fingerprint

drewandersonnz:
Gather information from the build environment and store this in /etc/oso-container-build-env-fingerprint.

This is to help quickly identify exactly which version of the container is currently deployed, along with who built it, and from which machine it was built.

It will help solve problems when older versions of the container are built, tagged as latest, and deployed.


twiest:
:+1:

mwoodson:
:+1 this is a pretty cool idea.  Thanks for putting this together.

PR #898, opened 2016-07-20
added tags support for the dm builder in gcp

mwoodson:


mwoodson:
@twiest approved #897

PR #899, opened 2016-07-20
Improved nodes_not_schedulable check

brs17:
Check avoids master nodes
Reincluded zabbix trigger for check


scripts/monitoring/cron-send-os-master-metrics.py
```diff
@@ -338,6 +339,28 @@ def pv_info(self):
...
+
+        print "\nPerforming nodes not schedulable check..."
+
+        response = self.ora.get('/api/v1/nodes')
+
+        nodes_not_schedulable = []
+
+        for n in response['items']:
+            if n['metadata']['labels']['type'] == 'master':
+                print "Node: %s is a master\n" % n['metadata']['name']
```
stenwt: I'd say only print this if the verbose flag is on

PR #901, opened 2016-07-20
Etcd metric sender update with config

ivanhorvath:
No changes to the spec file or crontabs deployed in oso_host_monitoring because I rewrote the existing etcd_stats sender and incorporated the old parts as a separate section coming from the config  in the new design. 

The reason there is 2 commits in this PR because I wanted to preserve the fully dynamic way in case we decide to go that route in the future or we need something like it in a different check.


twiest:
:+1:

PR #906, opened 2016-07-21
added ops-runner timeouts to create-app tests

stenwt:


stenwt:
@joelddiaz reviewed #905

PR #907, opened 2016-07-21
changing the location of inital config for osohm container

ivanhorvath:


joelddiaz:
LGTM :+1:

docker/oso-host-monitoring/centos7/root/config.yml
```diff
@@ -20,6 +20,24 @@
...
+
+  - name: copy all yaml files found in /container_setup to /etc/openshift_tools
+    copy:
+      src: "{{ item }}"
+      dest: /etc/openshift_tools/
+      owner: root
+      group: root
+      mode: 0644
+    with_fileglob:
+        - /container_setup/*
```
twiest: I don't think this is what we were talking about, right?

twiest:
``` yaml

- include_vars: "{{ item }}"
  with_fileglob:
  - /container_setup/*.yml


- copy:
    content: "{{ etcd_metrics_config }}"
    dest: /etc/openshift_tools/etcd_metric.yml
  when: etcd_metrics_config is defined

```

twiest:
:+1:

PR #910, opened 2016-07-21
changing the location of inital config for osohm container

ivanhorvath:


ivanhorvath:
@twiest :+1:  in #907

PR #912, opened 2016-07-21
fixing yml extension

ivanhorvath:


ivanhorvath:
@joelddiaz :+1:  in #911

PR #913, opened 2016-07-22
 container build env fingerprint

drewandersonnz:
Gather information from the build environment and store this in /etc/oso-container-build-env-fingerprint.

This is to help quickly identify exactly which version of the container is currently deployed, along with who built it, and from which machine it was built.

It will help solve problems when older versions of the container are built, tagged as latest, and deployed.

Was https://github.com/openshift/openshift-tools/pull/892

Speaking with @gburges I wasn't aware we don't merge stg->prod any time .. this PR should solve that


gburges:
+1 on monday :)

PR #916, opened 2016-07-22
temporary fix for the wrong file extension issue

ivanhorvath:


ivanhorvath:
@kwoodson and @dak1n1 :+1: in #914 
confirmation from @a13m on irc so this can go out

PR #919, opened 2016-07-22
Adding storage buckets to dm.

kwoodson:


kwoodson:
@mwoodson :+1: in #918

PR #921, opened 2016-07-22
Normalizing on oo vars instead of tag_.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #920

PR #927, opened 2016-07-25
Renamed host monitoring for better recognition in zabbix

brs17:


brs17:
I believe we concluded renaming the host-monitoring container was all that needed to be renamed.

joelddiaz:
:+1:

PR #930, opened 2016-07-25
added gcp service account, update haproxy passthrough

mwoodson:


mwoodson:
@kwoodson approved #929

PR #939, opened 2016-07-26
fixed dependency name

mwoodson:


mwoodson:
@kwoodson approved #938

PR #940, opened 2016-07-26
gcp registry support. dm storage support. service account json support.

kwoodson:


kwoodson:
@mwoodson :+1:  in #935

PR #947, opened 2016-07-26
service account keys json format fix.

kwoodson:


kwoodson:
@mwoodson :+1:  in #946

PR #948, opened 2016-07-26
removing push from osohm/rhel7 and adding tag

ivanhorvath:


docker/oso-host-monitoring/rhel7/build.sh
```diff
@@ -20,6 +20,12 @@ cd $(dirname $0)
   # Build ourselves
   echo
   echo "Building oso-rhel7-host-monitoring..."
+  TAG=''
+  GITBRANCH=$(git name-rev --name-only HEAD)
+  if [ "${GITBRANCH}" != prod ]
+  then
+    TAG=":${GITBRANCH}"
+  fi
```
twiest, line 28: Why don't we want this on prod?

kwoodson, line 28: From our meeting decision: 

prod will be latest for prod.   stg for stg and so on.  We want to nuke latest.

PR #949, opened 2016-07-26
Have the Zabbix config run in cron and report results to Zabbix

rharrison10:


rharrison10:
@kwoodson :+1:  in #942

PR #959, opened 2016-07-27
Print first IP address in json output. 

wshearn:
convert_to_ip will return 1 ipaddress in our use case(awsutil converts the host into an array and returns the array of ip address).


kwoodson:
:+1:

PR #962, opened 2016-07-27
Creating separate registries.

kwoodson:


kwoodson:
@mwoodson :+1: in #961

PR #966, opened 2016-07-28
using common kms names for each cluster

joelddiaz:
so no need to store anything. kms names can be generated from cluster names


joelddiaz:
:+1: #965 in stg

PR #967, opened 2016-07-28
added support for generic service account in gcp

mwoodson:

PR #968, opened 2016-07-28
added support for generic service account in gcp

mwoodson:


mwoodson:
@kwoodson approved #967

PR #972, opened 2016-07-28
module to return gcp zones.

kwoodson:


kwoodson:
@mwoodson :+1:  in #971

PR #974, opened 2016-07-28
added gcp zone getter

mwoodson:


mwoodson:
@kwoodson approved #973

PR #981, opened 2016-08-01
rename iam_kms to oo_iam_kms

joelddiaz:
...to indicate that this is our module (not something pulled/backported)

PR #983, opened 2016-08-01
rename OpenShift Ops roles to prefix with oo_

joelddiaz:
...to indicate that this is our module (not something pulled/backported)
iam_kms
cat_fact
sysconfig_fact


joelddiaz:
:+1: #981 in stg

PR #984, opened 2016-08-02
use renamed oo_ec2_ami20 module

joelddiaz:
and unconditionally pass the account ID to oo_ec2_ami20 to also grant AMI copy permissions (to allow creating encrypted AMIs in the target account).


mwoodson:
:+1 I like it

PR #986, opened 2016-08-02
added gcloud config ansible module, created a role with it

mwoodson:

PR #987, opened 2016-08-02
added gcloud config ansible module, created a role with it

mwoodson:


mwoodson:
@kwoodson approved #986

PR #988, opened 2016-08-02
Master reconciler

robotmaxtron:


dak1n1:
Isn't this more of a "policy update" than a "reconciler update"? I read over this because I was excited to see some automation for the openshift reconciler, which is a pod that runs on Online. The name "openshift reconciler" here might be confusing.

a13m:
Just saw this go by in email, and I mostly agree with @dak1n1 about the name.  Perhaps "policy reconciler" or "policy reconciliation" ?  It really is "reconciliation", and the subcommand is "reconcile", so I think it makes sense to have some form of that term in the role name, but without the word "policy" it's not clear what it is that you're reconciling.  (Of course, the reconciler pod in online should also have a descriptor in its name, for the same reason.  :-) )

robotmaxtron:
Closing in favor of renaming and trying to convert an existing playbook into roles.

PR #990, opened 2016-08-03
Adding gcp support for osohm.

kwoodson:


kwoodson:
@mwoodson :+1: in #989

PR #994, opened 2016-08-03
add oo_ec2_ami_copy Ansible module

joelddiaz:
can list AMI details (by AMI ID).
can create a copy of an AMI ID with or without encryption.
encryption can be default, or using IAM KMS arn, or KMS alias name.

determining whether an AMI already exists is done by AMI name, but there is a delay from when an AMI copy is started until when it appears in queries. so running repeatedly in succession will result in more than one AMI image with the same name. module will spin waiting for AMI to be in 'available' state before returning.


joelddiaz:
:+1: #993 in stg

PR #998, opened 2016-08-03
added the role openshift_gcp_add_users_to_project

mwoodson:


mwoodson:
@kwoodson approved #997

PR #999, opened 2016-08-04
add role that can be called during cluster creation

joelddiaz:
creates encrypted AMI using passed-in KMS alias name
saves ID of new AMI to specified directory (ie ../<account>/<env>/<cluster>/<sublocation>/ami.yml)


mwoodson:
:+1:

PR #1018, opened 2016-08-11
add item to hold number of haproxy processes found in CLOSE-WAIT state

joelddiaz:


joelddiaz:
:+1: #1017 in stg

joelddiaz:
ok to test

PR #1016, opened 2016-08-11
script to identify and stop stuck haproxy processes

joelddiaz:

PR #1022, opened 2016-08-12
script to identify and stop stuck haproxy processes

joelddiaz:
also pull it into openshift-scripts-monitoring-autoheal RPM


joelddiaz:
:+1: #1016 in stg

PR #1025, opened 2016-08-12
Openshift master upgrade

robotmaxtron:


ansible/roles/openshift_master_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+- name: Stop OpenShift Controller
+  service: 
+    name: atomic-openshift-master-controllers
+    state: stopped
+
+- name: Ensure that the Master IP address is out of the Kubernetes endpoint #This will only remove the IP from the enpoint in 3.2.1.9 and later
+  pause:
+    seconds: 5
+
+- name: Yum update docker
```
dak1n1, line 20: Will this actually update anything? I thought we were blocking docker updates in the yum config. I think the second update statement below is probably good enough. (line 33)

dak1n1:
Looks good so far. Though we might want to pause work on this until we can investigate using the openshift-ansible update playbook for our masters. I bet we can steal some of their roles to plug into this.

ansible/roles/openshift_master_upgrade/tasks/main.yml
robotmaxtron, line 20: Yes, if we don't explicitly install Docker 1.9.1 it'll pull in the latest, Docker 1.10 which has bitten me a couple times during upgrades.

detiber:
I'm assuming this role is going to be run serially. Having functionality like this embedded in a role scares me, since it would be easy to include this role in a play that isn't set for a serial run.

Other than that, after all of the masters have been updated, it will be necessary to reconcile the cluster policy, we do this in the product upgrade here: https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/upgrades/upgrade.yml#L112-L156

PR #1026, opened 2016-08-15
change logic to simply stop any haproxy process...

joelddiaz:
...that is older than the most recent haproxy process and has all connections in CLOSE_WAIT

### DIFF
--- a/scripts/monitoring/cron-haproxy-close-wait.py
+++ b/scripts/monitoring/cron-haproxy-close-wait.py
@@ -3,8 +3,9 @@
 
 # Disabling invalid-name because pylint doesn't like the naming convention we have.
 # pylint: disable=invalid-name
-''' Tool to detect haproxy processes in CLOSE_WAIT state, and stop ones
-    that have been running more than max_elapsed_time seconds '''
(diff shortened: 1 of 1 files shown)


scripts/monitoring/cron-haproxy-close-wait.py
```diff
@@ -49,34 +48,45 @@ def kill(self):
...
+        # that has connections in 'CLOSE-WAIT' state
+        kill_count = 0
         for proc in haproxy_procs_etimes.values():
             try:
+                kill = False
                 process = psutil.Process(proc)
-                self.dprint("Stopping PID: {}".format(process.pid))
-                process.kill()
+                for conn in process.connections():
+                    if conn.status == 'CLOSE_WAIT':
```
a13m: This logic still isn't right.  The characteristic we're looking for is a process which has _only_ connections (usually just one) in CLOSE_WAIT, and no other states.  Maybe it would be easier to invert it, i.e., pop everything with a connection _other_ than CLOSE_WAIT off of the list, and kill the rest.

a13m:
:+1: LGTM

PR #1027, opened 2016-08-15
change logic to simply stop any haproxy process...

joelddiaz:
...that is older than the most recent haproxy process and has all connections in CLOSE_WAIT


joelddiaz:
@a13m :+1: #1026 in stg

PR #1028, opened 2016-08-15
Changed role openshift_dedicated_admin to explicitly turn off non-primary master dedicated-admin-role services. Added oda_config param.

twiest:


joelddiaz:
:+1:

PR #1034, opened 2016-08-15
change link for existing DNS alerts

stenwt:


stenwt:
@zhiwliu reviewed #1030

PR #1041, opened 2016-08-16
sleep for a moment before getting newly create KMS data

joelddiaz:
while the KMS create command returns immediately, if you query for it immediately, it won't be found. so sleep for a moment after creating a new KMS key


mwoodson:
:+1:

PR #1044, opened 2016-08-17
Boolean variables now handled.

kwoodson:


kwoodson:
@mwoodson @joelddiaz :+1: in #1043

PR #1045, opened 2016-08-17
pass in cluster-wide KMS key to generated docker config

joelddiaz:


mwoodson:
wow, cool if this works! :) +1

PR #1050, opened 2016-08-18
added the openshift_volume_provisioner role

mwoodson:

PR #1051, opened 2016-08-18
added the openshift_volume_provisioner role

mwoodson:


mwoodson:
@kwoodson approved #1050

PR #1066, opened 2016-08-24
bind all of /etc/origin into host-monitoring

joelddiaz:
...so we can see/inspect x509 certificates


joelddiaz:
:+1: #1065 in stg

PR #1068, opened 2016-08-24
Minor fix for bool str.

kwoodson:


kwoodson:
@joelddiaz :+1: in #1067

PR #1074, opened 2016-08-24
Now supports in memory edits.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #1073

PR #1079, opened 2016-08-25
Initial commit

robotmaxtron:


ansible/roles/openshift_node_config/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+# tasks file for openshift_node_config
+- name: Edit OpenShift node config
+  lineinfile:
+    dest: /etc/sysconfig/atomic-openshift-node
+    line: "{{item.key }}={{ item.value }}"
```
joelddiaz, line 6: likely want a regex value passed so that we can catch the case where we want 'VAR=previous' set to 'VAR=new'

regex: "^{{ item.key }}="

robotmaxtron: @joelddiaz New regex is tested and working as expected but since the upgrade now throw a deprecation warning. 'Using bare variables is deprecated'.

joelddiaz: @robotmaxtron the line:

```
with_items: onc_openshift_node_config
```

should probably go to:

```
with_items: "{{ onc_openshift_node_config }}"
```

to get rid of the "bare variables" warning.

robotmaxtron: @joelddiaz Corrected, thank you.

PR #1090, opened 2016-08-29
added function to delete ES index after app create

stenwt:


stenwt:
@wshearn reviewed #1089

PR #1093, opened 2016-08-29
Passing in node specific labels.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #1092

PR #1094, opened 2016-08-29
add template/items/triggers for clusterwide config loop

joelddiaz:


kwoodson:
:+1:

ansible/roles/os_zabbix/vars/template_cluster.yml
```diff
@@ -0,0 +1,19 @@
+---
+g_template_cluster:
+  name: Template Cluster
```
twiest: Template OpenShift Cluster

twiest:
:+1:

PR #1010, opened 2016-08-10
Increase the interval for Zabbix checks to save on both DB load and space

rharrison10:

PR #1095, opened 2016-08-29
Increase the interval for Zabbix checks to save on both DB load and space

rharrison10:


rharrison10:
:+1: by @twiest on #1010

PR #1098, opened 2016-08-30
failed to check in generated start.sh

ivanhorvath:


ivanhorvath:
@joelddiaz :+1: in #1097

PR #1100, opened 2016-08-31
Use cache for ossh and oscp.

kwoodson:


kwoodson:
@joelsmith :+1:  in #1099

PR #1103, opened 2016-08-31
commented out the fluentd container usage config as it's causing issues on some boxes and it's not required.

twiest:


joelddiaz:
:+1:

PR #1109, opened 2016-09-01
[POC]  Send numerical metrics to Hawkular time-series database

yaacov:
**About**
Currently numerical metrics are sent to Zabbix. 
This PR is POC for sending _numerical_ data to Hawkular.

**Description**
Add a wrapper around the python hawkular client to mimic the Zabbix client.
Make an example cron-send-xxx that send numeric metrics to Hawkular server.

**Links**
https://github.com/hawkular/hawkular-client-python/pull/12

### DIFF
--- a/openshift_tools/monitoring/hawk_client.py
+++ b/openshift_tools/monitoring/hawk_client.py
@@ -5,30 +5,36 @@ HawkClient uses the openshift_tools.web/rest.py
 
 Example usage:
 
-
+    from openshift_tools.monitoring.metricmanager import UniqueMetric
     from openshift_tools.monitoring.hawk_common import HawkConnection
+    from openshift_tools.monitoring.hawk_client import HawkClient
 
-     ml = []
-     # UniqueMetric from openshift_tools.monitoring.metricmanager
-     new_metric = UniqueMetric('example.com','cpu/usage',12.34)
-     ml.append(new_metric)
-     new_metric = UniqueMetric('example.org','cpu/usage',4.2)
-     ml.append(new_metric)
+    ml = []
+    # UniqueMetric from openshift_tools.monitoring.metricmanager
+    new_metric = UniqueMetric('example.com','cpu/usage',12.34)
+    ml.append(new_metric)
+    new_metric = UniqueMetric('example.org','cpu/usage',4.2)
+    ml.append(new_metric)
 
-     # No Auth
-     zc = HawkClient(host='172.17.0.27:8000')
+    # No Auth
+    connection = HawkConnection(url='172.17.0.27:8000')
 
-     # Basic Auth
-     zc = HawkClient(host='172.17.0.27:8000', user='user', passwd='password')
+    # Basic Auth
+    connection = HawkConnection(url='172.17.0.27:8000', user='user', passwd='password')
 
-     print zc.add_metric(ml)
-
+    client = HawkClient(connection)
+    print client.add_metric(ml)
 """
 #These are not installed on the buildbot, disabling this
 #pylint: disable=no-name-in-module,unused-import
 from openshift_tools.monitoring.metricmanager import UniqueMetric, MetricManager
 from openshift_tools.monitoring.hawk_common import HawkConnection
-from hawkular.metrics import HawkularMetricsClient, MetricType, Availability
+
+# The hawkular client is optional
+try:
+    from hawkular.metrics import HawkularMetricsClient, MetricType, Availability
+except ImportError:
+    pass
 
 #This class implements rest calls. We only have one rest call implemented
 # add-metric.  More could be added here
@@ -42,6 +48,12 @@ class HawkClient(object):
         # pylint doesn't know where RestAPI is
         #pylint: disable=undefined-variable
         self.hawk_conn = hawk_connection
+        self.client = None
+
+        # Do not create a client if inactive
+        if self.hawk_conn.inactive:
+            return
+
         self.client = HawkularMetricsClient(host=self.hawk_conn.host,
                                             port=self.hawk_conn.port,
                                             scheme=self.hawk_conn.scheme,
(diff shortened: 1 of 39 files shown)


yaacov:
This is prof of concept for using hawkular metircs,

@simon3z @twiest @moolitayer please take a look.

**Convert cron scripts by import a mimic of ZaggSender**
When using hawkular we change:
`from openshift_tools.monitoring.zagg_sender import ZaggSender`
to:
`from openshift_tools.monitoring.hawk_sender import HawkSender as ZaggSender`
See:
https://github.com/yaacov/openshift-tools/blob/add-hawkular-sender/scripts/monitoring/cron-send-docker-timer-hawkular.py

**Use conventions to allow one big DB if needed**
metrics key is:
`key = '{0}/{1}/{2}'.format('ops', metric.host, metric.key)`
See:
https://github.com/yaacov/openshift-tools/blob/add-hawkular-sender/openshift_tools/monitoring/hawk_client.py

**Heartbeats are converted to availability metrics**

The algorithm used to parse the heartbeat object into hawkular key value pairs:

```
for each template:
    key = ops/heartbeat/template/:template-name, value = "up"
for each hostgroup:
    key = ops/heartbeat/hostgroup/:hostgroup-name, value = "up"
```

For example:

```
heartbeat = {"templates": ["t1"], "hostgroups": ["h1", "h2"]}

will send 3 availabilities to hawkular:
1. "ops/heartbeat/template/t1": "up"
2. "ops/heartbeat/hostgroup/h1": "up"
3. "ops/heartbeat/hostgroup/h2": "up"
```

yaacov:
@simon3z @burmanm :+1: using `Availability.Up` for the heartbeat now :-)

@twiest
I did not find a place where the ops team build the endpoints, where it will be easy to add tags for the UI ( e.g. units, graph-type, page, panel-in-page, line-color ... ) and for searching ( e.g. type, host, container ... )

scripts/monitoring/cron-send-docker-timer-hawkular.py
```diff
@@ -0,0 +1,49 @@
...
+#from openshift_tools.monitoring.zagg_sender import ZaggSender
+from openshift_tools.monitoring.hawk_sender import HawkSender as ZaggSender
+from openshift_tools.timeout import TimeoutException
+from openshift_tools.monitoring.dockerutil import DockerUtil
+import json
+import time
+
+if __name__ == "__main__":
+    keys = None
+    zs = ZaggSender()
```
simon3z: @yaacov why `ZaggSender`?

simon3z:
@yaacov I am not sure we want a brand new `cron-send-docker-timer-hawkular.py`. We don't want to collect the metrics twice, we probably want to extend the existing `cron-send-docker-metrics.py` to send the same metric (collected once) to the two systems (Zabbix and Hawkular).

scripts/monitoring/cron-send-docker-timer-hawkular.py
yaacov: I import HawkSender as ZaggSender
`from openshift_tools.monitoring.hawk_sender import HawkSender as ZaggSender`

I did not want to change the rest of the file ...

yaacov:
@simon3z now it's a copy of `cron-send-docker-metrics.py` that send data to `zabbix + hawkular`

lines added are 16, 25, 46, 51 to the file `cron-send-docker-metrics[-hawkular].py`

```
L16 from openshift_tools.monitoring.hawk_sender import HawkSender
L25 hs = HawkSender()
L46 hs.add_zabbix_keys(keys)
L51 hs.send_metrics()

```

simon3z:
> @simon3z now it's a copy of cron-send-docker-metrics.py that send data to zabbix + hawkular

@yaacov LBTM :+1:  (looks better to me :smile: ). @twiest who can review this?

twiest:
@yaacov   This is looking really good. Thank you for working on this and figuring this all out.

For the specific item definitions, they're here:
https://github.com/openshift/openshift-tools/blob/stg/ansible/roles/os_zabbix/vars/template_docker.yml#L10-L13

We have a separate process that adds the metrics to zabbix using these definitions.

If hawkular has a better way to do this, we're open to looking at that.

I've only skimmed this PR for now, but I'll do a proper review soon. Just wanted to let you know that this is on my todo list. :)

yaacov:
> If hawkular has a better way to do this, we're open to looking at that.

@twiest 
hawkualr except data without the need to declare them in advance, so you can just push them.

but then you loose the benefits of declaring meta-data that will help searching (e.g. host, pod, metric-group ... ) and hinting the view endpoint how to show the data ( e.g. legend, group with other metrics in same graph, group with other metrics in same page .... )

From looking at emplate_docker.yml I do not see meta-data that can help me build tags for search indexing and view hinting ( e.g. legend, units, graph grouping, graph type ... ) :eyeglasses: , do you have such data about the metrics you collect ?

yaacov:
@twiest we will probably need to parse  `template_docker.yml` to get the triggers and events for hawkular. but that will be a different PR @moolitayer ?

burmanm:
Looking at the file it seems the metadata has been mangled to a string already (losing its context) from what I could tell.

```
 - key: docker.storage.metadata.space.total
    applications:
    - Docker Storage
    value_type: float
```

It would make sense to split that back to describing metadata, such as

```
application: Docker Storage
metricName: space.total
metricGroup: storage
metricClass: docker
container_id: xxxx
pod_id: xxxx
pod_namespace: xxx
```

Or something alike (it's difficult to say by just looking at this one example how you bind them - but the idea is breaking it to meaningful parts). And luckily, even if you can't extract metadata yet, you can add that later to an existing metric (by pushing an update tags to the key).

alongoldboim:
@twiest @burmanm kinda need this pr to get merged for adding the docker version, rpms version, etc, to the inventory in Manageiq, Please have another look.
cc @kwoodson

yaacov:
Closing (split this PR to smaller parts)

[Add a hawkular configuration, client and sender](https://github.com/openshift/openshift-tools/pull/1178)
[Add a hawkular client command line tool](https://github.com/openshift/openshift-tools/pull/1179)

cc @simon3z @alongoldboim @twiest

PR #1113, opened 2016-09-01
use random project name, delete always at the end

stenwt:


stenwt:
@joelddiaz reviewed #1112

PR #1126, opened 2016-09-07
migrate to ansible2

joelddiaz:
depend on RPM ansible (not ansible1.9)
update ansible/simplezabbix.py to use Ansible 2's Python API

### DIFF
--- a/openshift_tools/ansible/simplezabbix.py
+++ b/openshift_tools/ansible/simplezabbix.py
@@ -1,7 +1,7 @@
 # vim: expandtab:tabstop=4:shiftwidth=4
 
 #
-#   Copyright 2015 Red Hat Inc.
+#   Copyright 2016 Red Hat Inc.
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


openshift_tools/ansible/simplezabbix.py
```diff
@@ -21,9 +21,22 @@
...
 """
 
-import ansible.runner
+from collections import namedtuple
+from ansible.parsing.dataloader import DataLoader
+from ansible.vars import VariableManager
+from ansible.inventory import Inventory
+from ansible.playbook.play import Play
+from ansible.executor.task_queue_manager import TaskQueueManager
+from ansible.plugins.callback import CallbackBase
```
detiber, line 30: These aren't considered supported APIs, you might be better to break this out to a callback plugin and a wrapper around ansible-playbook (which is exposed as a module in Ansible 2.x)

joelddiaz: @detiber are there some docs on this?

detiber: @joelddiaz I don't seem to be able to find any currently. I  know there is work being done to start documenting "supported" apis, but I'm not sure what the state of that is...

PR #1128, opened 2016-09-07
Setting config param to dict and allowing a config path to be sent separate.

kwoodson:


kwoodson:
@mwoodson :+1: in #1127

PR #1136, opened 2016-09-08
Ansible 2.2 new.

kwoodson:


kwoodson:
@joelddiaz :+1: in #1134

PR #1132, opened 2016-09-08
adding sharding based on first 2 character to zagg data files

ivanhorvath:
The two changes to None comparison I changed because pylint was crying about it.

PR #1138, opened 2016-09-08
adding sharding based on first 2 character to zagg data files

ivanhorvath:


ivanhorvath:
@kwoodson :+1: in #1132

PR #1141, opened 2016-09-09
changed kibana -> logs

mwoodson:


mwoodson:
@wshearn approved #1140

PR #1142, opened 2016-09-09
Fix broken root session logrotate

joelsmith:
The log rotate hack that does the work on session files requires
the touchfile /var/log/rootlog/sessions/logrotate_fakelog to be
present to trick logrotate into running the rotation script.

This card forgot to mention this step. https://trello.com/c/eJACONVg

Also remove trailing whitespace


a13m:
:+1:

PR #1143, opened 2016-09-09
Updating yedit for 2.2.

kwoodson:


joelddiaz:
:+1:

ansible/roles/lib_yaml_editor/build/ansible/yedit.py
```diff
@@ -32,6 +32,14 @@ def parse_value(inc_value, vtype=''):
         elif inc_value in false_bools:
             return False
 
+    # If type is unknown then go ahead and attempt to yaml load it.
+    if isinstance(inc_value, str):
```
detiber, line 36: Are these values coming from Ansible or from the API? If Ansible, a string might not be str, instead it could be AnsibleUnicode (as has caught us off guard recently). Ansible provides a AnsibleDumper class for serializing to yaml, but I haven't looked to see what they are doing for de-serializing from yaml.

There are also helper methods available in module_utils that will allow you to do parsing for booleans and interpreting some other data types.

PR #1144, opened 2016-09-09
Updating yedit for 2.2.

kwoodson:


kwoodson:
@joelddiaz :+1: in #1143

PR #1152, opened 2016-09-12
ansible 2.2 task naming

joelddiaz:
ansible says there's no such handler named "Restart the oso-rhel7-host-monitoring.service service"
simplify the naming of the handler to not include variable names


twiest:
:+1:

PR #1155, opened 2016-09-12
Yaml boolean string fix for 2.2.

kwoodson:


kwoodson:
@joelddiaz :+1: in #1154

PR #1157, opened 2016-09-12
gcp snapshot scripts.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #1156

PR #1159, opened 2016-09-12
Adding gcp files to spec.

kwoodson:


kwoodson:
@mwoodson :+1: in #1158

PR #1165, opened 2016-09-13
Make centos pass the sanity checks in generate container ansible scripts

yaacov:
Trying to run 
`openshift-tools/docker/build-local-setup-centos7.sh`
I got an error: `Unsupported build environment` because CentOS did not pass the sanity checks.
This PR make CentOS pass.


twiest:
ok to test

twiest:
:+1:

PR #1166, opened 2016-09-13
Installing gcp package into container.

kwoodson:


kwoodson:
@mwoodson :+1: in #1164

PR #1169, opened 2016-09-13
check for certs in /etc/origin/{master,node}

joelddiaz:
really just want to ignore /etc/origin/generated-configs


joelddiaz:
:+1: #1168 in stg

PR #1175, opened 2016-09-13
made edits to router/registry to use registry.ops

mwoodson:


mwoodson:
@dak1n1 approved #1174 in stg

PR #1178, opened 2016-09-14
Add a hawkular configuration, client and sender

yaacov:
**Description**
Add a wrapper code around the python hawkular client to mimic the Zabbix client.
This PR addes the library functions for sending numerical data to Hawkular.

**Only adds a library, does not do anything**

| New Classes |  |
| --- | --- |
| HawkConnection | Connection information (if inactive is True, the hawkular client will do nothing silently) |
| HawkClient | A client for hawkualr metrics API ( mimic ZaggClient, can be a drop in replacment ) |
| HawkSender | High level interface for sending metrics to hawkular ( mimic ZaggSender, can be a drop in replacment ) |

| Requirements |  |
| --- | --- |
| zbxsend | for UniqueMetric (If we drop zabbix, we will need to re-implement UniqueMetric) |
| hawkular-client-python | for low level hawkular API support |

### DIFF
--- a/openshift_tools/monitoring/hawk_client.py
+++ b/openshift_tools/monitoring/hawk_client.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python2
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+HawkClient uses the openshift_tools.web/rest.py
+
+Example usage:
+
+    from openshift_tools.monitoring.metricmanager import UniqueMetric
+    from openshift_tools.monitoring.hawk_common import HawkConnection
+    from openshift_tools.monitoring.hawk_client import HawkClient
+
+    ml = []
+    # UniqueMetric from openshift_tools.monitoring.metricmanager
+    new_metric = UniqueMetric('example.com','cpu/usage',12.34)
+    ml.append(new_metric)
+    new_metric = UniqueMetric('example.org','cpu/usage',4.2)
+    ml.append(new_metric)
+
+    # No Auth
+    connection = HawkConnection(url='172.17.0.27:8000')
+
+    # Basic Auth
+    connection = HawkConnection(url='172.17.0.27:8000', user='user', passwd='password')
+
+    client = HawkClient(connection)
+    print client.add_metric(ml)
+"""
+#These are not installed on the buildbot, disabling this
+#pylint: disable=no-name-in-module,unused-import,import-error
+from openshift_tools.monitoring.metricmanager import UniqueMetric, MetricManager
+from openshift_tools.monitoring.hawk_common import HawkConnection
+
+# The hawkular client
+from hawkular.metrics import HawkularMetricsClient, MetricType, Availability
+
+#This class implements rest calls. We only have one rest call implemented
+# add-metric.  More could be added here
+#pylint: disable=too-few-public-methods
+class HawkClient(object):
+    """
+    wrappers class around hawkular client python so use can use it with UniqueMetric
+    """
+
+    def __init__(self, hawk_connection):
+        # pylint doesn't know where RestAPI is
+        #pylint: disable=undefined-variable
+        self.hawk_conn = hawk_connection
+        self.client = None
+
+        # Do not create a client if inactive
+        if not self.hawk_conn.active:
+            return
+
+        self.client = HawkularMetricsClient(host=self.hawk_conn.host,
+                                            port=self.hawk_conn.port,
+                                            scheme=self.hawk_conn.scheme,
+                                            username=self.hawk_conn.username,
+                                            password=self.hawk_conn.password,
+                                            context=self.hawk_conn.context,
+                                            tenant_id=self.hawk_conn.tenant_id,
+                                           )
+
+    def add_metric(self, unique_metric_list):
+        """
+        Add a list of UniqueMetrics (unique_metric_list) via hawkular client
+        """
+        # Do not run if inactive
+        if not self.hawk_conn.active:
+            return
+
+        # We do not support zabbix heartbeat in hawkular
+        metric_list = [m for m in unique_metric_list if m.key != 'heartbeat']
+
+        for metric in metric_list:
+            # Hawkular metrics support any numeric values
+            value = metric.value
+            # Hawkular metrics use milliseconds
+            clock = metric.clock * 1000
+            # Add the type and host to the key, this is needed because in hawkular,
+            # these need to be namespaced otherwise multiple hosts will be adding
+            # values to the same key. See common metrics keys proposal.
+            _type = 'node'
+            _id = metric.host
+            key = '{0}/{1}/{2}'.format(_type, _id, metric.key)
+            # Use MetricType.Gauge for numeric metrics data
+            metric_type = MetricType.Gauge
+            self.client.push(metric_type, key, value, clock)
+
+        status, raw_respons = None, None
+        return (status, raw_respons)
--- a/openshift_tools/monitoring/hawk_common.py
+++ b/openshift_tools/monitoring/hawk_common.py
@@ -1,0 +1,38 @@
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+hawk_common contains common data structures and utils
+
+Example usage:
+    from openshift_tools.monitoring.hawk_common import HawkConnection, HawkHeartbeat
+
+    ZAGGCONN = HawkConnection(url='172.17.0.151', user='admin', password='pass')
+    ZAGGHEARTBEAT = HawkHeartbeat(templates=['template1', 'template2'], hostgroups=['hostgroup1', 'hostgroup2'])
+
+"""
+from collections import namedtuple
+from urlparse import urlparse
+import ssl
+
+# pylint: disable=too-few-public-methods,too-many-arguments,protected-access,too-many-instance-attributes
+# This is a DTO.  It needs to be a class because we need default values
+class HawkConnection(object):
+    ''' DTO for hawkular connection
+    '''
+    # pylint: disable=too-many-arguments
+    # This now supports ssl and need a couple of extra params
+    def __init__(self, url, user, password, ssl_verify=False, debug=False, active=True):
+        self.url = url if url.startswith('http') else 'http://' + url
+        self.username = user
+        self.password = password
+        self.ssl_verify = ssl_verify
+        self.debug = debug
+        self.active = active # We do not want this lib to work ansless explisitly activated
+
+        url_params = urlparse(self.url)
+        self.host = url_params.hostname
+        self.scheme = url_params.scheme
+        self.port = url_params.port or {'http': 80, 'https': 443}[self.scheme]
+        self.context = None if self.ssl_verify else ssl._create_unverified_context()
+        self.tenant_id = '_ops'
+
+HawkHeartbeat = namedtuple("HawkHeartbeat", ["templates", "hostgroups"])
--- a/openshift_tools/monitoring/hawk_sender.py
+++ b/openshift_tools/monitoring/hawk_sender.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+Collect metrics and send metrics to Hawk.  The data
+being send to Hawk is done using REST API using the HawkClient
+module
+
+Examples:
+
+    from openshift_tools.monitoring.hawk_common import HawkConnection, HawkHeartbeat
+    from openshift_tools.monitoring.hawk_sender import HawkSender
+    HOSTNAME = 'host.example.com'
+
+    ZAGGCONN = HawkConnection(url='https://172.17.0.151', user='admin', password='pass')
+    ZAGGHEARTBEAT = HawkHeartbeat(templates=['template1', 'template2'], hostgroups=['hostgroup1', 'hostgroup2'])
+
+    zs = HawkSender(host=HOSTNAME, hawk_connection=ZAGGCONN)
+    zs.add_heartbeat(ZAGGHEARTBEAT)
+    zs.add_zabbix_keys({ 'test.key' : '1' })
+    zs.send_metrics()
+"""
+
+from openshift_tools.monitoring.metricmanager import UniqueMetric
+from openshift_tools.monitoring.hawk_client import HawkClient
+from openshift_tools.monitoring.hawk_common import HawkConnection
+import os
+import yaml
+
+class HawkSenderException(Exception):
+    '''
+        HawkSenderException
+        Exists to propagate errors up from the api
+    '''
+    pass
+
+class HawkSender(object):
+    """
+    collect and create UniqueMetrics and send them to Hawk
+    """
+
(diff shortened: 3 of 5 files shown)


yaacov:
@alongoldboim @simon3z @twiest please review

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self, hawk_connection, headers=None):
+        # pylint doesn't know where RestAPI is
+        #pylint: disable=undefined-variable
+        self.hawk_conn = hawk_connection
+        self.client = None
+
+        # Do not create a client if inactive
+        if self.hawk_conn.inactive:
+            return
```
yaacov: If `connection` has `inactive=True` fails silently.

simon3z:
@burmanm anything you would like to extract from here into your `hawkular-client-python`? (cc @yaacov)

yaacov:
> anything you would like to extract from here into your hawkular-client-python

@simon3z @burmanm this PR is very specific to the ops team.

It adds:
1. A way to use a `yaml` files for setting default configurations. ( e.g. `/etc/openshift_tools/hawk_client.yaml`)
2. A way to fail silently if user only want to send to `zabbix`
3. A way to send `heartbeats` instead of `inventory`

I do not think this is useful for anyone not in the ops team.

yaacov:
@burmanm hi, while testing this in a clean machine I  `pip install hawkular-client` and things where missing ... :-)

Can you push the changes in hawkular-client-python to pip ?

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for template in heartbeat.get('templates'):
+            # Add the group and host to the key
+            key = '{0}/{1}/{2}'.format('ops/heartbeat', 'template', template)
+            self.client.push(metric_type, key, value, clock)
+
+        for hostgroup in heartbeat.get('hostgroups'):
+            # Add the group and host to the key
+            key = '{0}/{1}/{2}'.format('ops/heartbeat', 'hostgroup', hostgroup)
+            self.client.push(metric_type, key, value, clock)
```
alongoldboim: @simon3z @twiest  are you good with this key-formatting, would 'hostgroup' key be enough to cross link to the host?

yaacov: cc @moolitayer ^^

the alerts will have to use this keys ...

moolitayer: As I understand it In our end goal, instead of the host groups we would have each metric connected to an entity we know about: Node/Container/etc. 

With regards to hawkular I'm not sure if it makes more sense to concat keys or to use tags. Bare in mind we might have more then one key.

openshift_tools/monitoring/hawk_common.py
```diff
@@ -0,0 +1,38 @@
...
+    '''
+    # pylint: disable=too-many-arguments
+    # This now supports ssl and need a couple of extra params
+    def __init__(self, url, user, password, ssl_verify=False, debug=False, inactive=False):
+        self.url = url if url.startswith('http') else 'http://' + url
+        self.username = user
+        self.password = password
+        self.ssl_verify = ssl_verify
+        self.debug = debug
+        self.inactive = inactive # We do not want this lib to work ansless explisitly activated
```
yaacov: it's double negative, use active

yaacov: Fixed :-)

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,113 @@
...
+        value = Availability.Up
+        # Hawkular metrics use milliseconds
+        clock = metric.clock * 1000
+        # Use MetricType.Availability for heartbeat data
+        metric_type = MetricType.Availability
+
+        for template in heartbeat.get('templates'):
+            # Add the group and host to the key
+            key = '{0}/{1}/{2}'.format('ops/heartbeat', 'template', template)
+            self.client.push(metric_type, key, value, clock)
```
twiest: I don't think we should be pushing any template information into hawkular. Instead, we'll add tags to the items. So I think we can just do a single client.push here and not append template to the key name.

openshift_tools/monitoring/hawk_client.py
twiest: Same as templates. We don't want hostgroups on these keys. Templates and hostgroups are zabbix concepts that we won't be carrying over to hawkular. In zabbix those are ways to group metrics. In hawkular, we'll be using the far more powerful tags (we're anxious to use those as we see that as the biggest win over zabbix).

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,113 @@
...
+            # Add the group and host to the key
+            key = '{0}/{1}/{2}'.format('ops/heartbeat', 'hostgroup', hostgroup)
+            self.client.push(metric_type, key, value, clock)
+
+    def add_metric(self, unique_metric_list):
+        """
+        Add a list of UniqueMetrics (unique_metric_list) via hawkular client
+        """
+        # Do not run if inactive
+        if not self.client:
```
twiest: I'd prefer we do this check again as it makes it extremely clear what we're doing:

`if not self.hawk_conn.active:`

Then I don't think we need a comment as it makes sense by just reading the code.

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,113 @@
...
+            return
+
+        metric_list = [m for m in unique_metric_list if m.key != 'heartbeat']
+        heartbeat_list = [m for m in unique_metric_list if m.key == 'heartbeat']
+
+        for metric in metric_list:
+            # Hawkular metrics support only numeric values
+            value = float(metric.value)
```
twiest: Hmm, I don't like that we're forcing everything to be a float. Is the type not already in the metric?

Does hawkular support ints?   We really prefer to use ints as much as possible.

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,113 @@
...
+            # Hawkular metrics use milliseconds
+            clock = metric.clock * 1000
+            # Add the group and host to the key
+            key = '{0}/{1}/{2}'.format('ops', metric.host, metric.key)
```
twiest: Please add a comment to describe why this is needed.  Something simple like "This is needed because in hawkular, these need to be namespaced otherwise multiple hosts will be adding values to the same key."

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,113 @@
...
+            # Use MetricType.Gauge for metrics data
+            metric_type = MetricType.Gauge
```
twiest, line 86: This seems like it should be defined inside the metric, not forced to be Gauge here.

openshift_tools/monitoring/hawk_common.py
```diff
@@ -0,0 +1,38 @@
...
+        self.active = active # We do not want this lib to work ansless explisitly activated
+
+        url_params = urlparse(self.url)
+        self.host = url_params.hostname
+        self.scheme = url_params.scheme
+        self.port = url_params.port or {'http': 80, 'https': 443}[self.scheme]
+        self.context = context=None if self.ssl_verify else ssl._create_unverified_context()
+        self.tenant_id = '_ops'
+
+HawkHeartbeat = namedtuple("HawkHeartbeat", ["templates", "hostgroups"])
```
twiest, line 38: Like I said above, I don't think we want templates and hostgroups in hawkular. That's bringing over something we don't like from zabbix into hawkular which has a much better mechanism for grouping (tags).

openshift_tools/monitoring/hawk_sender.py
```diff
@@ -0,0 +1,190 @@
...
+"""
+Collect metrics and send metrics to Hawk.  The data
+being send to Hawk is done using REST API using the HawkClient
+module
+
+Examples:
+
+    from openshift_tools.monitoring.hawk_common import HawkConnection, HawkHeartbeat
+    from openshift_tools.monitoring.hawk_sender import HawkSender
+    HOSTNAME = 'use-tower1.ops.rhcloud.com'
```
twiest: Not sure where you got this hostname, but please use something else. Something like `host.example.com`.

openshift_tools/monitoring/hawk_sender.py
```diff
@@ -0,0 +1,190 @@
...
+        if not host:
+            host = self.host
+
+        hb_metric = UniqueMetric.create_heartbeat(host,
+                                                  heartbeat.templates,
+                                                  heartbeat.hostgroups,
+                                                 )
+        self.unique_metrics.append(hb_metric)
+
+    def add_zabbix_keys(self, zabbix_keys, host=None, synthetic=False):
```
twiest, line 139: We don't want any references to zabbix in hawk stuff.

Rename function `add_zabbix_keys` to something like `add_metric`.

Rename `zabbix_keys` to something like `metrics`

openshift_tools/monitoring/hawk_sender.py
```diff
@@ -0,0 +1,190 @@
...
+
+        for key, value in zabbix_keys.iteritems():
+            zabbix_metric = UniqueMetric(host, key, value)
+            zabbix_metrics.append(zabbix_metric)
+
+        self.unique_metrics += zabbix_metrics
+
+    # Allow for 6 arguments (including 'self')
+    # pylint: disable=too-many-arguments
+    def add_zabbix_dynamic_item(self, discovery_key, macro_string, macro_array, host=None, synthetic=False):
```
twiest, line 157: Dynamic items in zabbix are a way to send arbitrary data into zabbix in a flexible way.

For Hawk, all metrics are dynamic and flexible. So this shouldn't be needed at all. I'd remove it. I understand this has implications because we're trying to have a similar interface between the two, so maybe further discussion is necessary here.

This might just need to turn into something that creates normal hawkular metrics, but with extra tags. Yeah, more discussion is needed around this one.

At the very least, the name should be changed to not include `zabbix`. Suggested name `add_dynamic_metric`.

twiest (changes requested):
@yaacov This is looking VERY good. I left a few comments, but we can probably merge this once they've been addressed. Thanks!

twiest:
ok to test

openshift_tools/monitoring/hawk_client.py
yaacov: It need to be numeric float or Int 
I wanted to raise an exception if not numeric, but it's safe to remove it here if all data is numeric ( e.g. ints )
it will be converted to a string when sent using the Rest API :-)

openshift_tools/monitoring/hawk_sender.py
yaacov: It's a copy - paste thing -
https://github.com/openshift/openshift-tools/blob/stg/openshift_tools/monitoring/zagg_sender.py#L12

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,113 @@
...
+        # Hawkular metrics heartbeat value is "up"
+        value = Availability.Up
+        # Hawkular metrics use milliseconds
+        clock = metric.clock * 1000
+        # Use MetricType.Availability for heartbeat data
+        metric_type = MetricType.Availability
+
+        for template in heartbeat.get('templates'):
+            # Add the group and host to the key
+            key = '{0}/{1}/{2}'.format('ops/heartbeat', 'template', template)
```
simon3z: @yaacov I think @mwringe suggested `_ops` as prefix. I have the feeling that we want to constantize something here (at least the tenant `_ops`).

openshift_tools/monitoring/hawk_common.py
```diff
@@ -0,0 +1,38 @@
...
+        self.ssl_verify = ssl_verify
+        self.debug = debug
+        self.active = active # We do not want this lib to work ansless explisitly activated
+
+        url_params = urlparse(self.url)
+        self.host = url_params.hostname
+        self.scheme = url_params.scheme
+        self.port = url_params.port or {'http': 80, 'https': 443}[self.scheme]
+        self.context = context=None if self.ssl_verify else ssl._create_unverified_context()
+        self.tenant_id = '_ops'
```
simon3z, line 36: @yaacov why aren't you using this when you build the resource keys?

openshift_tools/monitoring/hawk_client.py
mwringe: I recommended "_ops" for the tenant name (mostly to make sure it doesn't conflict with metrics from the "_system" tenant, and anything with an "_" means that it can't be associated with an OpenShift project). For the metric id, I don't think it really matters what you pick, as long as its unique for each metric.

yaacov: @simon3z @mwringe 
a. does it make sense hawkular wise to make a different tenant for each node/pod/container metrics ?
b. if we create lots of tenants, what is a good convention regarding tenant names ?

yaacov: @moolitayer can the alerts we have handle multiple tenants ?

moolitayer: current code only looks on one tenant but that can change cc @cben
@yaacov can you explain based on what you looked at thus far what are ops vs system metrics?

yaacov: @simon3z @mwringe @moolitayer Will use the naming convention proposed in the common matrics naming draft.
**template**: `<type>/<id>/<group_id>`
**example**: 
`{ type=node, id=example.com, group_id=memory/usage }`
`key = node/example.com/memory/usage`

openshift_tools/monitoring/hawk_common.py
yaacov: @simon3z we will use here the common metrics key draft.

yaacov:
~~@twiest test is failing because it cant import hawkular-client,
https://github.com/hawkular/hawkular-client-python
how do I tell jenkins / pylint to ignore this import error ?~~

UPDATE: found the pylint disable ...

openshift_tools/monitoring/hawk_client.py
yaacov: > current code only looks on one tenant but that can change

@moolitayer this code is no longer in the PR ... I do not know why github did not fold this !?

burmanm: @yaacov Unless you need multitenancy, then no, there's no point in creating more tenants (it will make finding metrics more difficult as you can't request anything between multiple tenants)

yaacov: @burmanm Thanks !

moolitayer: @burmanm thanks.

 @mwringe would we need to collect tenant=ops and tenant=system metrics?
Do we already know if we will need to use tenancy for any kind of linking hawkular & cluster @simon3z?

simon3z: @moolitayer yeah tenancy may come into play when doing cross-linking.

yaacov:
close, continue at #1511

PR #1180, opened 2016-09-14
Add a Hawkular backend to the zabbix sender

yaacov:
**Description**
Currently the ops team is using only zabbix for metrics collection.
This PR adds a Hawkular back-end to the zabbix sender.

When the Hawkular is active any item sent to zabbix will also be sent to hawkular

**Caveat**
`add_zabbix_dynamic_item` is not implemented. `dynamic_item` will not be sent to hawkular.

Depends on #1178 
Default in hawk_client.yaml.example is incative=True

### DIFF
--- a/openshift_tools/monitoring/zagg_sender.py
+++ b/openshift_tools/monitoring/zagg_sender.py
@@ -27,6 +27,9 @@ import json
 import os
 import yaml
 
+# Add Hawkular sender backend
+from openshift_tools.monitoring.hawk_sender import HawkSender
+
 class ZaggSenderException(Exception):
     '''
         ZabbixSenderException
@@ -39,7 +42,7 @@ class ZaggSender(object):
     collect and create UniqueMetrics and send them to Zagg
     """
 
-    def __init__(self, host=None, zagg_connection=None, verbose=False, debug=False):
+    def __init__(self, host=None, zagg_connection=None, verbose=False, debug=False, hawk_connection=None):
         """
         set up the zagg client and unique_metrics
         """
@@ -57,6 +60,10 @@ class ZaggSender(object):
 
         self.host = host
         self.zaggclient = ZaggClient(zagg_connection=zagg_connection)
+
+        # Add Hawkular sender backend (inactive by default)
+        # When inactive will not try to send data to Hawkular.
+        self.hawk_sender = HawkSender(host, hawk_connection, verbose, debug)
 
     def print_unique_metrics_key_value(self):
         """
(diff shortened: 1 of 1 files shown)


yaacov:
@alongoldboim @simon3z @twiest please review

openshift_tools/monitoring/zagg_sender.py
```diff
@@ -27,6 +27,14 @@
 import os
 import yaml
 
+# Add Hawkular sender backend (Optional)
+HAS_HAWKULAR = False
+try
```
yaacov: missing :

openshift_tools/monitoring/zagg_sender.py
```diff
@@ -27,6 +27,14 @@
 import os
 import yaml
 
+# Add Hawkular sender backend (Optional)
```
simon3z: @yaacov why do we need this? I mean why this module is optional?

yaacov: @simon3z 
I understood from @twiest that they want hawkular to be _off_ by default, and everything to work without change if hawkular client is not installed, or the backend is not activated in the conf.yaml file.

p.s.
If we move to using hawkular why we need to send data to zabbix ?

simon3z: @yaacov enabling sending the data is different from the module dependency. I prefer that you pull in always this dependency (and yes sending data should be instead optional, default false).

simon3z: > p.s. If we move to using hawkular why we need to send data to zabbix ?

Because we can't replace zabbix without some time of overlapping (e.g. for feature parity, testing, etc.)

yaacov: @simon3z :+1: 
a. removed the optional module dependency. (ops will need to install hawkular-client on their systems)
b. sending data to hawkular is still off by default.

openshift_tools/monitoring/zagg_sender.py
```diff
@@ -27,6 +27,9 @@
 import os
 import yaml
 
+# Add Hawkular sender backend
+from openshift_tools.monitoring.hawk_sender import HawkSender
```
twiest, line 31: I really don't want zagg_sender to call hawk_sender.

I think we need a meta sender class, possible called `metric_sender` that then sends to both `zagg_sender` and `hawk_sender`.

I know that this will mean switching the collectors to call into `metric_sender` and that's a bigger change, but I think it's needed.

twiest (changes requested):

yaacov:
close, continue at #1511

PR #1184, opened 2016-09-14
updated oc_label to handle selectors and multiple labels at a time

mwoodson:


joelddiaz:
lgtm :+1: 

does the old oc_label test playbook need to be updated with these new module params?

PR #1185, opened 2016-09-14
raise severity on config loop trigger to high

joelddiaz:


joelddiaz:
:+1: #1167 in stg ( @a13m also agreed with this change )

PR #1186, opened 2016-09-14
updated oc_label to handle selectors and multiple labels at a time

mwoodson:


mwoodson:
@joelddiaz approved #1184

PR #1189, opened 2016-09-15
Adding cron and code fixes for snapshots.

kwoodson:


kwoodson:
@mwoodson :+1: in #1188

PR #1197, opened 2016-09-16
user/group should stay the same

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #1196

PR #1199, opened 2016-09-16
Fixing gcp snapshot scripts.

kwoodson:


kwoodson:
@mwoodson :+1: in #1198

PR #1202, opened 2016-09-17
Increase the number of trappers listening for data.

kwoodson:


kwoodson:
@ivanhorvath :+1:  in #1201

PR #1204, opened 2016-09-17
Bumping pollers to 20.

kwoodson:


kwoodson:
@ivanhorvath :+1: in #1203

PR #1213, opened 2016-09-20
Adding gcs monitoring.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #1211

PR #1223, opened 2016-09-21
Allow for "." as seprators.

kwoodson:


kwoodson:
@twiest, @joelddiaz, @mwoodson, @dak1n1 

This patch allows the user to specify a specific separator.  This means we can default our strings to '.' if we want but still support strings that have different characters in them by passing a "sep" parameter to the ansible module.

Let me know what you think.

dak1n1:
I think that sounds great. :+1:

twiest:
@kwoodson I love this. Get the build bot passing and I'll +1 it.

kwoodson:
@twiest, @dak1n1, @mwoodson, @joelddiaz 

This change would coincide with yedit updates across our code base.  We'll need to coordinate that work or at least make sure everyone is aware that it is going to happen.

PR #1224, opened 2016-09-22
add Terminating project check

BlueShells:
this will alert for these project which is in Terminating status and last longer than 5 minutes


drewandersonnz:
+1

SOP needs to be created though ..

PR #1226, opened 2016-09-22
Yedit fix for absent arrays.

kwoodson:


kwoodson:
@mwoodson :+1: in #1225

PR #1229, opened 2016-09-22
shuffle files around to make local development of multi_inventory easier

joelddiaz:
move gce.py and aws.py to a location where multi_inventoy.py can just work straight out of the git repository

update RPM spec to account for new location
clean up some unused/unpackaged files


twiest:
The build bot errors are from the vendored ec2.py and gce.py. 

:+1:

PR #1230, opened 2016-09-22
shuffle files around to make local multi_inventory development easier

joelddiaz:
move gce.py and aws.py to a location where multi_inventoy.py can just work straight out of the git repository

update RPM spec to account for new location
clean up some unused/unpackaged files


joelddiaz:
:+1: #1229 in stg
buildbot errors are on vendored ec2.py and gce.py files

PR #1232, opened 2016-09-22
cleaned up registry cert role

mwoodson:


mwoodson:
@joelddiaz approved #1231

PR #1239, opened 2016-09-22
zagg uses redis

ivanhorvath:


ivanhorvath:
approved by @kwoodson in #1237

PR #1240, opened 2016-09-22
change ohi/ossh/etc to just use cache by default

joelddiaz:
update cache using multi_inventory.py from now on


joelddiaz:
:+1: #1238 in stg

PR #1246, opened 2016-09-26
fixing redis persistence and heartbeater referencing non-existing var

ivanhorvath:


ivanhorvath:
@twiest approved in #1245

PR #1249, opened 2016-09-26
zagg web writes to wrong redis list

ivanhorvath:


ivanhorvath:
@kwoodson approved in #1248

PR #1250, opened 2016-09-26
Allow for "." as seprators.

kwoodson:


kwoodson:
@twiest @dak1n1 :+1:  in #1250

PR #1251, opened 2016-09-26
stop using unique project names till we can get a better handle on them

stenwt:


a13m:
@stenwt if we are going to stop using unique project names, we have to at least use unique app names, to avoid the previous failure case where a pod stuck terminating made the next app create fail.

PR #1256, opened 2016-09-26
Moving yedit changes to default to a period.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1255

PR #1258, opened 2016-09-26
add python-redis dependency

joelddiaz:


joelddiaz:
👍 @stenwt #1257 in stg

PR #1264, opened 2016-09-27
Fix yedit for ansible 2.2 and content_type: str.

kwoodson:


kwoodson:
@mwoodson :+1: in #1263

PR #1267, opened 2016-09-27
Adding a fix to router edits.  Default to an empty dict.

kwoodson:


kwoodson:
@mwoodson :+1: in #1266

PR #1270, opened 2016-09-27
update event watcher with regex ability

joelddiaz:
old event watcher only looked at the event 'reason'
to further be able to subcategorize events, allow passing a regex (and associated zabbix key to report the event under)


joelddiaz:
👍 #1265 in stg

PR #1272, opened 2016-09-27
basic info about redis

ivanhorvath:


ivanhorvath:
@joelddiaz approved in #1271

PR #1273, opened 2016-09-27
Router fixes for adding edits.  Updated yedit to allow for proper content loading.

kwoodson:

PR #1274, opened 2016-09-27
Router fixes for adding edits.  Updated yedit to allow for proper content loading.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1273

PR #1282, opened 2016-09-28
Cleaning up triggers from templates.

kwoodson:


joelsmith (approved):
:+1: I hope nothing is using these triggers, but I'm glad to have the trigger hooks disabled since the "secrets" are in a public repo

PR #1283, opened 2016-09-28
Cleaning up triggers from templates.

kwoodson:


kwoodson:
@joelsmith :+1:  in #1282

PR #1286, opened 2016-09-28
Scale the registry up to # of infra.

kwoodson:


kwoodson:
@mwoodson :+1: in #1285

PR #1290, opened 2016-09-28
increasing threshold numbers per discussion with twiest

ivanhorvath:


ivanhorvath:
@joelddiaz approved in #1289

PR #1292, opened 2016-09-28
Generate easily parsable list of AWS accounts

joelsmith:


joelsmith:
@kwoodson gave a :+1: in stg PR #1291

PR #1297, opened 2016-09-29
Add cluster capacity alerts in Zabbix

rharrison10:


rharrison10:
@joelddiaz :+1: on #1296

PR #1298, opened 2016-09-29
Add second cassandra node to deployments

rharrison10:


rharrison10:
@mwoodson :+1: on #1254

PR #1307, opened 2016-10-03
Updated lib_openshift to use communicate instead of wait.  This fixes deadlocking when buffers fill.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1306

PR #1313, opened 2016-10-04
Added logic to handle the 3.3 output as well as 3.2.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1312

PR #1315, opened 2016-10-05
changing file lock mutex

drewandersonnz:
this is causing a race condition, stopping the script from running at the same time as another very similar script


drewandersonnz:
+1 from @mmahut on https://github.com/openshift/openshift-tools/pull/1314

PR #1320, opened 2016-10-05
idempotent scheduable and empty list if no pods are found.

kwoodson:


kwoodson:
@mwoodson :+1: in #1319

PR #1324, opened 2016-10-06
migrate to ansible2

joelddiaz:
oso-ops-base install ansible 2.x
depend on RPM ansible (not ansible1.9)
update ansible/simplezabbix.py to use Ansible 2's Python API
change ops-zagg-heartbeat-processor to no use multiprocessing library (causing issues when Ansible does it's own multiprocessing)


joelddiaz:
👍 #1323 in stg

PR #1327, opened 2016-10-06
iam api key manager tool

rhdedgar:


scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+        parser.add_argument("-u", "--user",
+                            help="specify a username for the account")
+        args = parser.parse_args()
+
+        if not args.all and not args.profile:
+            print('Specify an account ID or profile name. \
+            To generate the keys for all ops accounts, use \"--all\"')
+            print('Usage:')
+            print('example: %s <account-id-number>' % parser.prog)
+            print('example: %s --all' % parser.prog)
```
blentz: Instead of multiple print statements, it would look cleaner to use a multiline string.

`print """  blah`
`blah %s`
`blah blah %s`
`""" % (foo, bar)`

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+            sys.exit(10)
+        else:
+            if args.user is None:
+                if getpass.getuser() != 'root' and os.getegid() < 1000:
```
blentz: Can this be a single if statement?

Also, you can just do:
`if not args.user`
the "is None" isn't necessary because None evaluates to False.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+                    args.user = getpass.getuser()
+            return args
+
+
+    @staticmethod
+    def check_accounts():
+        ''' retrieve the config-managed list of ops AWS accounts '''
+
+        path = '/etc/openshift_tools/aws_accounts.txt'
```
blentz: I'd consider moving this to either a class variable or to an externally configured file.

Having a hard-coded path buried deep in a class makes it tough to understand the program's requirements.

You do have it documented in the class docstring, which is good.  :smile:

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+        else:
+            if args.user is None:
+                if getpass.getuser() != 'root' and os.getegid() < 1000:
+                    args.user = getpass.getuser()
+            return args
+
+
+    @staticmethod
+    def check_accounts():
+        ''' retrieve the config-managed list of ops AWS accounts '''
```
blentz: The method docstring should note the return values for the method, particularly that this method raises an exception on an invalid path.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+                for line in stripped_line:
+                    if line is not None:
+                        accounts_list.append(line)
+                return accounts_list
+        else:
+            raise ValueError(path + ' does not exist')
+
+
+    def check_user(self, aws_account, user_name, client):
+        """ check if the user exists locally and in aws. creates iam user if not found """
```
blentz: Docstring is missing details about what the method returns. 

It looks like this method could return None, True, or a Response object. It might be worth commenting on under what conditions a user should expect those to be returned.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+                return True
+
+
+    @staticmethod
+    def create_user(aws_account, user_name, client):
+        """ create an iam user account """
+
+        response = client.create_user(
+            UserName=user_name
+            )
```
blentz: Minor nit-pick, but I'm pretty sure PyLint will say this is an incorrect continuation pattern.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+        client.add_user_to_group(GroupName='admin', UserName=user_name)
+        print('A new user account was added.\
+        Use change_iam_password.py -p %s to set your password' % aws_account)
+
+        return response
+
+
+    @staticmethod
+    def get_all_profiles():
+        """ if -a is specified, generate a list of all profiles found in ~/.aws/credentials """
```
blentz: Docstring should document the types returned by the method, including which exceptions it raises.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+
+        path = os.path.join(os.path.expanduser('~'), '.aws/credentials')
+        profile_list = []
+
+        if os.path.isfile(path):
+            with open(path) as open_file:
+                stripped_line = list([line.rstrip() for line in open_file.readlines()])
+                for line in stripped_line:
+                    account = re.match(r"^\[([A-Za-z0-9_\-]+)\]", line)
+                    if account is not None:
```
blentz: The more pythonic form is:

`if not account:`

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+                        profile_list.append(account.group(1))
+                return profile_list
+        else:
+            raise ValueError(path + ' does not exist')
+
+
+    @staticmethod
+    def get_keys(user_name, client):
+        """ get the Access Key IDs of any API keys the user has and return the oldest key"""
```
blentz: Why are we only returning the oldest key? Why not all keys or the newest key? Would this return a disabled key or does the API automatically filter those?

Some additional comments in the docstring on that might be helpful for future readers of this code.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+                keys_list.append(ekey['AccessKeyId'])
+
+        return keys_list
+
+
+    @staticmethod
+    def get_token(aws_account):
+        """ generate temporary SSO access credentials  """
+
+        sso_config_path = '/etc/openshift_tools/sso-config.yaml'
```
blentz: As above, this hard-coded path might be better as a class variable or in a config file.

Unlike the one above, this one isn't mentioned in the class docstring.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+        if all_keys:
+            for ekey in all_keys:
+                keys_list.append(ekey['AccessKeyId'])
+
+        return keys_list
+
+
+    @staticmethod
+    def get_token(aws_account):
+        """ generate temporary SSO access credentials  """
```
blentz: Docstring needs to document what the method returns, including exceptions that could be raised.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+
+        sso_config_path = '/etc/openshift_tools/sso-config.yaml'
+
+        if os.path.isfile(sso_config_path):
+            with open(sso_config_path, 'r') as sso_config:
+                yaml_config = yaml.load(sso_config)
```
blentz: yaml.load() can throw exceptions of its own if the file isn't right. 

Allowing those exceptions to break the script could be okay, but it could also be confusing to the user if they don't know how to fix it or if the error being thrown doesn't make it obvious what the problem is.  My suggestion is to think about it and decide if this is really how you want it to work.

scripts/aws_api_key_manager.py
```diff
@@ -0,0 +1,335 @@
...
+        else:
+            print('checked for file and it exists. no write was called. nothing to do here')
+            return True
+
+
+    @staticmethod
+    def write_credentials(aws_account, key_object):
+        ''' write the profile for the user account to the AWS credentials file '''
+
+        path = os.path.join(os.path.expanduser('~'), '.aws/credentials')
```
blentz: I suggest using os.environ('HOME') instead of expanduser('~'). It's clearer what the intent is.

blentz:
See my line comments.  Overall, I don't see any showstoppers. So, I'm a :+1: on this as-is.  

I do want you to review my comments and give them some consideration. :smile:

blentz (approved):
Some style considerations, but nothing major.  :+1:

PR #1330, opened 2016-10-06
Adding force option to oadm manage-node evacuate.

kwoodson:


kwoodson:
@mwoodson :+1: in #1329

PR #1331, opened 2016-10-06
added cron script to delete stuck projects for bz 1367432

dak1n1:
This is the result of stg testing. http://pastebin.test.redhat.com/418829

### DIFF
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -311,7 +311,7 @@ host_monitoring_cron:
 
 - name: Delete stuck projects for bz 1367432
   minute: "*/5"
-  job: /usr/bin/delete-stuck-projects
+  job: /usr/bin/delete-stuck-projects &> /var/log/delete-stuck-projects
 
 - name: "Prune builds/deployments/images"
   hour: "{{ osohm_pruning.cron.hour }}"
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -311,7 +311,7 @@ host_monitoring_cron:
 
 - name: Delete stuck projects for bz 1367432
   minute: "*/10"
-  job: /usr/bin/delete-stuck-projects &> /var/log/delete-stuck-projects
+  job: /usr/bin/delete-stuck-projects &>> /var/log/delete-stuck-projects
 
 - name: "Prune builds/deployments/images"
   hour: "{{ osohm_pruning.cron.hour }}"
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -310,7 +310,7 @@ host_monitoring_cron:
   job: ops-runner -f -s 60 -n ozc.send.synthetic.heartbeat.quick /usr/bin/ops-zagg-client -k heartbeat.ping -o 1 --synthetic
 
 - name: Delete stuck projects for bz 1367432
-  minute: "*/5"
+  minute: "*/10"
   job: /usr/bin/delete-stuck-projects &> /var/log/delete-stuck-projects
 
 - name: "Prune builds/deployments/images"
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -309,6 +309,10 @@ host_monitoring_cron:
   minute: "*/5"
   job: ops-runner -f -s 60 -n ozc.send.synthetic.heartbeat.quick /usr/bin/ops-zagg-client -k heartbeat.ping -o 1 --synthetic
 
+- name: Delete stuck projects for bz 1367432
+  minute: "*/5"
+  job: /usr/bin/delete-stuck-projects
+
 - name: "Prune builds/deployments/images"
   hour: "{{ osohm_pruning.cron.hour }}"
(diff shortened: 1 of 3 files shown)


scripts/monitoring/delete-stuck-projects.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+# Get a list of all projects
+oc get ns --template='{{ range .items }}{{.metadata.name}}{{ printf "\n" }}{{end}}' |sort > /tmp/namespaces
+
+# Compare with a list of all projects containing rolebindings
+oc get rolebindings --all-namespaces --template='{{ range .items }}{{.metadata.namespace}}{{ printf "\n" }}{{end}}' | sort -u > /tmp/namespaces_with_rolebindings
+
+# The diff between those are the stuck projects, to be removed on the next run of this script
+diff --changed-group-format='%>' --unchanged-group-format='' /tmp/namespaces_with_rolebindings /tmp/namespaces |grep -v -e kube-system -e default -e openshift-infra > /tmp/stuck_projects
```
abhgupta: exclude the `openshift` project from the stuck projects list as well

abhgupta: Instead of using `diff`, how about using `grep` instead? Seems cleaner in my opinion.

```
grep -Fxvf file1 file2
```

Source: http://stackoverflow.com/a/4079109

scripts/monitoring/delete-stuck-projects.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Look for any projects found during the last run of this script
+if [ -f /tmp/stuck_projects ]; then
+  projects_to_delete="$(cat /tmp/stuck_projects)"
+
+  # If the project still exists, is empty, and contains nothing, delete it.
+  for project in ${projects_to_delete}; do
+    oc get project ${project} &> /dev/null
+
+    if [ $? == 0 ] && [ -z "$(oc get all -n ${project})" ]; then
```
abhgupta, line 12: Instead of getting the project in the line above and then getting all resources within the project here in this condition, could we not simply just get all resources in the project (just that one call)? We can check if the exit code is 0 and if the result is empty, right?

scripts/monitoring/delete-stuck-projects.sh
```diff
@@ -0,0 +1,26 @@
...
+      oc get projects ${project} -o yaml
```
abhgupta, line 13: Where is this output stored on the machine?

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -309,6 +309,10 @@ host_monitoring_cron:
   minute: "*/5"
   job: ops-runner -f -s 60 -n ozc.send.synthetic.heartbeat.quick /usr/bin/ops-zagg-client -k heartbeat.ping -o 1 --synthetic
 
+- name: Delete stuck projects for bz 1367432
+  minute: "*/5"
```
abhgupta: I am assuming this means that we want it to run every 5 minutes. If so, I would suggest we run it every 10 minutes.

scripts/monitoring/delete-stuck-projects.sh
dak1n1: Because `oc get all -n $project` will return 0 regardless if a project exists or not. It just tells me if the project is empty. And apparently a non-existent project can be empty.

```
[root@dev-preview-stg-master-defb2 ~]# oc get all -n default &> /dev/null; echo $?
0
[root@dev-preview-stg-master-defb2 ~]# oc get all -n this_doesnt_exist &> /dev/null; echo $?
0
```

scripts/monitoring/delete-stuck-projects.sh
dak1n1: It will be stored in the cron log, but I should probably make a separate log file for it instead. I can make a change to the cron job for that part.

scripts/monitoring/delete-stuck-projects.sh
dak1n1: Nice! That is much cleaner.

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
dak1n1: If we run it every 10 minutes, a user could be unable to create projects for as long as 20 minutes, since it doesn't delete the project until the second run.

scripts/monitoring/delete-stuck-projects.sh
dak1n1: The logs are now saved in /var/log/delete-stuck-projects

robotmaxtron:
:+1:

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
abhgupta: I am mostly interested in ensuring that the cleanup delay is long enough to allow any timeouts to occur and not get in the way of the regular project/resource creation. Once the user is stuck, waiting for 10 minutes or 20 minutes won't make that big of a difference.

scripts/monitoring/delete-stuck-projects.sh
abhgupta: Makes sense - thanks for clarifying that

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
abhgupta: As a cron job that will be deleting projects without manual intervention, I want to err on the side of caution and avoid deleting any legitimate projects.

dak1n1: ok, changed to 10 minutes

abhgupta:
lgtm

PR #1333, opened 2016-10-06
Append creates array when it does not exist.

kwoodson:


kwoodson:
@joelddiaz :+1: in #1332

PR #1338, opened 2016-10-06
fixing zabbix redirect

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #1336

PR #1344, opened 2016-10-07
adding lrange to doc

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #1343

PR #1345, opened 2016-10-07
disable router and registry in the byo installer

mwoodson:


mwoodson:
@gburges approved #1342

PR #1347, opened 2016-10-07
Fixed deadlock issue in lib_git module.

kwoodson:


kwoodson:
@joelddiaz :+1: in #1346

PR #1349, opened 2016-10-09
[WIP] Adding collection of docker and openshift versions

alongoldboim:
This pr counts on having ops-metrics-client, which will send metric to Hawkular and Zabbix.

### DIFF
--- a/scripts/monitoring/cron-send-rpms-versions.sh
+++ b/scripts/monitoring/cron-send-rpms-versions.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Check OpenShift version
+OC_VERSION=$(oc version | sed -n '1{p;q}' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
+
+# Check Docker version
+DOCKER_VERSION=$(docker --version | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
+
+HOSTNAME=hostname --fqdn
+
+# Send results to zabbix
+ops-metrics-client -k "machine/$HOSTNAME/docker.version" -o "$DOCKER_VERSION"
+ops-metrics-client -k "machine/$HOSTNAME/oc.version" -o "$OC_VERSION"


scripts/monitoring/cron-send-rpms-versions.sh
```diff
@@ -0,0 +1,13 @@
...
+# Check OpenShift version
+OC_VERSION = oc version | sed -n '1{p;q}' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/'
+
+# Check Docker version
+DOCKER_VERSION = docker --version | sed 's/[^0-9.]*\([0-9.]*\).*/\1/'
+
+HOSTNAME = hostname --fqdn
+
+# Send results to zabbix
+ops-metrics-client -k "$HOSTNAME.docker-version" -o "$DOCKER_VERSION"
```
yaacov: The host name is added by the `ops-metrics-client` utility, for this line, the key `ops-metrics-client` creates will be: `/machine/$HOSTNAME/$HOSTNAME.docker-version`.
I think this line can be just:
`ops-metrics-client -k "docker.version" -o "$DOCKER_VERSION"`
and let the utility add the hostname.

scripts/monitoring/cron-send-rpms-versions.sh
```diff
@@ -0,0 +1,13 @@
...
+ops-metrics-client -k "$HOSTNAME.oc-version" -o "$OC_VERSION"
```
yaacov: I think the ops team like to use '.' when they separate words so the key is `docker.version` and not `docker-version`

This is important because we want to parse the key and use it to add missing tags (if they are not explicitly specified)
if you use `.` (ops notation) or `/` (heapster notation) we can parse it to tags `group=>'docker', metric=>'version'` , if you use `-` it will be   `group=>'docker-version', metric=>''` this is also ok ... since we do not know what to do with tags yet ...

scripts/monitoring/cron-send-rpms-versions.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Check OpenShift version
+OC_VERSION = oc version | sed -n '1{p;q}' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/'
```
yaacov: `OC_VERSION=$(oc version | sed -n '1{p;q}' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')` ?

yaacov: a. bash do not like space before `=`
b. bash will not return the return value of an expression unless you wrap it with `$(...)` or a tag

alongoldboim:
@zgalor have a look.

zgalor:
depends on https://github.com/openshift/openshift-tools/pull/1650

scripts/monitoring/cron-send-rpms-versions.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Check OpenShift version
+OC_VERSION=$(oc version | sed -n '1{p;q}' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
+
+# Check Docker version
+DOCKER_VERSION=$(docker --version | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
+
+HOSTNAME = hostname --fqdn
```
zgalor: Why is that required?

alongoldboim: This pr isn't final, the actual key will be "machine/$HOSTNAME/docker.version", thats why i'm collecting it, will add a fix soon.

alongoldboim:
@zgalor closing pr, take it over :)

PR #1350, opened 2016-10-10
from discussion with devs we are increasing the threshold of etcd con…

ivanhorvath:
…nection trigger


twiest:
:+1:  assuming the Devs passed off on these numbers (which I'm pretty sure they did).

ivanhorvath:
Dan McPherson, Jordan Liggitt and Tim St. Clair all agreed that this should be a safe set of numbers for now.

PR #1351, opened 2016-10-10
from discussion with devs we are increasing the threshold of etcd con…

ivanhorvath:
…nection trigger


ivanhorvath:
Approved by @twiest in #1350 , also see comment about devs agreeing with these numbers.

PR #1353, opened 2016-10-11
Fix aws user role to work with Ansible 2.2

joelsmith:
The complex module parameters member no longer exists. Instead,
loop over the creation results in parallel with the input params
that includes the generated password.


blentz:
:+1:  works for me.

mwoodson:
:+1:

Sounds like things changed in 2.2

PR #1358, opened 2016-10-12
Add legacy V2 AWS accounts to parsable list of AWS accounts

joelsmith:


joelsmith:
@drewandersonnz gave 👍 in stg PR #1357

PR #1359, opened 2016-10-12
almost-rebuild cron-send-create-app, phase 1

drewandersonnz:
- convert print to logging
- rewrite build progress conditions in main wait loop
- cleaner namespace
- unit testing style functions: setup / test / teardown
- try/except to ensure project can be deleted even if error
- enable verbose flag

This is the first stages of a multi-step rebuild.

Still to do, in new PRs:
- use library for oc and oadm commands
- redo build step main wait loop
- track specific repo commits for consistency in app builds
- test PV related app builds

### DIFF
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -216,7 +216,7 @@ host_monitoring_cron:
 {# Openshift Master checks #}
 - name: run create app every 5 minutes
   minute: "*/5"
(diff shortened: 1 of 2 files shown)


drewandersonnz:
ok, I need to do more pylint :-(

ivanhorvath:
I was just talking with Kenny earlier this week to maybe modify this script. Great that you got started on it. What we discussed and figured I should document here is that perhaps we need to think about not polling for the result, but watch oc events instead. Joel Diaz has an example of that in scripting/monitoring/cron-event-watcher.py if you are interested. Besides that what would be nice, and I see that you started down that road, is to notify zabbix about parts of the process. So build failure could send that it failed, but proceed with the test and create an app from an earlier build and see if that part completes, that way we would know from looking at zabbix which part is failing of the larger process. I of course started looking at this because being exposed to the app create failures for the first time was confused what exactly happened and what am I looking at when I'm checking the logs. Let me know if you want to chat some more about what we can do here.

dak1n1:
Looks good, but probably could use another set of eyes to verify.

PR #1363, opened 2016-10-12
Use admin.kubeconfig only for metrics checks.

wshearn:
The node config has too restricted permissions to access what we need.


wshearn:
failure unrelated.

ok to test

mwoodson (approved):

PR #1372, opened 2016-10-13
dry up some awsutil methods.

blentz:


blentz:
:+1: from @joelddiaz in #1316

PR #1367, opened 2016-10-12
added git_commit module to take files to commit argument

mwoodson:

PR #1373, opened 2016-10-13
git_commit fixes and upgrades

mwoodson:


mwoodson:
@kwoodson approved #1367

PR #1380, opened 2016-10-14
First attempt at a statuspage module.

kwoodson:


kwoodson:
@mwoodson :+1: in #1379

PR #1383, opened 2016-10-14
Add logging checks

wshearn:
The ocutil changes are so I can use oc exec and be able to easily parse the output. As seen on line 76 in cron-send-logging-checks.py

### DIFF
--- a/ansible/roles/os_zabbix/tasks/main.yml
+++ b/ansible/roles/os_zabbix/tasks/main.yml
@@ -66,7 +66,7 @@
   tags:
   - openshift_metrics
 
-- include_vars: template_openshift_metrics.yml
+- include_vars: template_openshift_logging.yml
   tags:
   - openshift_logging
 
@@ -259,4 +259,4 @@
     user: "{{ ozb_user }}"
     password: "{{ ozb_password }}"
   tags:
-  - openshift_metrics
+  - openshift_logging
--- a/ansible/roles/os_zabbix/tasks/main.yml
+++ b/ansible/roles/os_zabbix/tasks/main.yml
@@ -66,6 +66,10 @@
   tags:
   - openshift_metrics
 
+- include_vars: template_openshift_metrics.yml
+  tags:
+  - openshift_logging
+
 - name: Include Template OpenShift Cluster
   include: ../../lib_zabbix/tasks/create_template.yml
   vars:
@@ -241,6 +245,16 @@
   include: ../../lib_zabbix/tasks/create_template.yml
   vars:
     template: "{{ g_template_metrics }}"
+    server: "{{ ozb_server }}"
+    user: "{{ ozb_user }}"
+    password: "{{ ozb_password }}"
+  tags:
+  - openshift_metrics
+
+- name: Include Template OpenShift Logging
+  include: ../../lib_zabbix/tasks/create_template.yml
+  vars:
+    template: "{{ g_template_logging }}"
     server: "{{ ozb_server }}"
     user: "{{ ozb_user }}"
     password: "{{ ozb_password }}"
--- a/ansible/roles/os_zabbix/vars/template_openshift_logging.yml
+++ b/ansible/roles/os_zabbix/vars/template_openshift_logging.yml
@@ -60,3 +60,17 @@ g_template_logging:
     description: "elasticsearch pod health status"
     applications:
     - EFK
+  - discoveryrule_key: openshift.logging.elasticsarch.pods
+    name: "{% raw %}openshift.logging.elasticsarch.disk_used.{{ '{#' }}OSO_METRICS}{% endraw %}"
+    key: "{% raw %}openshift.logging.elasticsarch.disk_used[{{ '{#' }}OSO_METRICS}]{% endraw %}"
+    value_type: int
+    description: "elasticsearch pv disk usage"
+    applications:
+    - EFK
+  - discoveryrule_key: openshift.logging.elasticsarch.pods
+    name: "{% raw %}openshift.logging.elasticsarch.disk_free.{{ '{#' }}OSO_METRICS}{% endraw %}"
+    key: "{% raw %}openshift.logging.elasticsarch.disk_free[{{ '{#' }}OSO_METRICS}]{% endraw %}"
+    value_type: int
+    description: "elasticsearch pv free space"
+    applications:
+    - EFK
--- a/ansible/roles/os_zabbix/vars/template_openshift_logging.yml
+++ b/ansible/roles/os_zabbix/vars/template_openshift_logging.yml
@@ -1,0 +1,62 @@
+---
+g_template_logging:
+  name: Template OpenShift Logging
+
+  zitems:
+  - key: openshift.logging.fluentd.running
+    applications:
+    - EFK
+    value_type: int
+  - key: openshift.logging.fluentd.number_pods
+    applications:
+    - EFK
+    value_type: int
+  - key: openshift.logging.fluentd.node_mismatch
+    applications:
+    - EFK
+    value_type: int
+  - key: openshift.logging.fluentd.number_expected_pods
+    applications:
+    - EFK
+    value_type: int
+  - key: openshift.logging.kibana.site_up
+    applications:
+    - EFK
+    value_type: int
+  - key: openshift.logging.elasticsearch.single_master
+    applications:
+    - EFK
+    value_type: int
+  - key: openshift.logging.elasticsearch.all_numbers_registered
+    applications:
+    - EFK
+    value_type: int
+
+  ztriggers:
+  - name: "ElasticSearch is in split brain on {HOST.NAME}"
+    expression: "{Template OpenShift Logging:openshift.logging.elasticsearch.single_master}==0"
+    url: "https://github.com/openshift/ops-sop/blob/master/V3/Alerts/elasticsearch.asciidoc"
+    priority: high
+  - name: "FluentD/Node pod number mismatch on {HOST.NAME}"
+    expression: "{Template OpenShift Logging:openshift.logging.fluentd.number_pods}!={Template OpenShift Logging:openshift.logging.fluentd.number_expected_pods}"
+    url: "https://github.com/openshift/ops-sop/blob/master/V3/Alerts/fluentd.asciidoc"
+    priority: high
+  - name: "FluentD/Node mismatch on {HOST.NAME}"
+    expression: "{Template OpenShift Logging:openshift.logging.fluentd.node_mismatch}!=0"
+    url: "https://github.com/openshift/ops-sop/blob/master/V3/Alerts/fluentd.asciidoc"
+    priority: high
+
+  zdiscoveryrules:
+  - name: openshift.logging.elasticsarch.pods
+    key: openshift.logging.elasticsarch.pods
+    lifetime: 7
+    description: "Elasticsearch pods"
+ 
+  zitemprototypes:
+  - discoveryrule_key: openshift.logging.elasticsarch.pods
+    name: "{% raw %}openshift.logging.elasticsarch.pod_health.{{ '{#' }}OSO_METRICS}{% endraw %}"
+    key: "{% raw %}openshift.logging.elasticsarch.pod_health[{{ '{#' }}OSO_METRICS}]{% endraw %}"
+    value_type: int
+    description: "elasticsearch pod health status"
+    applications:
+    - EFK
--- a/scripts/monitoring/cron-send-logging-checks.py
+++ b/scripts/monitoring/cron-send-logging-checks.py
@@ -65,11 +65,6 @@ class OpenshiftLoggingStatus(object):
                 elif pod['metadata']['labels']['component'] == 'fluentd':
                     self.fluentd_pods.append(pod)
 
-    # Disabling all of these so we do not have to loop over pods again
-    # These will get their own proper setup once we have the sidecar pod
-    #pylint: disable=too-many-locals
-    #pylint: disable=too-many-branches
-    #pylint: disable=too-many-statements
     def check_elasticsearch(self):
         ''' Various checks for elasticsearch '''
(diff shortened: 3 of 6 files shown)


wshearn:
@kwoodson when you get a chance Monday can you review this?

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,224 @@
...
+        # Disable SSL to work around self signed clusters
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+
+        # Verify that the url is returning a valid response
+        kibana_status['site_up'] = 1
+        try:
+            res = urllib2.urlopen(kibana_url, context=ctx)
+            print res
```
portante: Could consider using `try/except/else` here and placing the `print res` under the else clause.

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,224 @@
...
+            node_matched = False
+            for pod in fluentd_pods:
+                if node['metadata']['labels']['kubernetes.io/hostname'] == pod['spec']['host']:
+                    node_matched = True
+
+            if node_matched == False:
+                fluentd_status['node_mismatch'] = 1
+
+        fluentd_status['running'] = 1
+        for pod in fluentd_pods:
```
portante: By the time this loop is done, we appear to have iterated the nodes twice, and the pods twice, and then iterated the fluentd pods for each node again.

Perhaps we can consider a way to avoid the number of iterations over the data.

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,224 @@
...
+        cluster_nodes = self.oc.run_user_cmd(node_cmd)['nodes']
+        es_status['all_nodes_registered'] = 1
+        # The internal ES node name is a random string we do not track anywhere
+        # pylint: disable=unused-variable
+        for node, data in cluster_nodes.items():
+        # pylint: enable=unused-variable
+            has_matched = False
+            for pod in es_pods:
+                if data['host'] == pod['metadata']['name']:
+                    has_matched = True
```
portante, line 106: Looks like you can exit the loop once you found a pod with a matching name.

portante (commented):
Looks good, fluentd check seems like it might not scale with high node/pod counts.

scripts/monitoring/cron-send-logging-checks.py
wshearn: Eh, we only care if the site returns a <500 status code. Though the print shouldn't have even slipped in.

portante (commented):
Looks good!

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,227 @@
...
+        for pod in self.es_pods:
+            pod_dc = pod['metadata']['labels']['deploymentconfig']
+            pod_name = pod['metadata']['name']
+            es_status['pods'][pod_dc] = {}
+
+
+            # exec into a pod and get cluster health
+            curl_cmd = "{} 'https://localhost:9200/_cluster/health?pretty=true'".format(self.es_curl)
+            cluster_health = "exec -ti {} -- {}".format(pod_name, curl_cmd)
+            health_res = self.oc.run_user_cmd(cluster_health)
```
twiest: I worry about the reliability of doing remote execution (oc exec) for health checks. AFAIK, this is the first time we've considered using it for health checks.

I feel the same way about remote execution based health checks as I do about remote autoheals. Basically, I think we should do them locally unless we absolutely can't.

Here are a few specifics:
- oc exec runs through the masters. If the masters are down, it's going to alert saying these are down as well.
- master load causing sporadic failures
- blue/green update causes sporadic failures (even though we should be going through the ELB, which should bleed connections, etc, it doesn't mean it'll be clean).

Is there a reason you can't do these checks from a sidecar container? I believe the card asked for a sidecar container as well.

What were the decision points on making this a remote health check?

ansible/roles/os_zabbix/vars/template_openshift_logging.yml
```diff
@@ -0,0 +1,62 @@
...
+    - EFK
+    value_type: int
+  - key: openshift.logging.elasticsearch.single_master
+    applications:
+    - EFK
+    value_type: int
+  - key: openshift.logging.elasticsearch.all_numbers_registered
+    applications:
+    - EFK
+    value_type: int
```
twiest, line 33: We need to also monitor the disk usage of the elastic search PV. This, IMO, is the most important thing to monitor as it's the thing I'm sure we'll run out of, possibly quickly.

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,227 @@
...
+        es_master_name = None
+        es_status['pods'] = {}
+        for pod in self.es_pods:
+            pod_dc = pod['metadata']['labels']['deploymentconfig']
+            pod_name = pod['metadata']['name']
+            es_status['pods'][pod_dc] = {}
+
+
+            # exec into a pod and get cluster health
+            curl_cmd = "{} 'https://localhost:9200/_cluster/health?pretty=true'".format(self.es_curl)
```
twiest: What happens if the oc exec succeeds, but the curl itself fails?

Like if elastic search is in a bad state or down for whatever reason?

From my cursory view, that seems like it would cause an ops-runner alert (as this script would back trace).  ops-runner should only alert if there's a bug in the script. The script itself should report typical failure cases.

twiest (changes requested):

scripts/monitoring/cron-send-logging-checks.py
wshearn: Even if we did the sidecar container it would still need to query the master to get the ES endpoints as we are running this across all the pods to make sure we are not in a split brain scenario.

As for why I went this way, it was the quickest way to get it up and running, and right now from when I need this done to when I could get the sidecar done there is not enough time.

ansible/roles/os_zabbix/vars/template_openshift_logging.yml
wshearn: I knew I forgot one. One PV check coming right up. (also how would checking the diskspace for the pods work with a sidecar container?)

scripts/monitoring/cron-send-logging-checks.py
wshearn: Good thinking, I will add a try/except around it and fail the check if it hits that.

ansible/roles/os_zabbix/vars/template_openshift_logging.yml
twiest: Since it's a PV, I believe PVs are acessible by all of the POD. I'm not 100% on that, though.

scripts/monitoring/cron-send-logging-checks.py
twiest: Doesn't elastic search know when it's in a split brain mode?

Each node in the elastic search cluster would get it's own sidecar container, so each would be monitored.

@a13m What are your thoughts on using remote `oc exec` calls for monitoring checks?

Does this worry you from a check reliability point of view?

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,231 @@
...
+        es_status['pods'] = {}
+        for pod in self.es_pods:
+            pod_dc = pod['metadata']['labels']['deploymentconfig']
+            pod_name = pod['metadata']['name']
+            es_status['pods'][pod_dc] = {}
+
+
+            # exec into a pod and get cluster health
+            try:
+                curl_cmd = "{} 'https://localhost:9200/_cluster/health?pretty=true'".format(self.es_curl)
```
twiest: Is this port not exposed on the POD IP?

In other words, could we do this curl directly from the master to the pod?

twiest (dismissed):

scripts/monitoring/cron-send-logging-checks.py
twiest: If we can do the curl from the master to the pod IP, then I'd also suggest changing this to pycurl (already inside of the host monitoring container), so that we have more control over the call as well as better error handling / messages.

wshearn: It is, but ssl certs are stored in a secret so it would still depend on the master being up.

scripts/monitoring/cron-send-logging-checks.py
wshearn: > Doesn't elastic search know when it's in a split brain mode?
> Sadly, not that I know of. When we ran into it with testing the only way we saw it was the nodes did not match across all ES instances.

a13m:
I certainly understand @twiest's concerns on this, but I think that OpenShift is designed for this method to work, and in the interest of time we should move forward with this.  I do agree that this may not be the right way to go in the long term, and we should watch it closely.

twiest:
@wshearn with @a13m's comments, I'll dismiss that review and start a new one.

ansible/roles/os_zabbix/vars/template_openshift_logging.yml
twiest: I think we need the disk check before merging this PR.

@wshearn do you disagree with this?

scripts/monitoring/cron-send-logging-checks.py
twiest: Understood, it would:
- get us away from doing an oc exec (which goes through the api server and might be flaky), at least for these types of calls
- would allow us more control over the check in general (able to use pycurl, able to check return codes, retries, etc)
- would allow for this to be refactored easier into a sidecar pod (obviously a future version where the secret containing the ssl certs would get mounted in by openshift).
- doesn't require curl to be installed on the pod

twiest (changes requested):

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,265 @@
...
+
+            # Exec into the pod and get diskspace, this will be cleaner once we
+            # have time to build a sidecar pod out of this.
+            try:
+                disk_used = 0
+                disk_free = 0
+                trash_var = 0
+
+                disk_output = self.oc.run_user_cmd("exec -ti logging-es-dy48r5sl-3-1po1n -- df").split(' ')
+                disk_output = filter(None, disk_output)
```
wshearn: @twiest any reason our bot does not like filter?

twiest: This is a pylint check. Here's more detail:
"Long story short, going forward you should prefer List comprehension instead of filter. pylint enforces this with this error message. – thefourtheye Feb 16 '14 at 3:08"

From: http://stackoverflow.com/questions/21806765/how-is-using-filter-a-bad-builtin

ansible/roles/os_zabbix/tasks/main.yml
```diff
@@ -246,3 +250,13 @@
...
+
+- name: Include Template OpenShift Logging
+  include: ../../lib_zabbix/tasks/create_template.yml
+  vars:
+    template: "{{ g_template_logging }}"
+    server: "{{ ozb_server }}"
+    user: "{{ ozb_user }}"
+    password: "{{ ozb_password }}"
+  tags:
+  - openshift_metrics
```
twiest: I think you mean `openshift_logging` for the tag.

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,265 @@
...
+                es_status['pods'][pod_dc]['elasticsearch_health'] = 0
+
+            # Exec into the pod and get diskspace, this will be cleaner once we
+            # have time to build a sidecar pod out of this.
+            try:
+                disk_used = 0
+                disk_free = 0
+                trash_var = 0
+
+                disk_output = self.oc.run_user_cmd("exec -ti logging-es-dy48r5sl-3-1po1n -- df").split(' ')
```
twiest: Isn't this part of the container name generated?  `dy48r5sl`

So, will that work in other clusters?

Will it even work in this cluster if you re-deploy the container?

For the curl above, you used `pod_name`. I'd expect something similar here.

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,265 @@
...
+                if pod['metadata']['labels']['component'] == 'es':
+                    self.es_pods.append(pod)
+                elif pod['metadata']['labels']['component'] == 'fluentd':
+                    self.fluentd_pods.append(pod)
+
+    # Disabling all of these so we do not have to loop over pods again
+    # These will get their own proper setup once we have the sidecar pod
+    #pylint: disable=too-many-locals
+    #pylint: disable=too-many-branches
+    #pylint: disable=too-many-statements
```
twiest: I'd prefer not to disable these. Instead, I'd prefer the code be refactored a little into smaller functions.

I understand what you're saying in the comment, but these specific pylint checks should never be disabled IMO.

Also, I don't know when we'll get to the sidecar container, so we'll likely be living with this code for some time.

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -0,0 +1,265 @@
...
+                    break
+
+            if has_matched == False:
+                es_status['all_nodes_registered'] = 0
+
+        return es_status
+
+    #pylint: enable=too-many-locals
+    #pylint: enable=too-many-branches
+    #pylint: enable=too-many-statements
```
twiest: Same comment as above. I'd prefer these not be disabled.

Also, whenever applying pylint disables, you need to justify why.

twiest (changes requested):

twiest:
:+1:

twiest (approved):

PR #1384, opened 2016-10-15
added playbook for in-place rpm hotfixes

dak1n1:


ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
```diff
@@ -0,0 +1,70 @@
...
+
+# ansible will return a '17' if a match is found. '-1' means 'not found'
+- name: Set update_successful=True if upgrade succeeded
+  set_fact: 
+    update_successful: True
+  when: "{{ post_upgrade.stdout.find(version_wanted) != -1 }}"
+
+- name: Upgrade failed, so try a downgrade
+  command: "yum -y downgrade atomic-openshift-{{ version_wanted }} atomic-openshift-clients-{{ version_wanted }} atomic-openshift-master-{{ version_wanted }} atomic-openshift-node-{{ version_wanted }} tuned-profiles-atomic-openshift-node-{{ version_wanted }} atomic-openshift-sdn-ovs-{{ version_wanted }}"
+  when: "{{ update_successful == False }}"
```
twiest, line 28: I'd prefer it if we had a separate "rollback.yml" playbook rather than automatically attempting to fix a failed upgrade cluster / host.

I prefer this because I want the rollback gated on a human making the decision of what's best. For instance, we might decide to just push forward, rather than roll back.

It will also allow us to run the rollback if, for instance, the upgrade succeeded _but_ there's some catastrophic bug that forces us to roll back to the previous version.

Also, here, you're specifying the version of the packages to downgrade to. And, it's the same version of the packages that we just tried to upgrade to.

That doesn't seem right. It should be the previous version, right?

Lastly, I'd prefer it if we used "yum history undo" as our rollback mechanism because:
1) it'll ensure we're back on the same rpm versions we were on before (we won't accidentally upgrade the cluster to a higher 3.2.1.x version than what it was before).
2) it'll ensure that any other RPMs that got updated during the attempted upgrade are also set back to their original versions (like maybe docker or anything else that are deps from these packages)

We'll need to ensure that we only run the "yum history undo" on the upgrade transaction (we wouldn't want to undo a security update, for instance).

twiest (changes requested):

ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
dak1n1: So the reason I like to have it attempt a downgrade is this case:

What if we have nodes with inconsistent versions? I tested this case with this script, and it gets all the nodes running the right version. The result is idempotent because you specify a version you want and it ensures that version gets installed one way or another.

Whereas running upgrade.yml would work on some, but not others, and running downgrade.yml would work on some but not others.

I like having it all-in-one like this, because the human who specifies the version they want will get that version. They don't have to worry about "do I need an upgrade, or a downgrade for this host?". They just request a version, and they get it. So that's what this particular task does, it just ensures the user gets the version they requested, in case the previous task failed.

twiest: Ah, I think I see what you're saying here.

The downgrade in this context is _NOT_ the rollback.

Instead, this is meant as "get this machine on the version I requested, whether that takes an upgrade -or- a downgrade".

Is that correct?

dak1n1: That's correct.

twiest: @dak1n1 ok, I like your logic on this. I researched it a little, and apparently the yum ansible module before ansible 1.9 allowed you do downgrade. But the current one doesn't allow it.

So, given that, what's the roll back plan then?

Is it re-run this, but with a different version (presumably with the current version installed)?

dak1n1: That's right. Before we start the upgrade, the person running the script should make note of the currently running (working) version. Then do the upgrade to the desired version. 

If we find we need a rollback at any time in the future (for example, if after a few days we find things are acting buggy), we can have a person run this script with the old version specified. That will put the cluster back onto the working version, whether that version is higher or lower than the one currently installed.

twiest: Ok, sounds good.

twiest (approved):

ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
jgouldin: can't we have the script output the currently running version before the change?

twiest:
:+1:  from me. We might want Matt to give a +1 as well.

ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
dak1n1: @jgouldin we can, yes, but it would still be up to the user to write down that version for future reference. (Since we want to be able to roll back days after the upgrade). 

There is actually a debug statement that says what the currently running version is for all the hosts. It says something like "version installed: ***_, version wanted: ***_*, upgrade needed? True/False". So that output is there for the user if they want to read it as part of the script run.

ansible/playbooks/adhoc/hotfix/roles/pre_update_node/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Choose a single master to run commands on
+  set_fact:
+    cluster_master: "{{ groups['oo_hosttype_master'] | intersect(groups['oo_clusterid_' ~ oo_clusterid]) | first }}"
+  run_once: true
+  when: "{{ update_needed == True }}" 
+
+- name: Make node unschedulable
+  command: "oadm manage-node {{ hostvars[oo_name].ec2_private_dns_name }} --schedulable=false"
```
mwoodson, line 10: consider using the oadm_manage_node module that is available.

Because of time constraints, consider adding a FIXME so this can get fixed/updated in future versions.

dak1n1: Neat, I didn't know we had a module for that. Ok, I'll add the FIXME.

ansible/playbooks/adhoc/hotfix/roles/update_node/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: Ensure node service is running
+  service:
+    name: atomic-openshift-node
+    state: running
+  when: "{{ update_needed == True }}" 
+  ignore_errors: True
+
+- name: Make node schedulable again
+  command: "oadm manage-node {{ hostvars[oo_name].ec2_private_dns_name }} --schedulable=true"
```
mwoodson, line 67: consider adding same FIXME as above

mwoodson (changes requested):

mwoodson (approved):

ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
jgouldin: I'd rather we not rely on someone writing it down.  Can we write a log file with the data and then read it later if we need to?

PR #1388, opened 2016-10-17
updated oc version with a few cleanup steps

mwoodson:


mwoodson:
@twiest approved #1387

PR #1392, opened 2016-10-17
added kube version to the metrics; cleaned up template installer

mwoodson:


mwoodson:
@wshearn approved #1391

PR #1395, opened 2016-10-18
pass the name when creating router template output

joelddiaz:
also, allow for replica=0 to be passed as a parameter
also, rename parameter latest-image to latest-images


joelddiaz:
👍 #1393 in stg

PR #1396, opened 2016-10-18
Fixed a couple bugs found by the team

dak1n1:
- The wrong version was being pulled in by dependencies during the install, causing the upgrade step to fail.
- The services were not being restarted, due to the wrong var name being used in the 'when' statement.

### DIFF
--- a/ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
+++ b/ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
@@ -63,5 +63,4 @@
   - atomic-openshift-master-api
   - atomic-openshift-master-controllers
   when: "{{ update_needed == True }}" 
-  ignore_errors: True
 
--- a/ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
+++ b/ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
@@ -6,10 +6,8 @@
     downgrade_successful: False
   when: "{{ update_needed == True }}" 
 
-- name: Update atomic-openshift-master
-  yum:
-    name: "{{ rpm_wanted }}"
-    state: present
+- name: Upgrade master
+  command: "yum -y upgrade atomic-openshift-{{ version_wanted }} atomic-openshift-clients-{{ version_wanted }} atomic-openshift-master-{{ version_wanted }} atomic-openshift-node-{{ version_wanted }} tuned-profiles-atomic-openshift-node-{{ version_wanted }} atomic-openshift-sdn-ovs-{{ version_wanted }}"
   when: "{{ update_needed == True }}" 
 
 - name: Check if update is successful
@@ -36,6 +34,7 @@
 - name: Set downgrade_successful=True if downgrade succeeded
   set_fact: 
     downgrade_successful: True
+    upgrade_successful: True
   when: "{{ post_downgrade.stdout.find(version_wanted) != -1 }}"
 
 - name: Alert if update is not successful, but continue
@@ -56,7 +55,7 @@
   with_items:
   - atomic-openshift-master-api
   - atomic-openshift-master-controllers
-  when: "{{ update_needed == True and update_successful == True }}" 
+  when: "{{ update_needed == True and upgrade_successful == True }}"
 
 - name: Ensure master services are running
   service:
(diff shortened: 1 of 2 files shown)


ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
```diff
@@ -36,6 +34,7 @@
 - name: Set downgrade_successful=True if downgrade succeeded
   set_fact: 
     downgrade_successful: True
+    upgrade_successful: True
```
twiest: "upgrade_successful" is confusing in the downgrade case.

Consider calling it something like "wanted_version_installed" or "operation_successful". Something that can cover both update and downgrade.

ansible/playbooks/adhoc/hotfix/roles/update_node/tasks/main.yml
```diff
@@ -53,7 +51,7 @@
   service:
     name: atomic-openshift-node
     state: restarted
-  when: "{{ update_needed == True and update_successful == True }}" 
+  when: "{{ update_needed == True and upgrade_successful == True }}"
```
twiest: Same, change upgrade_successful to something that covers update and downgrade better.

twiest (changes requested):

ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
dak1n1: Thanks, I like that a lot better. It helps avoid bugs like the previous one where I had the names confused.

ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
```diff
@@ -56,7 +53,7 @@
   with_items:
   - atomic-openshift-master-api
   - atomic-openshift-master-controllers
-  when: "{{ update_needed == True and update_successful == True }}" 
+  when: "{{ update_needed == True and wanted_version_installed == True }}"
 
 - name: Ensure master services are running
   service:
```
twiest, line 59: Why are we ignoring errors here?

ansible/playbooks/adhoc/hotfix/roles/update_node/tasks/main.yml
```diff
@@ -18,42 +15,41 @@
...
 
 - name: Restart atomic-openshift-node service if upgrade was successful
   service:
     name: atomic-openshift-node
     state: restarted
-  when: "{{ update_needed == True and update_successful == True }}" 
+  when: "{{ update_needed == True and wanted_version_installed == True }}"
 
 - name: Ensure node service is running
   service:
```
twiest, line 55: Same question, why ignore errors here?   Also on the oadm call below.

twiest (changes requested):
@dak1n1 This is looking really good. Just left a couple of comments that I noticed while reviewing the other changes.

ansible/playbooks/adhoc/hotfix/roles/update_master/tasks/main.yml
dak1n1: My thought was that if a master service failed to restart, I wanted to continue on to update the rest of the hosts, and add that master back to the ELB.

In retrospect, that doesn't make much sense for masters. I can remove that.

I think it makes more sense for nodes, where you might want to continue upgrading the rest of the nodes even though one node-restart failed.

ansible/playbooks/adhoc/hotfix/roles/update_node/tasks/main.yml
dak1n1: I think `ignore_errors` makes the most sense on the `oadm` command. That command will retry 10 times if it fails initially, so `ignore_errors` allows it to retry without causing the whole playbook to terminate. 

The reason I have so many retry attempts is that, in the past, Preview has been very busy and these calls would fail sometimes. But it's been working a lot better lately.

In the case of the nodes, I thought we might want to continue even if one node-restart fails. But I'm really fine either way. I can remove it, or keep it. What do you think?

twiest: So how Ansible works is that each host specified in a playbook executes separately.

So failure of 1 node won't fail the other nodes. They will each execute independently.

So I'd definitely prefer we don't ignore_errors here.

As far as oadm, that makes more sense. I'd prefer it fail overall if it isn't able to set scheduling=true, but this is likely fine.

twiest (approved):

twiest:
:+1:

PR #1375, opened 2016-10-13
Add OpenShift Online single sign-on SAML container

joelsmith:
This container runs a preconfigured version of simpleSAMLphp with PHP and Apache HTTPD. This initial revision is geared towards SSO to Amazon AWS accounts that have been configured to use the service provided by this container as a SAML Identity Provider (IdP).

PR #1397, opened 2016-10-18
Oso saml sso

joelsmith:


joelsmith:
@twiest gave 👍 in stg PR #1375

PR #1411, opened 2016-10-19
fixed the registry options in oadm_registry

mwoodson:


mwoodson:
@kwoodson approved #1410

PR #1412, opened 2016-10-19
Fixed a couple bugs found by the team

dak1n1:
- The wrong version was being pulled in by dependencies during the install, causing the upgrade step to fail.
- The services were not being restarted, due to the wrong var name being used in the `when` statement.
- Removed unnecessary `ignore_errors` statements.


dak1n1:
Approved in PR #1396

PR #1413, opened 2016-10-20
fix bug for Terminating project last for -1 days

BlueShells:
found that sometimes the project in Terminating project have a time early than create, that's why we keep seeing alert said that some project last for 70k s

### DIFF
--- a/scripts/monitoring/cron-send-project-stats.py
+++ b/scripts/monitoring/cron-send-project-stats.py
@@ -253,9 +253,11 @@ def main():
                 current_time = datetime.datetime.now()
                 time_keeps = current_time - old_time
                 print 'the project in Terminating status for %s' % time_keeps
-
-                time_keeps_max = max(time_keeps_max, (current_time - old_time).seconds)
-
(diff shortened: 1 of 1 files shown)


scripts/monitoring/cron-send-project-stats.py
```diff
@@ -253,9 +253,11 @@ def main():
                 current_time = datetime.datetime.now()
                 time_keeps = current_time - old_time
                 print 'the project in Terminating status for %s' % time_keeps
-
-                time_keeps_max = max(time_keeps_max, (current_time - old_time).seconds)
-
+                if current_time > old_time:
+                    time_keeps_max = max(time_keeps_max, (current_time - old_time).seconds)
```
drewandersonnz, line 256: Why (current_time - old_time).seconds ? When line 254 is the same claulation?

Why not change to time_keeps.seconds?

drewandersonnz: Looking further, time_keeps isn't used anywhere. Remove that line instead :-)

PR #1414, opened 2016-10-20
reduce scale of script, use library

drewandersonnz:
- uses library for OCUtil
- logging instead of print, including verbosity
- redundant script args removed
- redundant functions removed
- testProjects requires no external data so can be unit tested

### DIFF
--- a/openshift_tools/monitoring/ocutil.py
+++ b/openshift_tools/monitoring/ocutil.py
@@ -126,6 +126,14 @@ class OCUtil(object):
 
         return pods_yaml
 
+    def get_projects(self):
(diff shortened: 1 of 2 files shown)


drewandersonnz:
```
I: 23, 0: Locally disabling import-error (F0401) (locally-disabled)
F: 18, 0: Unable to import 'openshift_tools.monitoring.ocutil' (import-error)
```

Not sure why this error exists, I'm happy to commit without tests passing :-)

drewandersonnz:
Fail is accessing a private class member function, I don't yet know how to add this to the library

BlueShells:
:+1:

PR #1421, opened 2016-10-20
attempt at idempotent process.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1419

PR #1422, opened 2016-10-20
Add autogenerated versions of SAML2 SSO container

joelsmith:


joelsmith:
@twiest gave 👍 in stg PR #1420

PR #1424, opened 2016-10-21
add support for setting up multiple OpenShift routers

joelddiaz:
and fix up some incorrect ansible module parameters
create routers explicitly by the name passed into the ansible module
redefine the edits section to allow for more precise yaml editing


joelddiaz:
👍 #1418 in stg

PR #1426, opened 2016-10-21
reducing priority on config run, there is no reason for this to page

ivanhorvath:


ivanhorvath:
Approved by @robotmaxtron and @a13m in #1425

PR #1427, opened 2016-10-21
Separating accounts.

kwoodson:


kwoodson:
@mwoodson @twiest, @joelsmith 

Here are my latest changes for multi_inventory.  It now supports --account or --cluster with --refresh-cache.  I ran some tests and it appears to be working as intended without any real performance decrease.

What would you expect if the cluster or account is not found?  The current behavior will just pull from cache.  Would you prefer an error?

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,137 @@
...
+
+class MultiInventoryAccount(object):
+    """Object to represent an account"""
+    def __init__(self, name, config, cache_path, cache_age=1800):
+        self.name = name
+        self.config = config
+        self._account_data = None
+
+        # set to global cache_age
+        self.cache_age = cache_age
```
twiest: Hmm, `cache_age` feels like it's the actual age of the current cache.

But I think you're saying that this is the maximum age allowed of the cache.

Consider changing this to something like `cache_max_age`. That's what it's called down below in the MultiInventory class.

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,137 @@
...
+
+        return self._account_data
+
+
+    @account_data.setter
+    def account_data(self, data):
+        """setter for account_data"""
+        self._account_data = data
+
+    def run_provider(self, force=False, valid_cache=False):
```
twiest: `valid_cache` feels like it's asking if the cache is actually valid or not.

But here, I think it means if the end api user wants the cache validated.

Consider renaming this variable to `validate_cache`.

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,137 @@
...
+        """Setup the provider call with proper variables and call self.get_provider_tags
+
+           force:  Whether to force an update on the account
+           valid_cache:  Whether or not to validate if is_cache_valid.
+                         This happens in case of an account update
+
+        """
+        # force an update??
+        if not force:
+            # check if we want a valid cache or if the cache is valid
```
twiest: Consider changing this to `# check if the user cares if the cache is up to date and if they do if the cache is actually up to date.`

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,137 @@
...
+        return {'cached': False,
+                'proc': subprocess.Popen(cmds, stderr=subprocess.PIPE, stdout=subprocess.PIPE, env=env)}
+
+    def generate_config(self):
+        """Generate the provider_files in a temporary directory"""
+        prefix = 'multi_inventory.'
+        tmp_dir_path = tempfile.mkdtemp(prefix=prefix)
+        atexit.register(MultiInventoryUtils.cleanup, [tmp_dir_path])
+        for provider_file in self.config['provider_files']:
+            filedes = open(os.path.join(tmp_dir_path, provider_file['name']), 'w+')
```
twiest: I think this will fail to close the file descriptor if there's an exception in the call to Template / substitute or in writing to the file.

I understand that these are temp files, so probably not the end of the world, but still good practice.

Consider using `with` like this:

``` python
with open(os.path.join(tmp_dir_path, provider_file['name']), 'w+') as filedes:
    content = Template(provider_file['contents']).substitute(tmpdir=tmp_dir_path)
    filedes.write(content)
```

ansible/inventory/multi_inventory.py
```diff
@@ -37,33 +145,57 @@ class MultiInventory(object):
...
+           else the credentials are loaded from multi_inventory.yaml or from the env
+           and we attempt to get the inventory from the provider specified.
+        '''
+
+        # Finish loading configuration files
+        self.load_config_settings()
+
+        if self.args.get('refresh_cache', None) or \
+           not MultiInventoryUtils.is_cache_valid(self.cache_path, self.cache_max_age):
+            self.get_inventory()
```
twiest: Hmm, it feels awkward to not have a return value for get_inventory.

Consider using a return variable instead of `self.result` here.

ansible/inventory/multi_inventory.py
```diff
@@ -119,144 +227,103 @@ def load_yaml_config(self, conf_file=None):
...
-                # Update env vars after creating provider_config_files
-                # so that we can grab the tmp_dir if it exists
-                env = account.get('env_vars', {})
-                if env and tmp_dir:
-                    for key, value in env.items():
-                        env[key] = Template(value).substitute(tmpdir=tmp_dir)
-
-                provider = account['provider']
-                processes[acc_name] = self.get_provider_tags(provider, env)
+            for acc in tmp_accounts:
```
twiest: If there are multiple accounts with the same name, shouldn't we just error instead of looping?

ansible/inventory/multi_inventory.py
```diff
@@ -119,144 +227,103 @@ def load_yaml_config(self, conf_file=None):
...
+                    if proc.returncode != 0:
+                        err_msg = ['\nProblem fetching account: %s' % acc.name,
+                                   'Error Code: %s' % proc.returncode,
+                                   'StdErr: %s' % err,
+                                   'Stdout: %s' % out,
+                                  ]
+                        raise RuntimeError('\n'.join(err_msg))
+
+                    acc.account_data = json.loads(out)
+                    acc.write_to_cache()
```
twiest: This section feels like it should be part of the account class.

Is there a reason to run it outside of the account class?

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,137 @@
...
+
+        # collect a list of clusters for this account
+        self.clusters = []
+        if self.config.has_key('cluster_vars') and \
+            self.config['cluster_vars'].has_key('clusters'):
+            self.clusters = self.config['cluster_vars']['clusters'].keys()
+
+
+    @property
+    def account_data(self):
```
twiest: `account_data` feels like it's the config data.

Consider renaming this to `account_inventory`.

ansible/inventory/multi_inventory.py
```diff
@@ -119,144 +227,103 @@ def load_yaml_config(self, conf_file=None):
...
+                                  ]
+                        raise RuntimeError('\n'.join(err_msg))
+
+                    acc.account_data = json.loads(out)
+                    acc.write_to_cache()
+        except ValueError as exc:
+            print exc.message
+            sys.exit(1)
+
+    def load_account_data(self):
```
twiest: Consider renaming this to `load_account_inventory`.

ansible/inventory/multi_inventory.py
```diff
@@ -119,144 +227,103 @@ def load_yaml_config(self, conf_file=None):
...
-            filedes.close()
-
-        return tmp_dir_path
-
-    def run_provider(self):
-        '''Setup the provider call with proper variables
-           and call self.get_provider_tags.
-        '''
+    def run_account_providers(self, force=False, valid_cache=True, account=None):
+        """Setup the provider call with proper variables and call self.get_provider_tags"""
```
twiest: I don't see how you're handling the case where `account == None`.

What is expected to happen in that case?

`account` can be `None` if:
- They fail to pass it into this function
- A clusterid is specified, but doesn't exist in an account
- probably others

ansible/inventory/multi_inventory.py
```diff
@@ -375,122 +442,140 @@ def apply_group_selectors(self, inventory, group_selectors):
                         else:
                             inventory['_meta']['hostvars'][host][selector['name']] = False
 
-    def apply_account_config(self, acc_name, acc_config):
-        ''' Apply account config settings '''
-        results = self.all_inventory_results[acc_name]
+    def apply_account_config(self, acc):
```
twiest: This also feels like it should be part of the account class.

ansible/inventory/multi_inventory.py
```diff
@@ -375,122 +442,140 @@ def apply_group_selectors(self, inventory, group_selectors):
...
+                            help='Refresh an account')
+        parser.add_argument('--debug', action='store_true', default=False,
+                            help='Wether to print debug')
         self.args = parser.parse_args().__dict__
 
-    def write_to_cache(self):
-        ''' Writes data in JSON format to a file '''
+
+    @classmethod
+    def json_format_dict(cls, data, pretty=False):
```
twiest: This feels like it should be part of `MultiInventoryUtils`

twiest (changes requested):

ansible/inventory/multi_inventory.py
kwoodson: Moved under utils.

ansible/inventory/multi_inventory.py
kwoodson: Moved under account.

ansible/inventory/multi_inventory.py
kwoodson: Removed.

ansible/inventory/multi_inventory.py
kwoodson: Created a staticmethod and moved under the account class.

ansible/inventory/multi_inventory.py
kwoodson: I added error handling if you pass a cluster or account that does not exist.

ansible/inventory/multi_inventory.py
kwoodson: If account is none, then perform an update on all accounts.

If account is set but is not found, throw an exception.

ansible/inventory/multi_inventory.py
kwoodson: Now have a return value for get_inventory.  We need to keep the self.result for the library calls from awsutil.

ansible/inventory/multi_inventory.py
kwoodson: Updated to use with.

ansible/inventory/multi_inventory.py
kwoodson: Updated the verbage.

ansible/inventory/multi_inventory.py
kwoodson: renamed to validate_cache.

ansible/inventory/multi_inventory.py
kwoodson: renamed to inventory.

ansible/inventory/multi_inventory.py
kwoodson: updated to cache_max_age.

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,294 @@
...
+                                   'Stdout: %s' % out,
+                                  ]
+                        raise RuntimeError('\n'.join(err_msg))
+
+                    updated = True
+                    # The reason this exists here is that we
+                    # are using the futures pattern.  The account object
+                    # has run a subprocess and we are collecting the results.
+                    acc.inventory = json.loads(out)
+                    acc.write_to_cache()
```
twiest: Consider using a temp var to capture json.loads and pass to write_to_cache so we get a pre-config applied cache.

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,294 @@
...
+                    continue
+
+                key = '%s_%s' % (to_name, MultiInventoryUtils.get_entry(data, from_name))
+                if not self.inventory.has_key(key):
+                    self.inventory[key] = []
+                self.inventory[key].append(name)
+
+    def apply_group_selectors(self):
+        """Apply the account config for group selectors """
+        _ = self # Here for pylint.  wanted an instance method instead of static
```
twiest: Consider removing.

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,294 @@
...
+        self.apply_clone_vars()
+
+        self.apply_extra_groups()
+
+        self.apply_clone_groups()
+
+        self.apply_group_selectors()
+
+        # store the results back into all_inventory_results
+        #return self.inventory
```
twiest: Consider removing.

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,294 @@
...
+        tmp_dir_path = tempfile.mkdtemp(prefix=prefix)
+        atexit.register(MultiInventoryUtils.cleanup, [tmp_dir_path])
+        for provider_file in self.config['provider_files']:
+            with open(os.path.join(tmp_dir_path, provider_file['name']), 'w+') as filedes:
+                content = Template(provider_file['contents']).substitute(tmpdir=tmp_dir_path)
+                filedes.write(content)
+
+        return tmp_dir_path
+
+    def write_to_cache(self):
```
twiest: Consider adding a parameter for what to write (and remove calling self.inventory).

ansible/inventory/multi_inventory.py
```diff
@@ -5,29 +5,294 @@
...
 class MultiInventoryException(Exception):
     '''Exceptions for MultiInventory class'''
     pass
 
-# pylint: disable=too-many-public-methods
-# After a refactor of too-many-branches and placing those branches into
-# their own corresponding function, we have passed the allowed amount of functions(20).
+# pylint: disable=too-many-instance-attributes
+# Need to figure out how to reduce the attributes.  Most are used for default settings
+# and determining cache, cache_max_age, results, and storing account data
```
twiest, line 292: Consider removing.

twiest (changes requested):

ansible/inventory/multi_inventory.py
kwoodson: Couldn't get rid of it.  8/7 instance attributes.

twiest (approved):

PR #1434, opened 2016-10-24
removing push from rhel7 folder

ivanhorvath:


ivanhorvath:
Approved by @kwoodson in #1433

PR #1441, opened 2016-10-24
Inventory library include fix.

kwoodson:


kwoodson:
@mwoodson :+1: in #1440

PR #1442, opened 2016-10-25
OCUtil updates

drewandersonnz:
- removing code duplication and excess variables
- forcing yaml when yaml is requested, try/except removed allowing errors to show
- not using yaml on _run_cmd by default
- replaced string concat with array join to simplify spaces
- space for logging module (if logger: logger.debug())
- baseCmd to allow different commands to be used (oc, oadm)
- pylint

Tested on ops cluster by injecting the ocutil class and testing all functions with a simple script to run all commands.


BlueShells:
:+1:

PR #1450, opened 2016-10-25
Inventory library refactor fix. Take 2

kwoodson:


kwoodson:
@mwoodson :+1: in #1449

PR #1455, opened 2016-10-25
fixed oc version issue with metrics

mwoodson:


mwoodson:
@wshearn approved #1454

PR #1457, opened 2016-10-25
fixed oc version issue for templates

mwoodson:


mwoodson:
@wshearn approved #1456

PR #1458, opened 2016-10-25
vendored the latest version of openshift-ansible for 3.4

dak1n1:


dak1n1:
pylint is complaining, but that code is part of openshift-ansible, so I don't think that's a blocker.

mwoodson (approved):

PR #1460, opened 2016-10-25
fixing the call to the tagged image:

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #1459

PR #1465, opened 2016-10-25
Bypass templates that are not idempotent.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1464

PR #1471, opened 2016-10-25
moved nickhammond logrotate to logrotate

mwoodson:


mwoodson:
@kwoodson approved #1470

PR #1473, opened 2016-10-26
use current version of docker

mwoodson:


mwoodson:
@gburges approved #1472

PR #1477, opened 2016-10-26
fixed a bug with pv vars

mwoodson:


mwoodson:
@wshearn approved #1476

PR #1480, opened 2016-10-26
Added options to allow for more controlled upgrade

dak1n1:
Exclude atomic-openshift\* from package updates, so we can control the version installed.
Added option to clean up old kernels before potentially installing another, which solves disk space issue in /boot.


ansible/roles/os_update_latest/tasks/main.yml
```diff
@@ -1,3 +1,6 @@
 ---
-- name: Update all packages
-  action: "{{ ansible_pkg_mgr }} name=* state=latest"
+- name: Make room for new kernel on /boot
+  command: package-cleanup --oldkernels --count=2 -y
```
mwoodson, line 3: Nice fix

ansible/roles/os_update_latest/tasks/main.yml
```diff
@@ -1,3 +1,6 @@
...
+
+- name: Update all packages, except for atomic-openshift
+  action: "{{ ansible_pkg_mgr }} name=* state=latest exclude='atomic-openshift* tuned-profiles-atomic-openshift*'"
```
mwoodson, line 6: Nice!

mwoodson (commented):
I'm going to look at the roles and try to give feedback on them.

dak1n1:
@twiest @mwoodson These are changes needed for the release cycle. Can someone please +1?

rharrison10:
This looks very straight forward. I'm curious why there hasn't been a +1 yet.

ansible/roles/pre_update_node/tasks/main.yml
```diff
@@ -1,16 +1,16 @@
...
+#
+## FIXME: use the oadm_manage_node module instead
+#- name: Make node unschedulable
+#  command: "oadm manage-node {{ hostvars[oo_name].ec2_private_dns_name }} --schedulable=false"
+#  register: unschedulable
+#  until: unschedulable.stdout.find("SchedulingDisabled") != -1
+#  retries: 10
+#  delegate_to: "{{ cluster_master }}"
+#  ignore_errors: True
+#  when: "{{ update_needed == True }}"
```
mwoodson, line 16: I question if this role is necessary. 

The master is usually set within our playbook.  Here is an example:

https://github.com/openshift/openshift-ansible-ops/blob/stg/playbooks/release/post_byo/registry_certs.yml#L28

That leave a one task role.  I think as a rule, a one task role isn't preferred.

When we do call this, we should use oadm_manage_node .  Here is an example:

https://github.com/openshift/openshift-ansible-ops/blob/stg/playbooks/release/upgrades/post_scale.yml#L42

We should use "node" instead of selector in this instance.

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
 ---
```
mwoodson: 1st thing, I'd like to rename this role to something like:

"openshift_master_update" or "openshift_master_package_update".  

This is a bit more descriptive.  just having "master" could mean a lot of things.

Another question.  Is this role intended to be used during blue/green or in place? or both?
  I think it was written to do hotfix update in place.

The reason I ask is because I see some restarting services at the end.  I will comment there.

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
-  with_items:
-  - atomic-openshift-master-api
-  - atomic-openshift-master-controllers
-  when: "{{ update_needed == True }}" 
-
+#
+## Default to False before attempting upgrade
+#- set_fact: 
+#    wanted_version_installed: False
+#  when: "{{ update_needed == True }}"
```
mwoodson, line 6: When creating a role, we need to do a few things.  I'm going to give an overview here, and then mark them when i see them.  A lot of these are documented here:

https://github.com/openshift/openshift-ansible/blob/master/docs/style_guide.adoc#ansible-role-variables

Let me give a high level overview:
1. all role variables need to be named with an abbreviation of the role, then the var.  For example, if this role is called "openshift_master_package_update" then the role variable prefix would be "osmpu" and the var names could be "osmpu_var_name"
2. We need to pass in variables.  We don't use global variables inside of roles.   This way all of our roles are portable and don't depend on external roles/vars.  Of course we can tie them together and use output of one as input of a new one.
   
   In this example, you have an "update_needed"  This is a global variable that got set inside of another role.  The way to handle this would be to pass in a var called "osmpu_updated_needed".  Then, use this as we go.

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
-
+#
+## Default to False before attempting upgrade
+#- set_fact: 
+#    wanted_version_installed: False
+#  when: "{{ update_needed == True }}" 
+#
+#- name: Upgrade atomic-openshift-master and dependencies
+#  command: "yum -y upgrade atomic-openshift-{{ version_wanted }} atomic-openshift-clients-{{ version_wanted }} atomic-openshift-master-{{ version_wanted }} atomic-openshift-node-{{ version_wanted }} tuned-profiles-atomic-openshift-node-{{ version_wanted }} atomic-openshift-sdn-ovs-{{ version_wanted }}"
+#  when: "{{ update_needed == True }}"
```
mwoodson, line 10: role variable needed here.

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
-  when: "{{ update_needed == True }}" 
-
+#
+## Default to False before attempting upgrade
+#- set_fact: 
+#    wanted_version_installed: False
+#  when: "{{ update_needed == True }}" 
+#
+#- name: Upgrade atomic-openshift-master and dependencies
+#  command: "yum -y upgrade atomic-openshift-{{ version_wanted }} atomic-openshift-clients-{{ version_wanted }} atomic-openshift-master-{{ version_wanted }} atomic-openshift-node-{{ version_wanted }} tuned-profiles-atomic-openshift-node-{{ version_wanted }} atomic-openshift-sdn-ovs-{{ version_wanted }}"
```
mwoodson, line 9: is it possible to use the yum module here?  I honestlty don't know.  The less command version, the better

```
yum:
  name: "{{ item }}"
  state: installed
with_items:
- "atomic-openshift-{{ version_wanted }}"
- "atomic-openshift-clients-{{ version_wanted }}"
- ...
```

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
+#  when: "{{ update_needed == True }}" 
+#
+#- name: Check if update is successful
+#  command: "rpm -qa atomic-openshift"
```
mwoodson, line 13: repoquery would be preferred:

```
- name: Check if package is installed
  repoquery:
    name: "{{ pun_package_name }}"
    show_duplicates: True
    query_type: installed
  register: package_check_out
```

then, in the meta/main.yml, add "tools_roles/lib_repoquery" as a dependency.

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
+#  register: post_upgrade
+#  changed_when: False
+#
+## ansible will return a '17' if a match is found. '-1' means 'not found'
+#- name: Set wanted_version_installed=True if upgrade succeeded
+#  set_fact: 
+#    wanted_version_installed: True
+#  when: "{{ post_upgrade.stdout.find(version_wanted) != -1 }}"
```
mwoodson, line 21: Here is a repoquery version of this check.

```
- name: The installed version of the rpm didn't match what we wanted. Set update_needed=True
  set_fact:
    pun_retval_update_needed: True
  when: pun_wanted_version not in package_check_out.results.versions.available_versions_full

```

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
+#
+#- name: Upgrade failed, so try a downgrade
+#  command: "yum -y downgrade atomic-openshift-{{ version_wanted }} atomic-openshift-clients-{{ version_wanted }} atomic-openshift-master-{{ version_wanted }} atomic-openshift-node-{{ version_wanted }} tuned-profiles-atomic-openshift-node-{{ version_wanted }} atomic-openshift-sdn-ovs-{{ version_wanted }}"
+#  when: "{{ wanted_version_installed == False }}"
+#
+#- name: Check if downgrade is successful
```
mwoodson, line 27: again, repoquery.

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
+#- name: Set wanted_version_installed=True if downgrade succeeded
+#  set_fact: 
+#    wanted_version_installed: True
+#  when: "{{ post_downgrade.stdout.find(version_wanted) != -1 }}"
+#
+#- name: Alert if update is not successful, but continue
+#  debug:
+#    msg: "[WARNING] The desired rpm version does not match installed version! Upgrade and downgrade failed!"
+#  when: "{{ wanted_version_installed == False }}"
+#  failed_when: True
```
mwoodson, line 42: curious why you chose debug + failed_when.  Why not use "fail module + when"

ansible/roles/update_master/tasks/main.yml
```diff
@@ -1,70 +1,70 @@
...
+#  with_items:
+#  - atomic-openshift-master-api
+#  - atomic-openshift-master-controllers
+#  register: restart_master_services
+#  until: restart_master_services|succeeded
+#  retries: 3
+#  delay: 10
+#  when: "{{ update_needed == True and wanted_version_installed == True }}"
+#
+#- name: Ensure master services are running
```
mwoodson, line 62: Is this task needed?  It seems like the one above, the service restarted, would take care of this task. 

If the restart fails, ansible module will fail (above).

mwoodson:
I noticed that the code in update_master and update_node is about 90% the same.  I think we could combine them.

I think we should name it something like "openshift_update_packages"

I would remove the oadm manage schedule false from this role, make it specific to just updating packages.  We can then add the scheduling to another part of the playbook.

ansible/roles/update_node/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
+---
```
mwoodson, line 1: I have discussed with @dak1n1 about combining this role with the update_master.  Please see the comment page for some thoughts.

mwoodson (commented):
I have left some comments about this pr.

ansible/roles/pre_update_node/tasks/main.yml
dak1n1: I like the example of setting the master that you showed me. I'll try to implement that.

You could be right that the role is not necessary. But it's possible that we'll be left with a one-task role until `when` statements work with `include_role`. I'll see if I can address both these issues.

There's one part of your comment that I didn't understand: 'We should use "node" instead of selector in this instance'. What do you mean by that? 'selector' doesn't appear in the above code that I can see.

ansible/roles/update_master/tasks/main.yml
dak1n1: I'm using the update_master role to update the masters for blue/green upgrades in INT. So we should be able to use it for both in-place and blue/green upgrades, since the master portion of the process is the same. I can go ahead and change the name.

ansible/roles/update_master/tasks/main.yml
dak1n1: I don't think so. I tried that originally and we ended up with yum installing version 3.3 and then downgrading to version 3.2 because one of the dependencies it found was the wrong version. So I had to be really explicit about updating _all_ the dependencies in one transaction.

ansible/roles/update_master/tasks/main.yml
dak1n1: Because `failed_when` colors the debug statement red and makes it stand out. I want it to say "hey, something went wrong" in a way that someone will notice. But I don't want it to stop the update. (I don't want it to fail, which would cause the update to exit).

ansible/roles/update_master/tasks/main.yml
dak1n1: I wanted to handle the case of the restart being successful, but the service failing afterwards. Sometimes the service can restart successfully and still leave us in a bad state. Though I'm not certain that checking the service state will alert us to that case. I thought it would be more thorough than assuming the service has kept running after the restart.

ansible/roles/update_master/tasks/main.yml
rharrison10: :+1:

ansible/roles/update_master/tasks/main.yml
rharrison10: @mwoodson I tried using the yum module in exactly the manor you've described in the past without good results. I have not tried using specific version with the yum module since Ansible 2.x came out though.

@dak1n1 it turns out the yum module when using `with_items` does in fact bundle the items into a single transaction. It didn't always behave that way, but it is now one place where Ansible has a huge advantage over Puppet.

We'll need to test the full implementation with the yum module to see if its possible.

ansible/roles/update_master/tasks/main.yml
rharrison10: Is the `repoquery` module one of ours? Isn't it by nature (being yum repo queries in python) going to be exponentially slower than querying using rpm directly which is C code querying the much smaller `rpmdb`?

I would think the speed advantages of calling `rpm` instead of `repoquery` would far exceed the advantages of using a true module instead of creating behavior in the `command` module.

ansible/roles/update_master/tasks/main.yml
rharrison10: I'd like to know more about the `repoquery` module. I know from experience that the arguments passed into repoquery have changed dramatically through the life of the command and I'm fairly sure the functionality is being rolled into `dnf` with a future goal of removing the command entirely. Does the module work on both RHEL 6 and RHEL 7?

ansible/roles/update_master/tasks/main.yml
rharrison10: Interesting I would have thought that [`failed_when: True`](http://docs.ansible.com/ansible/playbooks_error_handling.html#controlling-what-defines-failure) would have caused the playbook to exit at this point. I might have tried done the same thing by using [`ignore_errors: yes`](http://docs.ansible.com/ansible/playbooks_error_handling.html#ignoring-failed-commands) on the `fail` module.

Perhaps its worth discussing a standard way of creating the high level of alert without aborting a playbook should be handled with the rest of the team.

ansible/roles/pre_update_node/tasks/main.yml
mwoodson: Let me be a bit more clear.

In the above post_scale.yml playbook I was shoing an examle of how to use oadm_manage_node

```
 - name: oadm manage-node --schedulable=true --selector=ops_node=new
    oadm_manage_node:
      selector: ops_node=new
      schedulable: True
```

In this case, we used a "selector" option.  I am suggesting we don't use a selector, but a "node".  This will duplicate the functionality you have above.  This module is also idempotent

ansible/roles/update_master/tasks/main.yml
mwoodson: This may need testing, thanks for the discussion.

We confirmed with Ansible that certain module, yum being one of them, will actually take the "with_items" and combine them.  We asked for use of this feature, because we wanted to implement it with yedit (do all the changes at once, not each change separately".  We were told we couldn't. 

But, for yum, it does take all the with_items and combine them.

I have had luck using yum and versioning.  I actually used it for docker for quite some time.

If time allows, this would be great to test.

ansible/roles/update_master/tasks/main.yml
mwoodson: repoquery is a module we wrote.  It can query local and remote repos.

If there was an "rpm" module, i'd look at using that, but for now, I'd like to use non-command/shell modules.

ansible/roles/update_master/tasks/main.yml
mwoodson: https://github.com/openshift/openshift-tools/tree/stg/ansible/roles/lib_repoquery

It uses the command line and then presents the data in an ansible friendly way.

ansible/roles/update_master/tasks/main.yml
mwoodson: I thought like @rharrison10 did, that the failed_when: True happened, the playbook would exit.

If that isn't the behavior you want, I understand then.

PR #1487, opened 2016-10-27
Make sure IAM scripts are execuatable

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1486

PR #1489, opened 2016-10-27
changing script to find the image instead of hardcoding it

ivanhorvath:


scripts/monitoring/cron-send-docker-dns-resolution.py
```diff
@@ -21,8 +21,14 @@
 
 if __name__ == "__main__":
     cli = AutoVersionClient(base_url='unix://var/run/docker.sock')
-    container = cli.create_container(image='docker-registry.ops.rhcloud.com/ops/oso-rhel7-host-monitoring',
+
+    container_id = ''
+    for con in cli.containers(filters={'name': 'oso-rhel7-host-monitoring'}):
```
twiest: Can we use a pattern here?

Don't forget that we have centos containers that can run this code too.

Also, what happened to getting the container_id from the environment since we're running inside of the container?

twiest (changes requested):
I like where this is going, MUCH more dynamic. Just a couple of questions to sort out.

twiest (approved):

PR #1491, opened 2016-10-27
changing script to find the image instead of hardcoding it

ivanhorvath:


ivanhorvath:
Approved by @twiest in #1489

PR #1493, opened 2016-10-27
hide synthetic hosts from ossh

blentz:


blentz:
@kwoodson gave :+1: in #1492

PR #1495, opened 2016-10-27
aws_account_list: Pass in path to aws_accounts.txt

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1494

PR #1501, opened 2016-10-27
sadly cron is running without the right environmental vars

ivanhorvath:


ivanhorvath:
Approved by @robotmaxtron in #1500

PR #1502, opened 2016-10-28
Fail early if pod not found early in process

drewandersonnz:
This happens due to a race condition. Further development is needed to avoid this completely.


BlueShells:
:+1: 
this will fix the fake alert of app-create , We don't want this page anyone for the fake alert ,so we decide push the fix today

PR #1506, opened 2016-10-28
ignore_errors to solve undefined var issue

drewandersonnz:


BlueShells:
:+1: 

this is short-term fix for the error

PR #1509, opened 2016-10-28
updated the verfiy aws role

mwoodson:


mwoodson:
@kwoodson approved #1508

PR #1510, opened 2016-10-28
Idempotently add or remove repo ids.

kwoodson:


ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,202 @@
...
+  - whether to list or set the repositories.
+  required: false
+  default: present
+  aliases: []
+query:
+  choices: ['all', 'enabled', 'disabled']
+  description:
+  - The query type when looking at repositories.
+  required: false
+  default: present
```
twiest: `present` isn't in the list of `choices`.

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,202 @@
...
+        if qtype == 'enabled':
+            cmd.append('--list-enabled')
+        elif qtype == 'disabled':
+            cmd.append('--list-disabled')
+        else:
+            cmd.append('--list')
+
+        rcode, out, err = self.module.run_command(cmd)
+        if rcode != 0:
+            raise RHSMError('error querying repositories. rc=[%s] cmd=[%s] err=[%s]' % (rcode, cmd, err))
```
twiest: Should this be a `self.module.fail_json` instead?

-or- should we wrap the caller in a try/except and do a `self.module.fail_json`  ?

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,202 @@
...
+        """verify that the enabled repos are enabled"""
+        repos = self.query_repos(qtype='enabled')
+        if len(repos) != len(self.module.params['enabled']):
+            return True, repos
+
+        # build sets of the repo ids and compare them
+        rids = set([rep['id'] for rep in repos])
+        ren = set(self.enabled)
+        if rids - ren or ren - rids:
+            return True, repos
```
twiest: Love this.

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,202 @@
...
+def main():
+    """Create the ansible module and run the ansible code"""
+    module = AnsibleModule(
+        argument_spec=dict(
+            enabled=dict(default=None, type='list'),
+            disabled=dict(default=None, type='list'),
+            state=dict(default='present', choices=['list', 'present']),
+            query=dict(default='all', choices=['all', 'enabled', 'disabled']),
+        ),
+        supports_check_mode=True
```
twiest, line 212: How is `check_mode` being supported in this module?

I probably missed it, but I looked through the code twice for it and don't see anything that's checking for check_mode.

Is this maybe an automatic feature of `module.run_command`?

The [docs](http://docs.ansible.com/ansible/dev_guide/developing_modules.html#check-mode) say you need to do an if statement using `module.check_mode`.

twiest (changes requested):
@kwoodson This is looking VERY good. Just a few comments to address.

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
kwoodson: You are right.  This is an artifact.  Let me fix this.

kwoodson: @twiest, added the check.

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # step 2: if we are state=present, check if we need to update
+        elif module.params['state'] == 'present':
+            update, repos = rhsm.needs_update()
+            if not update:
+                module.exit_json(changed=update, repos=repos)
+
+            # exit if check mode before we make changes
+            if module.check_mode:
+                module.exit_json(changed=False, repos=repos)
```
twiest: I think the purpose of check_mode is to say if changes would have been made.

So I think we want `changed=True` here, even though changes weren't made.

From the check_mode [doc:](http://docs.ansible.com/ansible/playbooks_checkmode.html)
"Instead, any module instrumented to support ‘check mode’ (which contains most of the primary core modules, but it is not required that all modules do this) will report what changes they would have made rather than making them."

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,211 @@
...
+        if rhsm.module.params['state'] == 'list':
+            repos = rhsm.query_repos(qtype=module.params['query'])
+
+            rhsm.module.exit_json(changed=False, repos=repos)
+
+        # step 2: if we are state=present, check if we need to update
+        elif module.params['state'] == 'present':
+            update, repos = rhsm.needs_update()
+            if not update:
+                module.exit_json(changed=update, repos=repos)
```
twiest: I prefer to be more explicit here and just have `changed=False`. It's MUCH clearer what's being returned and there isn't a case where this will run with `update == True`.

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,211 @@
...
+
+            # exit if check mode before we make changes
+            if module.check_mode:
+                module.exit_json(changed=False, repos=repos)
+
+            # step 3: we need to set our repositories
+            rhsm.set_repositories()
+            repos = rhsm.query_repos(qtype='enabled')
```
twiest: Should we check needs_update again and fail the module if it comes back `True`?

ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -0,0 +1,211 @@
...
+            module.exit_json(changed=update, repos=repos)
```
twiest: Same here as above, we know that it changed because we wouldn't be running this code if it didn't, so I prefer `changed=True` just to be explicit for the next person reading the code.

twiest (changes requested):
A few more cleanups.

twiest (approved):

PR #1511, opened 2016-10-31
Add MetricSender wrapper class to monitoring lib

zgalor:
Continuing work done in https://github.com/openshift/openshift-tools/pull/1178.

- **MetricSender** class that wraps **HawkSender** and **ZaggSender**. Metrics are sent to each sender that's configured "**active**" ( by default Zagg is active. Hawk is inactive)

- config file: **metric_sender.yaml** (replacing zagg_client.yaml and adding hawk connectivity options).

- cli's for shell scripts:

1. **ops-metric-client** (replacing ops-zagg-client).
2. **ops-metric-pcp-client** (replacing ops-zagg-pcp-client)

- Following cli's were not replaced (used in ansible modules, and docker scripts):

1. ops-zagg-hearbet-processor
2. ops-zagg-hearbeter
3. ops-zagg-metric-processor

- Scripts in monitoring and cloud libraries were updated to use new MetricSender, and tested.

- ansible playbooks and docker shell scripts not updated and still use ops-zagg-client

- tags are not yet being sent to hawkular. We need to define the way we want to send them and add them as another commit or even a different PR

### DIFF
--- a/openshift_tools/monitoring/generic_metric_sender.py
+++ b/openshift_tools/monitoring/generic_metric_sender.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+Base class for sending metrics
+"""
+
+import os
+import yaml
+
+
+class GenericMetricSenderException(Exception):
+    '''
+        GenericMetricSenderException
+        Exists to propagate errors up from the api
+    '''
+    pass
+
+class GenericMetricSender(object):
+    """
+    Abstract class that encapsulates all metric sending operations.
+    """
+
+    # pylint: disable=unidiomatic-typecheck
+    def __init__(self):
+        """ raise exception - Abstract class shouldn't instantiate"""
+        if type(self) is GenericMetricSender:
+            raise TypeError("This is an abstract class and shouldn't be directly instantiated")
+
+        self.unique_metrics = []
+        self.config = None
+        self.config_file = None
+
+
+    def parse_config(self):
+        """ parse default config file """
+        if not self.config:
+            if not os.path.exists(self.config_file):
+                raise GenericMetricSenderException(self.config_file + " does not exist.")
+            self.config = yaml.load(file(self.config_file))
+
+    # Allow for 6 arguments (including 'self')
+    # pylint: disable=too-many-arguments
+    def add_dynamic_metric(self, discovery_key, macro_string, macro_array, host=None, synthetic=False):
+        """ empty implementation overridden by derived classes """
+        pass
+
+    def add_metric(self, metric, host=None, synthetic=False):
+        """ empty implementation overridden by derived classes """
+        pass
+
+    def add_heartbeat(self, add_heartbeat, host=None):
+        """ empty implementation overridden by derived classes """
+        pass
+
+    def send_metrics(self):
+        """ empty implementation overridden by derived classes """
+        pass
+
+    def print_unique_metrics_key_value(self):
+        """
+        This function prints the key/value pairs the UniqueMetrics that sender
+        currently has stored
+        """
+        print "\n" + self.__class__.__name__ + " Key/Value pairs:"
+        print "=============================="
+        for unique_metric in self.unique_metrics:
+            print("%s:  %s") % (unique_metric.key, unique_metric.value)
+        print "==============================\n"
+
+    def print_unique_metrics(self):
+        """
+        This function prints all of the information of the UniqueMetrics that sender
+        currently has stored
+        """
+        print "\n" + self.__class__.__name__ + " UniqueMetrics:"
+        print "=============================="
+        for unique_metric in self.unique_metrics:
+            print unique_metric
+        print "==============================\n"
+
--- a/openshift_tools/monitoring/hawk_client.py
+++ b/openshift_tools/monitoring/hawk_client.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python2
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+HawkClient uses the openshift_tools.web/rest.py
+
+Example usage:
+
+    from openshift_tools.monitoring.metricmanager import UniqueMetric
+    from openshift_tools.monitoring.hawk_common import HawkConnection
+    from openshift_tools.monitoring.hawk_client import HawkClient
+
+    ml = []
+    # UniqueMetric from openshift_tools.monitoring.metricmanager
+    new_metric = UniqueMetric('example.com','cpu/usage',12.34)
+    ml.append(new_metric)
+    new_metric = UniqueMetric('example.org','cpu/usage',4.2)
+    ml.append(new_metric)
+
+    # No Auth
+    connection = HawkConnection(url='172.17.0.27:8000')
+
+    # Basic Auth
+    connection = HawkConnection(url='172.17.0.27:8000', user='user', passwd='password')
+
+    client = HawkClient(connection)
+    print client.add_metric(ml)
+"""
+#These are not installed on the buildbot, disabling this
+#pylint: disable=no-name-in-module,unused-import,import-error
+from openshift_tools.monitoring.metricmanager import UniqueMetric, MetricManager
+from openshift_tools.monitoring.hawk_common import HawkConnection
+
+# The hawkular client
+from hawkular.metrics import HawkularMetricsClient, MetricType, Availability
+
+#This class implements rest calls. We only have one rest call implemented
+# add-metric.  More could be added here
+#pylint: disable=too-few-public-methods
+class HawkClient(object):
+    """
+    wrappers class around hawkular client python so use can use it with UniqueMetric
+    """
+
+    def __init__(self, hawk_connection):
+        # pylint doesn't know where RestAPI is
+        #pylint: disable=undefined-variable
+        self.hawk_conn = hawk_connection
+        self.client = None
+
+        # Do not create a client if inactive
+        if not self.hawk_conn.active:
+            return
+
+        self.client = HawkularMetricsClient(host=self.hawk_conn.host,
+                                            port=self.hawk_conn.port,
+                                            scheme=self.hawk_conn.scheme,
+                                            username=self.hawk_conn.username,
+                                            password=self.hawk_conn.password,
+                                            context=self.hawk_conn.context,
+                                            tenant_id=self.hawk_conn.tenant_id,
+                                           )
+
+    def add_metric(self, unique_metric_list):
+        """
+        Add a list of UniqueMetrics (unique_metric_list) via hawkular client
+        """
+        # Do not run if inactive
+        if not self.hawk_conn.active:
+            return
+
+        # We do not support zabbix heartbeat in hawkular
+        metric_list = [m for m in unique_metric_list if m.key != 'heartbeat']
+
+        for metric in metric_list:
+            # Hawkular metrics support any numeric values
+            value = metric.value
+            # Hawkular metrics use milliseconds
+            clock = metric.clock * 1000
+            # Add the type and host to the key, this is needed because in hawkular,
+            # these need to be namespaced otherwise multiple hosts will be adding
+            # values to the same key. See common metrics keys proposal.
+            _type = 'node'
+            _id = metric.host
+            key = '{0}/{1}/{2}'.format(_type, _id, metric.key)
+            # Use MetricType.Gauge for numeric metrics data
+            metric_type = MetricType.Gauge
+            self.client.push(metric_type, key, value, clock)
+
+        status, raw_respons = None, None
+        return (status, raw_respons)
--- a/openshift_tools/monitoring/hawk_client.py
+++ b/openshift_tools/monitoring/hawk_client.py
@@ -25,16 +25,16 @@ Example usage:
     client = HawkClient(connection)
     print client.add_metric(ml)
 """
+# The hawkular client
+from hawkular.metrics import HawkularMetricsClient, MetricType
+
 #These are not installed on the buildbot, disabling this
 #pylint: disable=no-name-in-module,unused-import,import-error
-from openshift_tools.monitoring.metricmanager import UniqueMetric, MetricManager
+from openshift_tools.monitoring.metricmanager import UniqueMetric
 from openshift_tools.monitoring.hawk_common import HawkConnection
 
-# The hawkular client
-from hawkular.metrics import HawkularMetricsClient, MetricType, Availability
-
 #This class implements rest calls. We only have one rest call implemented
-# add-metric.  More could be added here
+# add_metric.  More could be added here
 #pylint: disable=too-few-public-methods
 class HawkClient(object):
     """
@@ -42,8 +42,6 @@ class HawkClient(object):
     """
 
     def __init__(self, hawk_connection):
-        # pylint doesn't know where RestAPI is
-        #pylint: disable=undefined-variable
         self.hawk_conn = hawk_connection
         self.client = None
 
@@ -60,7 +58,8 @@ class HawkClient(object):
                                             tenant_id=self.hawk_conn.tenant_id,
                                            )
 
-    def add_metric(self, unique_metric_list):
+    # pylint: disable=unidiomatic-typecheck
+    def push_metrics(self, unique_metric_list):
         """
         Add a list of UniqueMetrics (unique_metric_list) via hawkular client
         """
@@ -68,11 +67,8 @@ class HawkClient(object):
         if not self.hawk_conn.active:
             return
 
-        # We do not support zabbix heartbeat in hawkular
-        metric_list = [m for m in unique_metric_list if m.key != 'heartbeat']
-
-        for metric in metric_list:
-            # Hawkular metrics support any numeric values
+        for metric in unique_metric_list:
+            # Hawkular metrics support any numeric/string values
             value = metric.value
             # Hawkular metrics use milliseconds
             clock = metric.clock * 1000
@@ -82,9 +78,15 @@ class HawkClient(object):
             _type = 'node'
             _id = metric.host
             key = '{0}/{1}/{2}'.format(_type, _id, metric.key)
-            # Use MetricType.Gauge for numeric metrics data
-            metric_type = MetricType.Gauge
+
+            if type(value) == str:
+                # Use MetricType.String for string metrics data
+                metric_type = MetricType.String
+            else:
+                # Use MetricType.Gauge for numeric metrics data
+                metric_type = MetricType.Gauge
+
             self.client.push(metric_type, key, value, clock)
 
-        status, raw_respons = None, None
-        return (status, raw_respons)
+        status, raw_response = None, None
+        return (status, raw_response)
--- a/openshift_tools/monitoring/hawk_client.py
+++ b/openshift_tools/monitoring/hawk_client.py
@@ -26,12 +26,9 @@ Example usage:
     print client.add_metric(ml)
 """
 # The hawkular client
+# hawkular client is not installed on the buildbot, disabling this
+#pylint: disable=import-error
 from hawkular.metrics import HawkularMetricsClient, MetricType
-
-#These are not installed on the buildbot, disabling this
-#pylint: disable=no-name-in-module,unused-import,import-error
-from openshift_tools.monitoring.metricmanager import UniqueMetric
-from openshift_tools.monitoring.hawk_common import HawkConnection
 
 #This class implements rest calls. We only have one rest call implemented
 # add_metric.  More could be added here
--- a/openshift_tools/monitoring/hawk_common.py
+++ b/openshift_tools/monitoring/hawk_common.py
@@ -1,0 +1,38 @@
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+hawk_common contains common data structures and utils
+
+Example usage:
+    from openshift_tools.monitoring.hawk_common import HawkConnection, HawkHeartbeat
+
+    ZAGGCONN = HawkConnection(url='172.17.0.151', user='admin', password='pass')
+    ZAGGHEARTBEAT = HawkHeartbeat(templates=['template1', 'template2'], hostgroups=['hostgroup1', 'hostgroup2'])
+
+"""
+from collections import namedtuple
+from urlparse import urlparse
+import ssl
+
+# pylint: disable=too-few-public-methods,too-many-arguments,protected-access,too-many-instance-attributes
+# This is a DTO.  It needs to be a class because we need default values
+class HawkConnection(object):
+    ''' DTO for hawkular connection
+    '''
+    # pylint: disable=too-many-arguments
+    # This now supports ssl and need a couple of extra params
+    def __init__(self, url, user, password, ssl_verify=False, debug=False, active=True):
+        self.url = url if url.startswith('http') else 'http://' + url
+        self.username = user
+        self.password = password
+        self.ssl_verify = ssl_verify
+        self.debug = debug
+        self.active = active # We do not want this lib to work ansless explisitly activated
+
+        url_params = urlparse(self.url)
+        self.host = url_params.hostname
+        self.scheme = url_params.scheme
+        self.port = url_params.port or {'http': 80, 'https': 443}[self.scheme]
+        self.context = None if self.ssl_verify else ssl._create_unverified_context()
+        self.tenant_id = '_ops'
+
+HawkHeartbeat = namedtuple("HawkHeartbeat", ["templates", "hostgroups"])
--- a/openshift_tools/monitoring/hawk_common.py
+++ b/openshift_tools/monitoring/hawk_common.py
@@ -9,7 +9,6 @@ Example usage:
     ZAGGHEARTBEAT = HawkHeartbeat(templates=['template1', 'template2'], hostgroups=['hostgroup1', 'hostgroup2'])
 
 """
-from collections import namedtuple
 from urlparse import urlparse
 import ssl
 
@@ -26,7 +25,7 @@ class HawkConnection(object):
         self.password = password
         self.ssl_verify = ssl_verify
         self.debug = debug
-        self.active = active # We do not want this lib to work ansless explisitly activated
+        self.active = active # We do not want this lib to work unless explicitly activated
 
         url_params = urlparse(self.url)
         self.host = url_params.hostname
@@ -34,5 +33,3 @@ class HawkConnection(object):
         self.port = url_params.port or {'http': 80, 'https': 443}[self.scheme]
         self.context = None if self.ssl_verify else ssl._create_unverified_context()
         self.tenant_id = '_ops'
-
-HawkHeartbeat = namedtuple("HawkHeartbeat", ["templates", "hostgroups"])
--- a/openshift_tools/monitoring/hawk_sender.py
+++ b/openshift_tools/monitoring/hawk_sender.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+Collect metrics and send metrics to Hawk.  The data
+being send to Hawk is done using REST API using the HawkClient
+module
+
+Examples:
+
+    from openshift_tools.monitoring.hawk_common import HawkConnection, HawkHeartbeat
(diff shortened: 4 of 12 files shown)


openshift_tools/monitoring/generic_metric_sender.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def parse_config(self):
+        """ parse default config file """
+        if not self.config:
+            if not os.path.exists(self.config_file):
+                raise GenericMetricSenderException(self.config_file + " does not exist.")
+            self.config = yaml.load(file(self.config_file))
+
+    def add_dynamic_metric(self, discovery_key, macro_string, macro_array, host=None, synthetic=False):
+        pass
```
twiest, line 45: As an Abstract class, should this raise if called?

If so, the other functions should be changes as well (I'm not going to add a comment for each).

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Basic Auth
+    connection = HawkConnection(url='172.17.0.27:8000', user='user', passwd='password')
+
+    client = HawkClient(connection)
+    print client.add_metric(ml)
+"""
+#These are not installed on the buildbot, disabling this
+#pylint: disable=no-name-in-module,unused-import,import-error
+from openshift_tools.monitoring.metricmanager import UniqueMetric, MetricManager
```
twiest: `MetricManager` isn't used in this file.

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,87 @@
...
+from openshift_tools.monitoring.hawk_common import HawkConnection
+
+# The hawkular client
+from hawkular.metrics import HawkularMetricsClient, MetricType, Availability
+
+#This class implements rest calls. We only have one rest call implemented
+# add-metric.  More could be added here
```
twiest: `add_metric`

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,87 @@
...
+#pylint: disable=too-few-public-methods
+class HawkClient(object):
+    """
+    wrappers class around hawkular client python so use can use it with UniqueMetric
+    """
+
+    def __init__(self, hawk_connection):
+        # pylint doesn't know where RestAPI is
+        #pylint: disable=undefined-variable
```
twiest: Do we still need this pylint disable?

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,87 @@
...
+        self.client = HawkularMetricsClient(host=self.hawk_conn.host,
+                                            port=self.hawk_conn.port,
+                                            scheme=self.hawk_conn.scheme,
+                                            username=self.hawk_conn.username,
+                                            password=self.hawk_conn.password,
+                                            context=self.hawk_conn.context,
+                                            tenant_id=self.hawk_conn.tenant_id,
+                                           )
+
+    def add_metric(self, unique_metric_list):
```
twiest: Since the abstract class has `synthetic` as a param, I think this needs it as well, right?

Even if it's not used for hawkular (as all are semi-synthetic already).

openshift_tools/monitoring/hawk_common.py
```diff
@@ -0,0 +1,38 @@
...
+
+# pylint: disable=too-few-public-methods,too-many-arguments,protected-access,too-many-instance-attributes
+# This is a DTO.  It needs to be a class because we need default values
+class HawkConnection(object):
+    ''' DTO for hawkular connection
+    '''
+    # pylint: disable=too-many-arguments
+    # This now supports ssl and need a couple of extra params
+    def __init__(self, url, user, password, ssl_verify=False, debug=False, active=True):
+        self.url = url if url.startswith('http') else 'http://' + url
```
twiest, line 23: Does this work with https?

Seems like it's being hard coded to http.

openshift_tools/monitoring/hawk_common.py
```diff
@@ -0,0 +1,38 @@
...
+        self.username = user
+        self.password = password
+        self.ssl_verify = ssl_verify
+        self.debug = debug
+        self.active = active # We do not want this lib to work ansless explisitly activated
```
twiest: `ansless explisitly`  should be `unless explicitly`

openshift_tools/monitoring/hawk_common.py
```diff
@@ -0,0 +1,38 @@
...
+
+        url_params = urlparse(self.url)
+        self.host = url_params.hostname
+        self.scheme = url_params.scheme
+        self.port = url_params.port or {'http': 80, 'https': 443}[self.scheme]
+        self.context = None if self.ssl_verify else ssl._create_unverified_context()
+        self.tenant_id = '_ops'
+
+HawkHeartbeat = namedtuple("HawkHeartbeat", ["templates", "hostgroups"])
```
twiest: Is this needed?

`templates` and `hostgroups` are zabbix concepts that we're hoping to not have to bring to hawkular.

openshift_tools/monitoring/hawk_sender.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env python
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+Collect metrics and send metrics to Hawk.  The data
+being send to Hawk is done using REST API using the HawkClient
```
twiest: `send` should be `sent`

openshift_tools/monitoring/hawk_sender.py
```diff
@@ -0,0 +1,150 @@
...
+            print unique_metric
+        print "==============================\n"
+
+    def get_default_host(self):
+        """ get the 'host' value from the config file """
+        self.parse_config()
+
+        return self.config['host']['name']
+
+    def _get_default_hawk_connecton(self):
```
twiest: `_get_default_hawk_connecton` should be `_get_default_hawk_connection`

openshift_tools/monitoring/metric_sender.py
```diff
@@ -0,0 +1,70 @@
...
+    from openshift_tools.monitoring.hawk_sender import HawkSender
+    HOSTNAME = 'host.example.com'
+
+    ZAGGCONN = HawkConnection(url='https://172.17.0.151', user='admin', password='pass')
+    ZAGGHEARTBEAT = HawkHeartbeat(templates=['template1', 'template2'], hostgroups=['hostgroup1', 'hostgroup2'])
+
+    zs = HawkSender(host=HOSTNAME, hawk_connection=ZAGGCONN)
+    zs.add_heartbeat(ZAGGHEARTBEAT)
+    zs.add_zabbix_keys({ 'test.key' : '1' })
+    zs.send_metrics()
```
twiest: This example is specific to hawk sender. It should probably be changed to show how to use MetricSender.

openshift_tools/monitoring/metric_sender.py
```diff
@@ -0,0 +1,70 @@
...
+        self.config_file = '/etc/openshift_tools/metric_sender.yaml'
+        self.verbose = verbose
+        self.debug = debug
+        self.active_senders = None
+
+        super.parse_config()
+        if self.is_zagg_active:
+            self.active_senders += ZaggSender(verbose, debug)
+
+        if self.is_hawk_active:
```
twiest: same

openshift_tools/monitoring/metric_sender.py
```diff
@@ -0,0 +1,70 @@
...
+        '''
+        super.__init__
+
+        self.config_file = '/etc/openshift_tools/metric_sender.yaml'
+        self.verbose = verbose
+        self.debug = debug
+        self.active_senders = None
+
+        super.parse_config()
+        if self.is_zagg_active:
```
twiest: Is this a function call? If so, why no `()` on the end?

openshift_tools/monitoring/zagg_sender.py
```diff
@@ -151,9 +147,13 @@ def add_zabbix_keys(self, zabbix_keys, host=None, synthetic=False):
...
         self.unique_metrics += zabbix_metrics
 
+    def add_metric(self, metric, host=None, synthetic=False):
+        """ Wrapper for _add_zubbix_keys """
+        self._add_zabbix_keys(metric, host, synthetic)
+
     # Allow for 6 arguments (including 'self')
     # pylint: disable=too-many-arguments
-    def add_zabbix_dynamic_item(self, discovery_key, macro_string, macro_array, host=None, synthetic=False):
+    def _add_zabbix_dynamic_item(self, discovery_key, macro_string, macro_array, host=None, synthetic=False):
```
twiest: Is there a reason not to just rename this to `add_dynamic_metric` directly instead of doing the forward method?

openshift_tools/monitoring/zagg_sender.py
```diff
@@ -135,7 +131,7 @@ def add_heartbeat(self, heartbeat, host=None):
                                                  )
         self.unique_metrics.append(hb_metric)
 
-    def add_zabbix_keys(self, zabbix_keys, host=None, synthetic=False):
+    def _add_zabbix_keys(self, zabbix_keys, host=None, synthetic=False):
```
twiest: Is there a reason not to just rename this to `add_metric` directly instead of doing the forward method?

twiest (changes requested):
This is really looking good, I really like how this is shaping up.

This is exactly what we're looking for. I added a few comments.

openshift_tools/monitoring/generic_metric_sender.py
zgalor: Actually I want this implemented as "Do-Nothing" in the parent class so in the case of add_dynamic_metric for example which is not implemented in HawkSender, function will default to parent. This will allow generic calls to MetricSender no matter what concrete sender supports.
I will however, raise an exception in the constructor, to prevent instantiation of this abstract class. Thanks.

openshift_tools/monitoring/hawk_client.py
zgalor: Will fix

openshift_tools/monitoring/hawk_client.py
zgalor: :+1:

openshift_tools/monitoring/hawk_common.py
zgalor: I will check

openshift_tools/monitoring/hawk_common.py
zgalor: I will do more cleanup once I see things working

openshift_tools/monitoring/metric_sender.py
zgalor: off course

openshift_tools/monitoring/metric_sender.py
zgalor: No need in python, if function doesn't take parameters

openshift_tools/monitoring/zagg_sender.py
zgalor: I will. This is temporary for backwards compatibility and for change tracking

openshift_tools/monitoring/hawk_common.py
yaacov: checking for `url.startswith('http')` is ok is ok for http and https ...

```
url                   | self.url
----------------------|------------------------
example.com           | http://example.com
http://example.com    | http://example.com
https://example.com   | https://example.com
```

yaacov:
:+1:

openshift_tools/monitoring/generic_metric_sender.py
twiest: I see what you're going for, and that's fine.

However, we DO need dynamic metric supported by the hawkular calls as well. Dynamic metrics are anything where the script needs to append a run time specific value.

So, for instance, our disk space check. When we write it, we have no idea what devices the host will have. So we write it as a dynamic metric and then pass into the dynamic metric /dev/xvda or /dev/xvdb, etc to it.

In hawkular, what I envision is that we'll simply add more tags for the dynamic items.

I think this will make more sense once you try to convert a collector script that uses dynamic items over to work with hawkular as well.

openshift_tools/monitoring/hawk_common.py
twiest: Oh, I see. Of course, it's only adding http:// if the url doesn't start with http. I read the `else` as a `then`.

openshift_tools/monitoring/metric_sender.py
twiest: I could be wrong about this, but what's happening here is that since python treats functions as first class citizens, you're checking if the the function exists.

Here's an example:

``` python
>>> def blah():
...    print "blah"
... 
>>> blah
<function blah at 0x7ff99e3da758>
>>> blah()
blah
>>> 
```

Note that without the `()` then we get back the function pointer.

I'm pretty sure we want `()` on these if statements.

openshift_tools/monitoring/zagg_sender.py
twiest: To be clear, it's not backwards compatible because of the _ at the front.  Backwards compatible would be to leave the name exactly the same.

Can you explain the change tracking part in more detail?

We use git for change tracking, so I'm not sure why this is needed.

twiest (dismissed):

openshift_tools/monitoring/metric_sender.py
zgalor: You're right. Got confused with ruby... :-)

openshift_tools/monitoring/zagg_sender.py
zgalor: What I meant is that I kept this so I will remember the names of the functions I need to change in the calling scripts. This is still work in progress.

openshift_tools/monitoring/generic_metric_sender.py
zgalor: Ok let's take https://github.com/openshift/openshift-tools/blob/stg/scripts/monitoring/cron-send-disk-metrics.py#L70 for example. So you want to replace this call with add_dynamic_metric. but do we want to add anther API for adding tags, and change the script to call it or do we want the hawk_sender to infer tags from the parameters sent to add_dynamic_metric?

twiest: @yaacov correct me if I'm wrong on this, but what we discussed before, is to have a separate file that would contain a key/value pair of item name and all tags that should be associated with it.

So IMO, a dynamic item would be that plus extra tags for items passed into the function.

Thoughts?

yaacov: a. in hawkular all metrics are dynamic, you can invent new ids/items as you go.
`send_metrics(key='some.new.key.i.just.think.about', value=42, tags={...})` is ok.
b. the new keys can have tags:
b.1. from the `send_metrics` command it'self.
b.2. or from some `rule` in the config file, for example:

```
some.new.key.*:
   tags:
      type: 'node'
```

---

I think a nice algorithm will be:

a. use [optional] tags given by user when sending metrics:
    for example: `send_metric(key = 'cpu.usage', value = 2, tags = {type:'node', hostname: ... })`
b. use separate file to fill in missing tags if not given by the `send_metric` command.
    for example, if the config file has the lines:

```
cpu.usage:
  tags:
     units: 'cpus'
```

   the `send_metric` command will _also_ update the `units` tag

openshift_tools/monitoring/hawk_client.py
zgalor: **add_metric** in **hawk_client** (and also in zagg_client) is called by hawk_sender.send_metrics. This sends the already stored metrics which already holds the host data determined by synthetic param in **hawk_sender.add_metric**.
**HawkClient** is not in the class hierarchy of **MetricSender**, but it also has **add_metric** method which  actually sends the metrics to hawk/zabbix. That why it's confusing. I'll rename it to **push_metrics**, i think it will be much clearer.

zgalor:
@twiest I made the changes you requested, also ran pylint and fixed some lint warnings.
I already transformed half of the scripts in scripts/monitoring/ to use metric_sender.
Still testing. Still not sending tags.

Few questions:
1. Should scripts under https://github.com/openshift/openshift-tools/tree/stg/scripts/cloud/aws/ and https://github.com/openshift/openshift-tools/tree/stg/scripts/cloud/gcp/ also send to hawkular in addition to zabbix?
2. I see some shell scripts in https://github.com/openshift/openshift-tools/tree/stg/scripts/monitoring/ that use ops-zagg-client. I also see wide usage of ops-zagg-metric-processor and other programs that use **ZaggSender** and **ZaggClient**. Should all those be expended to send to hawkular? do you want a cli that wraps **MetricSender** class? If so this might be a bit more complex to configure. It will require a duplication of the set of zabbix params to hawkular.

zgalor:
Updated all cron scripts to use MetricSender. Tested all active cron scripts.

built 2 cli tools for bash scripts:
ops-metric-client (to replace ops-zagg-client) - Still having issues with heartbeat sending
ops-metric-pcp-client (to replace ops-zagg-pcp-client) - Still needs fixing.

zgalor:
I removed the wip label and updated PR description.
@twiest could you please review. Also let me know if you want ansible scripts and docker start.sh updated with the ability to send to hawk.

@dkorn could you please review changes made in https://github.com/zgalor/openshift-tools/tree/27bd5ec1afb0adc449e8b86bc252fe2144afcaf9/openshift_tools/monitoring package:
- https://github.com/zgalor/openshift-tools/blob/27bd5ec1afb0adc449e8b86bc252fe2144afcaf9/openshift_tools/monitoring/generic_metric_sender.py
- https://github.com/zgalor/openshift-tools/blob/27bd5ec1afb0adc449e8b86bc252fe2144afcaf9/openshift_tools/monitoring/metric_sender.py
- https://github.com/zgalor/openshift-tools/blob/27bd5ec1afb0adc449e8b86bc252fe2144afcaf9/openshift_tools/monitoring/zagg_sender.py
- https://github.com/zgalor/openshift-tools/blob/27bd5ec1afb0adc449e8b86bc252fe2144afcaf9/openshift_tools/monitoring/hawk_sender.py

Also please review:
https://github.com/zgalor/openshift-tools/blob/27bd5ec1afb0adc449e8b86bc252fe2144afcaf9/scripts/monitoring/ops-metric-client.py

zgalor:
I split this to 3 PRs to ease review. This one contains only the addition of MetricSender class hierarchy.
@twiest  Please review.

joelddiaz:
@zgalor the new files being added into openshift_tools/monitoring/\* need to be packaged up in openshift_tools/python-openshift-tools.spec

RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/generic_metric_sender.py
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/generic_metric_sender.pyc
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/generic_metric_sender.pyo
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_client.py
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_client.pyc
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_client.pyo
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_common.py
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_common.pyc
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_common.pyo
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_sender.py
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_sender.pyc
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/hawk_sender.pyo
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/metric_sender.py
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/metric_sender.pyc
   /usr/lib/python2.7/site-packages/openshift_tools/monitoring/metric_sender.pyo

zgalor:
@joelddiaz Done

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,91 @@
...
+    connection = HawkConnection(url='172.17.0.27:8000')
+
+    # Basic Auth
+    connection = HawkConnection(url='172.17.0.27:8000', user='user', passwd='password')
+
+    client = HawkClient(connection)
+    print client.add_metric(ml)
+"""
+# The hawkular client
+from hawkular.metrics import HawkularMetricsClient, MetricType
```
joelddiaz, line 31: @zgalor this import is failing on me (when trying to run /usr/bin/ops-metric-client) because 'hawkular.metrics' doesn't exist. Where is the implementation of HawkularMetricsClient?

openshift_tools/monitoring/zagg_sender.py
```diff
@@ -20,83 +20,47 @@
...
-    pass
-
-class ZaggSender(object):
+class ZaggSender(GenericMetricSender):
     """
     collect and create UniqueMetrics and send them to Zagg
     """
 
-    def __init__(self, host=None, zagg_connection=None, verbose=False, debug=False):
+    def __init__(self, config_file=None, host=None, zagg_connection=None, verbose=False, debug=False):
```
joelddiaz: Adding 'config_file' to the front of the parameter list breaks the pre-existing ops-zagg-client. Is there a reason for adding the parameter to the front?

zgalor: Good point. I moved config_file to be the last param.
I also restored:
**add_zabbix_keys** to wrap **add_metric**
**add_zabbix_dynamic_item** to wrap **add_dynamic_metric**

So that existing scripts will remain working. 
After this PR is merged, I will remove those methods in https://github.com/openshift/openshift-tools/pull/1620 which makes them obsolete.

openshift_tools/monitoring/hawk_client.py
zgalor: pip install hawkular-client

unless you have a hawkular-cleint RPM, in which case you can use it.

twiest: @zgalor it's ok to use pip in dev, but we don't commit any code that requires pip.

The hawkular-client RPM will only be available in the rhel container (not centos) since we only do publicly available sources for the centos container.

As a short term fix, since we're disabling the haw_client by default right now anyway, would you mind moving this import down into the constructor _after_ the check to see if the client is disabled?

zgalor:
In order to avoid dependency in external hawkular-client RPM I added (with the help of @yaacov) a temporary fix to import a local mini-hawkular package in another commit.
Once we have the hawkular-client RPM ready I will remove this commit.
@joelddiaz @twiest please review and see if something is missing in order to merge this.

openshift_tools/monitoring/hawk_client.py
```diff
@@ -0,0 +1,92 @@
...
+    connection = HawkConnection(url='172.17.0.27:8000')
+
+    # Basic Auth
+    connection = HawkConnection(url='172.17.0.27:8000', user='user', passwd='password')
+
+    client = HawkClient(connection)
+    print client.add_metric(ml)
+"""
+# The hawkular client
+# Temporary import from local mini-hawkular-client to avoid dependency in external package
```
yaacov: can you add a `TODO:` or `FIXME:` so it will pop up in code editors, that we need to change this once we will have an RPM ?

twiest:
ok to test

twiest:
@zgalor can you look into fixing the failing jenkins checks please?

Sorry, I meant to run that yesterday, but forgot.

We've added you and @yaacov  to the Jenkins whitelist, so it should start testing your PRs automatically.

zgalor:
@twiest I fixed the pylint issues. Can you please re-run tests?

twiest:
@zgalor as you can see from the comments, the bot is now auto-running on changes you make to this PR (and all future PRs). The bot says the tests are still failing.

Looking at the output from jenkins, and comparing it to the code in the PR, I see that Jenkins is right, there are still some pylint issues in the PR.

For instance, pylint is saying this method is missing a doc string, and it is actually missing a doc string:
https://github.com/openshift/openshift-tools/pull/1511/files#diff-de1d93fda027d0eacfec12c597c543a6R138

zgalor:
@twiest can we merge?

twiest:
@zgalor It's very close. @joelddiaz said he's going to do some final testing today.

joelddiaz (changes requested):
Need to package up the mini_hawkular files into an RPM.

Scripts fail with ImportError right now.

yaacov:
> Scripts fail with ImportError right now.

What scripts fail ? Can we run this tests in our environment to test before pushing to this PR ?
I'm sure we can add the calls to Hawkular using `urllib` without triggering  `ImportError`

> Need to package up the mini_hawkular files into an RPM.

@twiest @joelddiaz @zgalor :stuck_out_tongue_winking_eye:  Ahhh.... the idea of implementing Hawkular calls using `urllib` instead of `hawkular-client` is a temporary fix until you can pack an RPM for the compete hawkular client.
Packing the mini client (using urllib) does not solve our problem ... :sunglasses:

twiest:
@yaacov what @joelddiaz was asking is that it be added to one of our openshift tools RPMs.

The only way a library like this makes it into the container is through an RPM (either ours or external). In this case, we want it in one of our RPMs.

Specifically, I believe it should be added to this spec file:
https://github.com/openshift/openshift-tools/blob/stg/openshift_tools/python-openshift-tools.spec

To test, you can do tito test builds and ensure that your files are in the rpm.

Tito test builds are described here (specifically steps 1 and 4):
https://github.com/openshift/openshift-tools/blob/stg/docs/build.adoc#build-python-openshift-tools-rpms

Basically change directory into openshift-tools/openshift_tools (where the spec file is) and then run this:
```tito build --test --rpm --output "$HOME/.tito-build"```

And then look at the contents of the package you added it to with ```rpm -qpl <path_to_package>```

yaacov:
> Specifically, I believe it should be added to this spec file:
https://github.com/openshift/openshift-tools/blob/stg/openshift_tools/python-openshift-tools.spec

@twiest 👍 thanks :-)
//cc @zgalor

zgalor:
@twiest @joelddiaz  can I pack the mini_hawkular files inside monitoring-zagg subpackage or do I need a different subpackage for that?

joelddiaz:
@zgalor i'd say just throw it in there with the monitoring-zagg subpackage for now, and we'll deal with renaming things later

yaacov:
p.s.
@zgalor and remember to copy it to `%{buildroot}` ... :-)

zgalor:
@joelddiaz I added the files to the spec but I don't see them in the package after I run the tito build.
Can you see what I missed?
https://github.com/openshift/openshift-tools/pull/1511/commits/a69f53c629d3a2b210d0ff097d074dcd20c39e63#diff-d8f3d0203f20f09d4630606d180fe099

zgalor:
Ah! I see them now!! :-)
@joelddiaz can you please check?

joelddiaz:
@zgalor LGTM

zgalor:
Can you merge it?
@joeldiaz @twiest

joelDiaz:
@zgalor hey man what you do???? jejeje you are confused...

zgalor:
Sorry wrong @joeldiaz (no auto-complete on my phone)
The right @joelddiaz. 
:-)

joelddiaz:
@zgalor we try not to merge large(ish) changes on Fridays, but we can do this on Monday

joelDiaz:
jajaja @zgalor  dont worry man,.

zgalor:
@joelddiaz @twiest please wait with the merge.
I'm considering replacing the local import solution with a configuration dependent import of hawkular-client.

joelddiaz (approved):

zgalor:
I decided that handling ImportError when hawkular-client is missing is a much more elegant solution, and will require no code changes in the future, in case hawkular-client changes.

I only attempt to import hawkular-client in case hawk is configured active in metric_sender.yaml
If import failed print error and treat as inactive but don't fail program.

I tested all scenarios locally: package installed/not - hawk enabled/disabled.

cc @simon3z

@joelddiaz @twiest Would you please make a final review and merge?

PR #1513, opened 2016-10-31
fixed the user writing adding in gcp

mwoodson:


mwoodson:
@kwoodson approved #1512

PR #1514, opened 2016-10-31
Idempotently add or remove repo ids.

kwoodson:


kwoodson:
@twiest :+1:  in #1510

PR #1516, opened 2016-11-02
added package_update_needed and os_update_latest

mwoodson:


mwoodson:
@dak1n1 approved #1515

PR #1518, opened 2016-11-02
Adding rc for retry loops.

kwoodson:


kwoodson:
@mwoodson :+1: in #1518

PR #1522, opened 2016-11-02
add oinv to inventory_clients

blentz:


blentz:
@rhdedgar gave :+1: in #1519

PR #1526, opened 2016-11-03
[stg] use image uuid

drewandersonnz:
seems @kwoodson has a commit in prod and not stg?


kwoodson:
@drewandersonnz,

Looks like I had a merge commit that was in prod but not stage.  I have
merged the prod branch to stg and int and it looks like it picked up that
merge commit.

Please rebase your branch with prod and it should resolve.

Thanks.

drewandersonnz:
completed cleanly in https://github.com/openshift/openshift-tools/pull/1563

PR #1537, opened 2016-11-04
modify the /usr/bin/cron-send-docker-metrics path

zhiwliu:


zhiwliu:
get +1 from Drew 
https://github.com/openshift/openshift-tools/pull/1536

PR #1541, opened 2016-11-04
lets update docker, for now

mwoodson:


mwoodson:
@kwoodson approved #1540

PR #1542, opened 2016-11-04
refactor yum_repos.

kwoodson:


kwoodson:
@mwoodson :+1: in #1535

PR #1544, opened 2016-11-04
Fixing files for base docker build to use rhsm.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1543

PR #1549, opened 2016-11-07
router monitoring

joelddiaz:
add script to check health on each individual router pod
also report on whether the number of running pods (for each router DC) matches the definition in the DC


joelddiaz:
👍 #1546 in stg

PR #1551, opened 2016-11-07
fixed some output of package_update_needed

mwoodson:


mwoodson:
@kwoodson approved #1550

PR #1557, opened 2016-11-07
updated new version of 3.3 vendored

mwoodson:


mwoodson:
@kwoodson approvedc #1556

PR #1560, opened 2016-11-07
Adding entitlements back for builds.

kwoodson:


kwoodson:
@joelddiaz :+1: in #1558

PR #1564, opened 2016-11-08
changed the package cleanup option

mwoodson:


mwoodson:
@wshearn approved #1565

PR #1568, opened 2016-11-08
Fix for zabbix maintenance.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1567

PR #1570, opened 2016-11-08
enable extras repo in ops-base

joelddiaz:
just because the host has a repo enabled, doesn't mean the container has it enabled. so explicitly enable it during build.


joelddiaz:
👍 #1569 in stg

PR #1575, opened 2016-11-09
app create v2

drewandersonnz:
* use same project - unique podname
* use ocutil instead of custom oc wrapper
* exception management to allow teardown

Hoping to get a few comments and +1's here .. Its a big change! 

There is still a lot I want to do, but this should easily replace the existing code.

Need to be considerate with deployment, as the arguments have changed we need to sync this with config loop.

Yes it has been tested.

### DIFF
--- a/scripts/monitoring/cron-send-create-app.py
+++ b/scripts/monitoring/cron-send-create-app.py
@@ -9,28 +9,25 @@
 # main() function has a lot of setup and error handling
 # pylint: disable=too-many-statements
 
+# main() function raises a captured exception if there is one
+# pylint: disable=raising-bad-type
+
 # Adding the ignore because it does not like the naming of the script
 # to be different than the class name
 # pylint: disable=invalid-name
 
-import subprocess
-import json
-import time
-import re
-import urllib2
-import os
-import atexit
-import shutil
-import string
-import random
 import argparse
 import datetime
-import sys
+import random
+import string
+import time
+import urllib2
 
 # Our jenkins server does not include these rpms.
 # In the future we might move this to a container where these
 # libs might exist
 #pylint: disable=import-error
+from openshift_tools.monitoring.ocutil import OCUtil
 from openshift_tools.monitoring.zagg_sender import ZaggSender
 
 import logging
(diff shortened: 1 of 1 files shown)


drewandersonnz:
Hoping to get a few comments and +1's here .. Its a big change!

scripts/monitoring/cron-send-create-app.py
```diff
@@ -280,11 +81,22 @@ def send_zagg_data(build_ran, create_app, http_code, run_time):
         zgs.add_zabbix_keys({'openshift.master.app.create.code': http_code})
         zgs.add_zabbix_keys({'openshift.master.app.create.time': run_time})
 
+    zgs.send_metrics()
+    logger.info("Data sent to Zagg in %s seconds", str(time.time() - zgs_time))
+
+def curl(ip_addr, port, timeout=30):
+    """ Open an http connection to the url and read """
+    url = 'http://%s:%s' % (ip_addr, port)
+    logger.debug("curl(%s timeout=%ss)", url, timeout)
```
dak1n1, line 90: Nice! I think this will take care of a card I opened recently about curling the route of our test app, to ensure routers are working.
https://trello.com/c/vnkE8ZEn/2497-v3-monitoring-monitor-router-pod

scripts/monitoring/cron-send-create-app.py
```diff
@@ -297,31 +109,56 @@ def getPodStatus(pod):
...
+        project = runOCcmd_yaml("get project {}".format(config.namespace))
+        logger.debug(project)
+    except Exception:
+        pass # don't want exception if project not found
+
+    if not project:
+        try:
+            runOCcmd("new-project {}".format(config.namespace))
+            time.sleep(commandDelay)
+        except Exception:
```
dak1n1, line 140: In Preview, I believe this will have to be `oadm new-project` instead of `oc new-project`. You might want to test in dev-preview-stg to be sure.

In general, I prefer `oadm` over `oc` for creating new projects, because using `oc` changes the current working project. (Though I know you have specified a separate kubeconfig so it shouldn't cause any problems to change the current working project.)

dak1n1:
:+1: Although I don't have in-depth knowledge of our app-create script, I think your changes will make the whole process more robust. It seems well thought-out.

scripts/monitoring/cron-send-create-app.py
drewandersonnz: This has been tested in preview, we can push oadm command through this function if required, but it does work :-)

scripts/monitoring/cron-send-create-app.py
drewandersonnz: While we _could_ use this for that purpose, currently none of the apps we build in test create a route, so we currently don't test routes.

PR #1586, opened 2016-11-10
patches to support 3.2

drewandersonnz:
* allow use of oadm command for new-project
* setup(), if not exception: test(), teardown()


BlueShells:
:+1:

PR #1589, opened 2016-11-10
Adding region to container.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #1589

PR #1592, opened 2016-11-10
correcting file name so it matches what we create from the config

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #1590

PR #1599, opened 2016-11-11
adding exception handling for new account metadata not found

rhdedgar:


scripts/iam-tools/aws_api_key_manager.py
```diff
@@ -225,19 +225,25 @@ def get_token(aws_account):
...
+                    client = boto3.client(
+                        'iam',
+                        aws_access_key_id=creds['AccessKeyId'],
+                        aws_secret_access_key=creds['SecretAccessKey'],
+                        aws_session_token=creds['SessionToken']
+                        )
+                    return client
+
+                except ValueError as client_exception:
+                    if 'Error retrieving SAML token' and 'Metadata not found'\
```
joelsmith: this looks like a synonym of `if True and 'Metadata not found' in ...`
you probably meant:
`if 'Error retrieving SAML token' in client_exception.message \
  and 'Metadata not found' in client_exception.message`

joelsmith (changes requested):
Looks good, just one minor logic error.

PR #1600, opened 2016-11-14
add region info for the s3 check

BlueShells:
this should fix the error of s3 check failed to list the info without region

### DIFF
--- a/openshift_tools/monitoring/awsutil.py
+++ b/openshift_tools/monitoring/awsutil.py
@@ -62,10 +62,13 @@ class AWSUtil(object):
 
         return results.stdout.read()
 
-    def get_bucket_list(self, verbose=False):
+    def get_bucket_list(self, verbose=False, BucketRegion=''):
         ''' Get list of all S3 buckets visible to AWS ID '''
 
-        s3ls_cmd = "aws s3api list-buckets"
+        s3ls_cmd = "aws s3api list-buckets  "
+        if not BucketRegion == '':
+            s3ls_cmd = "aws s3api list-buckets --region " + BucketRegion
+
         buckets = self._run_cmd(s3ls_cmd)
 
         json_list = json.loads(buckets)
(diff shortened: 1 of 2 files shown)


openshift_tools/monitoring/awsutil.py
```diff
@@ -62,10 +62,10 @@ def _run_cmd(self, cmd):
 
         return results.stdout.read()
 
-    def get_bucket_list(self, verbose=False):
+    def get_bucket_list(self, verbose=False, BucketRegion='eu-central-1'):
```
drewandersonnz: I suggest:

```
def get_bucket_list(self, verbose=False, BucketRegion=''):
    if not BucketRegion == '':
        BucketRegion = " --region " + BucketRegion
```

As this is a library function we do not want to change the default functionality.

openshift_tools/monitoring/awsutil.py
```diff
@@ -62,10 +62,10 @@ def _run_cmd(self, cmd):
...
         ''' Get list of all S3 buckets visible to AWS ID '''
 
-        s3ls_cmd = "aws s3api list-buckets"
+        s3ls_cmd = "aws s3api list-buckets --region " + BucketRegion
```
drewandersonnz: Then change this to : 

```
s3ls_cmd = "aws s3api list-buckets " + BucketRegion
```

openshift_tools/monitoring/awsutil.py
```diff
@@ -79,20 +79,20 @@ def get_bucket_list(self, verbose=False):
 
         return bucket_list
 
-    def get_bucket_info(self, s3_bucket, verbose=False):
+    def get_bucket_info(self, s3_bucket, verbose=False, BucketRegion='eu-central-1'):
```
drewandersonnz: here too

drewandersonnz:
+1, I have seen this working

PR #1602, opened 2016-11-14
Fix required variable quoting for Ansible 2.2

rharrison10:


rharrison10:
@wshearn :+1: on #1595

PR #1604, opened 2016-11-15
Fix ansible_tower role item not compatible with new Ansible

joelsmith:


joelsmith:
@zhiwliu gave 👍 in stg PR #1603

PR #1608, opened 2016-11-15
Make sure bad SSL certs won't break the routers

rharrison10:


rharrison10:
@dak1n1 :+1: on #1607

PR #1614, opened 2016-11-15
make sure latest openshift-ansible-roles is installed

mwoodson:


mwoodson:
@wshearn approved #1613

PR #1620, opened 2016-11-17
Change a few monitoring scripts to use metric_sender

zgalor:
Use MetricSender instead of ZaggSender in monitoring scripts

depends on https://github.com/openshift/openshift-tools/pull/1747


zgalor:
@twiest PTAL

zgalor:
@joelddiaz PTAL

twiest (approved):
@zgalor This looks great!

PR #1623, opened 2016-11-17
SSO: Add config template support for memcache session storage

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1622

PR #1625, opened 2016-11-17
Add generated versions of latest SSO container

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1624

PR #1626, opened 2016-11-17
adding sso service check for the oso-monitor-sso container

rhdedgar:


scripts/iam-tools/check_sso_service.py
```diff
@@ -0,0 +1,117 @@
...
+        """ Retrieves a list of the config-managed ops AWS accounts.
+
+        Returns:
+            A list containing each of the lines found in the aws accounts file
+
+        Raises:
+            A ValueError if the path does not exist
+        """
+
+        path = '/secrets/aws_accounts.txt'
```
joelsmith: How about we aim this at /etc/openshift_tools/aws_accounts.txt, then have a symlink to the secret (only in the container). That way the script will work on tower and in the container.

scripts/iam-tools/check_sso_service.py
```diff
@@ -0,0 +1,117 @@
...
+    def get_token(aws_account):
+        """ Generate temporary SSO access credentials.
+
+        Requires the config file containing the IDP hostname.
+
+        Returns:
+            A temporary boto3 client created with a session token provided by the IDP host.
+        """
+
+        ops_idp_host = 'login.ops.openshift.com'
```
joelsmith: I'd like to keep this hostname out of public repos. Could you use /etc/openshift_tools/sso-config.yaml like you do for other stuff, then we can generate that file in the container so it'll work there too?

scripts/iam-tools/check_sso_service.py
```diff
@@ -0,0 +1,117 @@
...
+            return client
+
+        except ValueError as client_exception:
+            if 'Error retrieving SAML token' in client_exception.message and \
+            'Metadata not found' in client_exception.message:
+                print(client_exception)
+                print('Metadata for %s missing or misconfigured, skipping' % aws_account)
+
+            else:
+                raise
```
joelsmith: you're never catching (later) anything else that might throw and get re-raised here, so in this error path you won't ever send. I suppose that's fine if you're planning to use ops runner to run it or whatever, so that it will be noticed if it breaks. Are you planning to use ops runner?

scripts/iam-tools/check_sso_service.py
```diff
@@ -0,0 +1,117 @@
...
+        """ Main function. """
+
+        zag = ZaggSender()
+        ops_accounts = self.check_accounts()
+        zabbix_key = "sso.iam.not.reachable"
+        key_value = 0
+
+        for account in ops_accounts:
+            account_number = account[0].split(':')[1]
+            temp_client = self.get_token(account_number)
```
joelsmith: You're not handling the "Metadata not found" path. In that path, nothing is returned byget_token(). Perhaps you could add something here to skip to the next account here, like this:

```
  if not temp_client:
    continue
```

scripts/iam-tools/check_sso_service.py
```diff
@@ -0,0 +1,117 @@
...
+                aws_secret_access_key=creds['SecretAccessKey'],
+                aws_session_token=creds['SessionToken']
+                )
+
+            return client
+
+        except ValueError as client_exception:
+            if 'Error retrieving SAML token' in client_exception.message and \
+            'Metadata not found' in client_exception.message:
+                print(client_exception)
```
joelsmith: I don't think it's especially useful to print the client_exception here. The below print gives a nice-enough summary and is much less noisy.

scripts/iam-tools/check_sso_service.py
```diff
@@ -0,0 +1,117 @@
...
+        ops_accounts = self.check_accounts()
+        zabbix_key = "sso.iam.not.reachable"
+        key_value = 0
+
+        for account in ops_accounts:
+            account_number = account[0].split(':')[1]
+            temp_client = self.get_token(account_number)
+            acc_status = temp_client.get_role(RoleName='iam_monitoring')
+
+            if acc_status['ResponseMetadata']['HTTPStatusCode'] != 200:
```
joelsmith: A few things here...
1. Whenever you increment your counter, you need to print out why you did, so that whoever is investigating the alert can see what the source of the alert was.
2.  If boto blows up trying to get the role, we need to catch that exception and increment the counter. That's indicative of a misconfigured AWS account
3. I think it's worth at least doing a cursory check to make sure that the requested data is returned.

So here's my suggestion for this section:

```
            try:
                acc_status = temp_client.get_role(RoleName='iam_monitoring')

                if acc_status['ResponseMetadata']['HTTPStatusCode'] != 200:
                    print("HTTP request failed on account %s (%s)" % (account_name, account_number))
                    key_value += 1
                if not acc_status['Role']['AssumeRolePolicyDocument']:
                    print("No policy document returned for account %s (%s)" % (account_name, account_number))
                    key_value += 1

            except botocore.exceptions.ClientError as boto_exception:
                print("Failed on account %s (%s) due to exception: %s" %
                        (account_name, account_number, str(boto_exception)))
                key_value += 1
```

joelsmith (changes requested):

scripts/iam-tools/check_sso_service.py
```diff
@@ -0,0 +1,117 @@
...
+    def main(self):
+        """ Main function. """
+
+        zag = ZaggSender()
+        ops_accounts = self.check_accounts()
+        zabbix_key = "sso.iam.not.reachable"
+        key_value = 0
+
+        for account in ops_accounts:
+            account_number = account[0].split(':')[1]
```
joelsmith: I don't think account is an array, so it should be:

```
account_number = account.split(':')[1] 
```

Also, my other example code used account_name in addition to account_number to make it easier to diagnose problems that this check will catch. So better yet,

```
account_name, account_number = account.split(':')
```

joelsmith (approved):
👍 nice

PR #1632, opened 2016-11-17
Allow openshift_cmd to take input to pipe to oc's stdin

joelsmith:
oc_process: also, optionally accept template data via "content" instead of
via imported template.


joelsmith:
@kwoodson gave 👍 in stg PR #1628

PR #1634, opened 2016-11-17
Ignore auto-generated values when comparing resources in oc_process

joelsmith:


joelsmith:
@kwoodson gave 👍 in stg PR #1633

PR #1635, opened 2016-11-18
set ReclaimPolicy to Delete so controller will delete used volumes

dak1n1:


robotmaxtron (approved):
:+1:

rharrison10 (approved):
Clearly a simple policy change

PR #1637, opened 2016-11-21
Fix return code getting overwritten

wshearn:


yocum137:
:+1: to be pushed after the holiday code freeze.

PR #1640, opened 2016-11-24
Add tags to metric sender

zgalor:
Add tags to MetricSender.
Tags can be set using regular expression in metric_sender.yaml or as parameter in add_metric method.

Depends on https://github.com/openshift/openshift-tools/pull/1511

### DIFF
--- a/openshift_tools/monitoring/generic_metric_sender.py
+++ b/openshift_tools/monitoring/generic_metric_sender.py
@@ -44,7 +44,7 @@ class GenericMetricSender(object):
         """ empty implementation overridden by derived classes """
         pass
 
-    def add_metric(self, metric, host=None, synthetic=False):
+    def add_metric(self, metric, host=None, synthetic=False, key_tags=None):
         """ empty implementation overridden by derived classes """
         pass
 
--- a/openshift_tools/monitoring/hawk_sender.py
+++ b/openshift_tools/monitoring/hawk_sender.py
@@ -19,7 +19,7 @@ Examples:
     zs.add_zabbix_keys({ 'test.key' : '1' })
     zs.send_metrics()
 """
-
+import re
 from openshift_tools.monitoring.metricmanager import UniqueMetric
 from openshift_tools.monitoring.hawk_client import HawkClient
 from openshift_tools.monitoring.hawk_common import HawkConnection
@@ -90,8 +90,11 @@ class HawkSender(GenericMetricSender):
 
         return hawk_connection
 
-    def add_metric(self, metrics, host=None, synthetic=False):
+    def add_metric(self, metrics, host=None, synthetic=False, key_tags=None):
         """ create unique metric from key value pair """
+
+        if not key_tags:
+            key_tags = {}
 
         if synthetic and not host:
             host = self.config['synthetic_clusterwide']['host']['name']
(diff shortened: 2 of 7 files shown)


zgalor:
@yaacov @joelddiaz PTAL
This is based on https://github.com/openshift/openshift-tools/pull/1511, so only the last commit needs review

zgalor:
Tested all scenarios locally.

yaacov:
LGTM 👍

zgalor:
@joelddiaz please review

openshift_tools/monitoring/generic_metric_sender.py
```diff
@@ -44,7 +44,9 @@ def add_dynamic_metric(self, discovery_key, macro_string, macro_array, host=None
         """ empty implementation overridden by derived classes """
         pass
 
-    def add_metric(self, metric, host=None, synthetic=False):
+    # key_tags is not changed inside method. disabling warning
+    # pylint: disable=dangerous-default-value
+    def add_metric(self, metric, host=None, synthetic=False, key_tags={}):
```
twiest: Same suggestion as below (set key_tags to None and then check that it's set).

openshift_tools/monitoring/metricmanager.py
```diff
@@ -64,9 +64,10 @@ class UniqueMetric(zbxsend.Metric):
     # Reason: disable pylint too-many-arguments because this is a data only
     #         object, and we like to use the constructor to populate the
     #         data easily.
+    #         Sending tags={} is not dangerous here because __init__ will only be called once.
     # Status: permanently disabled.
-    # pylint: disable=too-many-arguments
-    def __init__(self, host, key, value, clock=None, unique_id=None):
+    # pylint: disable=too-many-arguments,dangerous-default-value
+    def __init__(self, host, key, value, clock=None, unique_id=None, tags={}):
```
twiest: Same as above, consider tags=None pattern.

openshift_tools/monitoring/zagg_sender.py
```diff
@@ -101,7 +101,9 @@ def add_heartbeat(self, heartbeat, host=None):
                                                  )
         self.unique_metrics.append(hb_metric)
 
-    def add_metric(self, metrics, host=None, synthetic=False):
+    # key_tags is not changed inside method. disabling warning
+    # pylint: disable=dangerous-default-value
+    def add_metric(self, metrics, host=None, synthetic=False, key_tags={}):
```
twiest: Same

openshift_tools/monitoring/hawk_sender.py
```diff
@@ -90,7 +90,9 @@ def _get_default_hawk_connection(self):
 
         return hawk_connection
 
-    def add_metric(self, metrics, host=None, synthetic=False):
+    # key_tags is not changed inside method. disabling warning
+    # pylint: disable=dangerous-default-value
+    def add_metric(self, metrics, host=None, synthetic=False, key_tags={}):
```
twiest: That's true that it's not changing right now, but it could change in a future PR. That future dev might not know about python's weird default value behavior and so won't change it then. That would cause weird, unexpected behavior.

We try to never disable pylint checks, especially when the fix is so easy.

Consider changing this to:
```python
     def add_metric(self, metrics, host=None, synthetic=False, key_tags=None):
￼          """ create unique metric from key value pair """
￼  
          if not key_tags:
              key_tags = {}

￼          if synthetic and not host:
```

twiest (changes requested):
This all looks good except the pylint disables.

Pylint disables are red flags to us and always bring extra scrutiny. In this case, I don't believe we should be using pylint disables.

zgalor:
@twiest good point regarding the default and the warning.
Fixed and tested locally. Please confirm

twiest (approved):
Looks great!

Thanks for working on this for us.

PR #1643, opened 2016-11-28
Add auto-deploy role for SSO app

joelsmith:


joelsmith:
@kwoodson approved changes in #1642

PR #1645, opened 2016-11-28
wait rather than exit while the iptables lock is held

joelddiaz:
rather than wait forever, bound the wait to 10 minutes


joelddiaz:
👍 #1644 in stg

PR #1646, opened 2016-11-28
Fix default edit actions for router creation

rharrison10:


rharrison10:
@joelddiaz :+1: on #1639

PR #1647, opened 2016-11-28
Fix return code getting overwritten

wshearn:


wshearn:
Approved in #1637 by @yocum137

PR #1649, opened 2016-11-28
adding scripts to iam-tools package

rhdedgar:
Check tested with Zabbix.

### DIFF
--- a/scripts/iam-tools/check_sso_http_status.py
+++ b/scripts/iam-tools/check_sso_http_status.py
@@ -1,0 +1,73 @@
+#!/bin/env python
+# vim: expandtab:tabstop=4:shiftwidth=4
+
+"""
+ This script is used to check the return code from /status.php IDP host and report to Zabbix.
+ It will report an issue to Zabbix if any received HTTP status codes != 200.
+"""
+
+from __future__ import print_function
+
+import os
+import requests
+import yaml
+
+# Reason: disable pylint import-error because our modules aren't loaded on jenkins.
+# pylint: disable=import-error
+from openshift_tools.monitoring.zagg_sender import ZaggSender
+
+
+class CheckStatus(object):
+    """ Class to check HTTP status code from IDP host. """
+
+
+    def __init__(self):
+        self.sso_status_code = None
+
+
+    @staticmethod
+    def check_http(host):
+        """ Check HTTP status codes returned by the IDP host. """
+
+        url_path = 'https://' + host + '/status.php'
+
+        try:
+            check_status = requests.get(url_path)
+            sso_status_code = check_status.status_code
+            return sso_status_code
+
+        except requests.exceptions.RequestException as request_exception:
+            print(request_exception)
+            return None
+
+
+    def main(self):
+        """ Main function. """
+
+        zag = ZaggSender()
+        yaml_config = {}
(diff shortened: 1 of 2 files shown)


scripts/iam-tools/check_sso_http_status.py
```diff
@@ -0,0 +1,63 @@
...
+        sso_status_code = check_status.status_code
+
+        return sso_status_code
+
+
+    def main(self):
+        """ Main function. """
+
+        zag = ZaggSender()
+        zabbix_key = "sso.http.not.reachable"
```
joelsmith: we need this code to do 2 different checks. One of the actual service and one of the other (main) container in the pod. It'll take a little moving around (like moving the config file reader above the loop, etc., but this is basically what I had in mind:
```
checks = {
  "sso.service.not.reachable": yaml_config["idp_host"],
  "sso.container.not.reachable": "127.0.0.1:8443"
}
for zabbix_key, host in checks.iteritems():
   # ... do checks, pass in host to self.check_http()
```

joelsmith (changes requested):
Other than needing to add the other (localhost) check, it looks good.

scripts/iam-tools/check_sso_http_status.py
```diff
@@ -20,42 +20,52 @@
...
+        try:
+            check_status = requests.get(url_path)
+            sso_status_code = check_status.status_code
+            print(sso_status_code, 'status code for', host)
+            return sso_status_code
 
-        return sso_status_code
+        except requests.exceptions.RequestException as request_exception:
+            print(request_exception)
+            return self.sso_status_code
```
joelsmith: consider just "return None". self.sso_status_code seems useless and having both a self.sso_status_code and a sso_status_code seems needlessly confusing.

scripts/iam-tools/check_sso_http_status.py
```diff
@@ -20,42 +20,53 @@
...
+            print(request_exception)
+            return None
 
 
     def main(self):
         """ Main function. """
 
         zag = ZaggSender()
-        zabbix_key = "sso.http.not.reachable"
         key_value = 0
```
joelsmith: This needs to be inside the loop or a failure of check 1 will automatically mean a failure of check 2. We don't want that kind of cascading failure since the containers might be fine but the service/route might be broken

scripts/iam-tools/check_sso_http_status.py
```diff
@@ -65,8 +65,8 @@ def main(self):
             if sso_status != 200:
                 key_value += 1
 
-            zag.add_zabbix_keys({zabbix_key: key_value})
-            zag.send_metrics()
+                zag.add_zabbix_keys({zabbix_key: key_value})
+                zag.send_metrics()
```
joelsmith, line 69: This looks wrong to me. Seems like these were already inside the loop, and now they'll only send when there's a failure. My previous change request was to move the key_value = 0 initialization inside the loop so that the key_value += 1 that could happen on the first loop pass won't affect the second check's value. Sorry if the GH review format didn't lend itself to knowing which line I was talking about.

joelsmith (approved):

joelsmith:
Looks good. Feel free to squash your fix commits into your main commit.

PR #1653, opened 2016-11-29
[WIP] Modify PR validation tests to run in a container on openshift

tiwillia:
NOTE: This pull request should not be merged until we are ready to stop using ci.openshift.redhat.com and have moved to using our own jenkins environment running on an openshift instance.

The three tests previously run to validate a pull request have been merged into a single test - pr_validation.py.

A Dockerfile has been added, as well as a buildConfig and a pod json. The image built will sleep forever when run, waiting for commands to be run on the container itself. A single environment variable is provided, TOOLS_PULL_ID, that is used by the pr_validation.py test. The test will reach out to github and pull all details for the pull request with the matching ID.

The container was designed to be created for a specific pull request, have something run pr_validation.py and consume the output, then have the container deleted. Specifically, this should be used alongside https://github.com/tiwillia/jenkins-test-pull-requests


tiwillia:
Tests are expected to fail, since the current tests use scripts that will no longer exist with this PR

PR #1655, opened 2016-11-29
Make a way to get file contents from aws_account_list role

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1654

PR #1658, opened 2016-11-29
updated the aws elb instance tries

mwoodson:


mwoodson:
@kwoodson approved #1652

PR #1663, opened 2016-11-30
Allow OO_PAUSE_ON_{START,BUILD} pauses to be cancelled

joelsmith:
With the "sleep 10" becoming the while condition, one can kill the
sleep to cause the container or build to resume.


joelsmith:
@BlueShells gave 👍 in stg PR #1662

joelsmith:
ok to test (since I forgot to merge the stg PR)

PR #1666, opened 2016-11-30
Fix improperly formated keys in multi_inventory

rharrison10:


rharrison10:
@mwoodson :+1: on #1665

PR #1667, opened 2016-11-30
Add a pause to allow curator edits to finish before redeploying

rharrison10:


rharrison10:
@wshearn :+1: on #1641

PR #1668, opened 2016-11-30
add role to move PVs from one AZ to another

joelddiaz:
this includes an ansible module that handles the EC2 side of things
extend oc_edit module to allow editing PV objects


ansible/roles/lib_utils/library/oo_ec2_elb_migrate.py
```diff
@@ -0,0 +1,194 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# Jenkins environment doesn't have all the required libraries
+# pylint: disable=import-error
+import boto3
+# Ansible modules need this wildcard import
+# pylint: disable=unused-wildcard-import, wildcard-import, redefined-builtin
+from ansible.module_utils.basic import *
```
kwoodson: Please place this under the if __name__ statement.

kwoodson (approved):

PR #1670, opened 2016-12-01
Move generated "secrets" to emptydir volume in SSO app

joelsmith:
This is required for the OpenShift 3.2->3.3 upgrade since
the secrets plugin nukes the secrets volume if any changes are made to it


joelsmith:
@BlueShells gave 👍 in stg PR #1669

PR #1671, opened 2016-12-01
Add monitoring container to SSO app template

joelsmith:


joelsmith:
@twiest gave a 👍 in stg PR #1664

PR #1672, opened 2016-12-01
Switch line endings from DOS to UNIX for saml_aws_creds.py

joelsmith:


drewandersonnz:
whitespace change only, tested by copy/paste into meld .. +1

PR #1673, opened 2016-12-01
Switch line endings from DOS to UNIX for saml_aws_creds.py

joelsmith:


joelsmith:
@drewandersonnz gave 👍 in stg PR #1672

PR #1677, opened 2016-12-01
adding sso checks to zabbix

rhdedgar:


ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  zbx_host:
+    zbx_server: "{{ ozcs_zbx_server }}"
+    zbx_user: "{{ ozcs_zbx_user }}"
+    zbx_password: "{{ ozcs_zbx_password }}"
+    state: list
+    hostname: "{{ oss_clusterid }}-synthetic"
+  register: out
+
+- name: "item: oso.sso.app.aggregate"
+  when: out.results.results | length == 1
```
joelsmith: I'd prefer "{{ out.results.results | length == 1 }}" since the syntax works more reliably across various ansible versions

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: "item: openshift.master.cluster.synthetic.exists"
+  zbx_host:
+    zbx_server: "{{ ozcs_zbx_server }}"
+    zbx_user: "{{ ozcs_zbx_user }}"
+    zbx_password: "{{ ozcs_zbx_password }}"
```
joelsmith: we don't want to re-use role variable names from another role. That's because there may be cases where you want to pass one set of vars to one role and a different set of vars to another. If we re-use the names, the two have their inputs locked together. So you'll need to use something besides 'ozcs_' for a prefix for all role vars. I think I use `osso_` for the openshift_sso_app role, so maybe you could use `ossoz_` or `ozsso_` or something.

ansible/roles/openshift_sso_app_zabbix/README.md
```diff
@@ -0,0 +1,34 @@
...
+------------
+
+Ansible Modules:
+
+- tools_roles/lib_openshift_3.2
+
+
+Role Variables
+--------------
+
```
joelsmith, line 16: Be sure and all all your prefixed role variables that need to be passed in to the Role Variables section. You can see an example at https://github.com/openshift/openshift-tools/blob/master/ansible/roles/openshift_sso_app/README.md

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    zabbix_type: aggregate
+    interval: 900
+    delta: 0
+    units: '%'
+  with_items: "{{ ozcs_masters }}"
+  delegate_to: "{{ oss_clusterid }}-synthetic"
+  run_once: true
+
+- name: "SSO containers missing information for more than 15 minutes"
+  expression: "{ops-aws-synthetic:grpsum[ops-aws,sso.monitoring.container.running,sum,15m].min(#3)}<0"
```
joelsmith: I'm not sure how often zabbix aggregate item datapoints get added, but it might be more meaningful to have it be across a time frame instead of a number of samples. Also, we want the alert to happen if we don't have 3 containers. This check is supposed to detect when one or more of the required 3 isn't running. So maybe .max(1800) } < 3. That would say that if during the entire previous half hour, there were never 3 containers running, alert.

ansible/roles/os_zabbix/tasks/main.yml
```diff
@@ -70,6 +70,10 @@
   tags:
   - openshift_logging
 
+- include_vars: template_ops_sso.yml
+  tags:
+  - sso
```
joelsmith: I currently have some `sso` tags and some `sso_app` tags. Currently, the distinction is, if it's setting up the app (like this), it should use `sso_app`, and if it's setting up the AWS side, `sso`. But maybe we should just use `sso_app` across the board.

joelsmith (changes requested):
Looks pretty good, just a couple of minor things to look at.

ansible/roles/openshift_sso_app_zabbix/meta/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  author: OpenShift Ops
+  description: Openshift Operations SSO Zabbix Role
+  company: Red Hat, Inc
+  license: ASL 2.0
+  min_ansible_version: 2.0
+  platforms:
+  - name: EL
+    versions:
+    - 7
+dependencies:
```
rhdedgar, line 12: I could specify these in ops_tech_profiles/sso_app/meta/main.yml instead.

ansible/roles/openshift_sso_app_zabbix/README.md
```diff
@@ -18,6 +18,12 @@ Role Variables
 Dependencies
 ------------
 
+- `ossoz_zbx_server`: The name of the Zabbix server
+- `ossoz_zbx_user`: The user with which to authenticate to the Zabbix server
+- `ossoz_zbx_password`: The password of the user
+- `ossoz_masters`: The list of masters in the cluster
+- `ossoz_clusterid`: The id of the cluster
+
```
joelsmith, line 21: sorry, GH comment ambiguity strikes again. I meant that these should go under "Role Variables"

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -1,19 +1,19 @@
...
   zbx_host:
-    zbx_server: "{{ ozcs_zbx_server }}"
-    zbx_user: "{{ ozcs_zbx_user }}"
-    zbx_password: "{{ ozcs_zbx_password }}"
+    zbx_server: "{{ ossoz_zbx_server }}"
+    zbx_user: "{{ ossoz_zbx_user }}"
+    zbx_password: "{{ ossoz_zbx_password }}"
     state: list
     hostname: "{{ oss_clusterid }}-synthetic"
   register: out
```
joelsmith, line 12: this zbx_host should also have a run_once: true on it. doesn't hurt anything if it doesn't have it, but runs a little faster and quieter with it.

ansible/roles/openshift_sso_app_zabbix/meta/main.yml
```diff
@@ -0,0 +1,19 @@
...
+- role: tools_roles/lib_openshift_3.2
+- role: tools_roles/os_zabbix_cluster_stats
```
joelsmith: this seems like a mistake. that's probably what your other comment meant about putting it in the tech profile, and I agree, but take care that you call your new role, and not os_zabbix_cluster_stats

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    hostname: "{{ item }}"
+    zabbix_type: aggregate
+    interval: 900
+    delta: 0
+    units: '%'
+  with_items: "{{ ossoz_masters }}"
+  delegate_to: "{{ oss_clusterid }}-synthetic"
+  run_once: true
+
+- name: "SSO containers missing information for more than 15 minutes"
```
joelsmith: This name should probably change to say "Add alert for when not enough SSO containers are running for 30 minutes" or something like that.

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    data_type: decimal
+    description: "The sum of oso-monitor-sso containers checking in."
+    multiplier: ''
+    hostname: "{{ item }}"
+    zabbix_type: aggregate
+    interval: 900
+    delta: 0
+    units: '%'
+  with_items: "{{ ossoz_masters }}"
+  delegate_to: "{{ oss_clusterid }}-synthetic"
```
joelsmith: See above comment. Definitely don't delegate_to the synthetic host since it doesn't exist and ansible won't be able to ssh to it to run this task.

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    zbx_server: "{{ ossoz_zbx_server }}"
+    zbx_user: "{{ ossoz_zbx_user }}"
+    zbx_password: "{{ ossoz_zbx_password }}"
+    name: oso.sso.app.aggregate
+    key: 'grpsum[ops-aws,sso.monitoring.container.running,sum,15m]'
+    value_type: int
+    data_type: decimal
+    description: "The sum of oso-monitor-sso containers checking in."
+    multiplier: ''
+    hostname: "{{ item }}"
```
joelsmith: You want the hostname to be {{ oss_clusterid }}-synthetic. Also, I don't think you need to run it in a with_items loop on all the masters, so you can probably remove the ossoz_masters variable entirely. You could delegate to localhost if you want, but which host the zabbix api call (to add the item) is made from isn't that important. It's important that it happen just once (so that there aren't two hosts trying to add it at the same time), so I think that's why we run it with the run_once: true

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    delta: 0
+    units: '%'
+  with_items: "{{ ossoz_masters }}"
+  delegate_to: "{{ oss_clusterid }}-synthetic"
+  run_once: true
+
+- name: "SSO containers missing information for more than 15 minutes"
+  expression: "{ops-aws-synthetic:grpsum[ops-aws,sso.monitoring.container.running,sum,15m].max(1800)}<3"
+  url: "https://github.com/openshift/ops-sop/blob/master/V3/Alerts/check_ops_sso.asciidoc"
+  priority: warning
```
joelsmith: probably want a run_once: true here too.

ansible/roles/openshift_sso_app_zabbix/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    zabbix_type: aggregate
+    interval: 900
+    delta: 0
+    units: '%'
+  with_items: "{{ ossoz_masters }}"
+  delegate_to: "{{ oss_clusterid }}-synthetic"
+  run_once: true
+
+- name: "SSO containers missing information for more than 15 minutes"
+  expression: "{ops-aws-synthetic:grpsum[ops-aws,sso.monitoring.container.running,sum,15m].max(1800)}<3"
```
joelsmith: seems like this line and the next two should be indented and underneath a zbx_* module call. I didn't look at the API, but should it be `- zbx_trigger`

joelsmith (changes requested):
It looks like it's getting close enough to test once you make these few fixes. Sorry I didn't catch them on the first go-around. There are several instances of the `- sso` tag in this PR, so you'll want them all to be consistent. Finally, you should probably start adding the code to the ops_tech_profile so that it calls your new role, and you can test it all out, end-to-end from `openshift-ansible-ops/playbooks/site/ops-aws/` with:

```
./openshift_master.yml -e g_branch=prod -t sso_app
```
Just make sure that you've rebased to a recent prod in your openshift-tools and openshift-ansible-ops, and pulled recently from openshift-ansible-private where you should have the prod branch checked out. Otherwise you'll update the SSO container with old data for the secrets.

joelsmith (approved):
👍 looks good to go!

PR #1680, opened 2016-12-01
Minor fixes to openshift-tools-scripts-iam-tools RPM

joelsmith:
* Hide extensions for scripts installed to /usr/local/bin
* Put python module in python module directory
* When run in container, have check_sso_service.py use ssh key from secrets


joelsmith:
@ivanhorvath approved in stg PR #1678

PR #1681, opened 2016-12-01
Add at-login AWS credential checker script

joelsmith:


scripts/iam-tools/aws_creds_check.sh
```diff
@@ -0,0 +1,115 @@
...
+# There are several places where -n "$ZSH_NAME" is tested to see if we're running zsh, and then
+# the zsh version of the syntax is used instead of the bash version.
+
+# only bash v4+ and zsh are supported. do nothing for other shells
+if [ -n "$BASH_VERSION" -a "$BASH_VERSINFO" -gt 3 -o "$ZSH_NAME" ]; then
+
+function _check_creds()
+{
+    local -A accounts
+    local account_count creds_actual name num line regex now credfile=~/.aws/credentials expfile=~/.aws/credentials_expiration
```
blentz: You're mixing variable assignments with variable scope declarations, which hampers readability.

I recommend altering this to split up the two tasks:

`local account_count creds_actual name num line <etc.>`
`local credfile=/some/file expfile=/other/file`

scripts/iam-tools/aws_creds_check.sh
```diff
@@ -0,0 +1,115 @@
...
+
+    if ! private_encfs_mounted; then
+        echo "~/.private not mounted. Skipping AWS credential check"
+        return
+    fi
+
+    # if ~/.private/.aws doesn't exist, create it
+    test -d ~/.private/.aws || mkdir ~/.private/.aws
+    # if ~/.private/.aws/credentials doesn't exist, create it
+    test -f ~/.private/.aws/credentials || { touch ~/.private/.aws/credentials; chmod 600 ~/.private/.aws/credentials; }
```
blentz: For the sake of readability, I recommend using the conditional test syntax.  Also, avoid putting multiple statements on a single line.

`if [ ! -f /the/creds ]; then`
` touch /the/creds `
` chmod 600 /the/creds `
`fi`

joelsmith:
[test]

scripts/iam-tools/aws_creds_check.sh
```diff
@@ -0,0 +1,115 @@
...
+    [ -f "$expfile" ] && read -r line < "$expfile"
+    line=$((line+0))
+    if [ -z "$ZSH_NAME" ]; then
+        #    like 'date +%s' without a fork/exec
+        now="$(printf '%(%s)T' -1)"
+    else
+        # zsh's printf doesn't have time formatters so we're forced to fork/exec date
+        now="$(date +%s)"
+    fi
+    if [ $((now + 60 * 60 * 24 * 7)) -gt "$line" ]; then
```
blentz: Because you're doing the same date math several times in a row, it'd make this clearer to abstract those out into variables.

`HOUR=60*60`
`DAY=${HOUR}*24`
`WEEK=${DAY}*7`
`if [ $((now+${WEEK})) -gt "$line" ]; then`

scripts/iam-tools/aws_creds_check.sh
```diff
@@ -0,0 +1,115 @@
...
+            for name in "${!accounts[@]}"; do
+                echo -n " -p '$name'"
+            done
+        fi
+        echo
+    fi
+}
+
+# Only bother to check aws credentials for non-root users with UID<1000 who are members of libra_ops
+if [ $UID -lt 1000 -a $UID -ne 0 ] && id -nG | grep -qw libra_ops; then
```
blentz, line 118: Does the uid<1000 part of the check really matter?  Why wouldn't we just care about non-root users who are in the libra_ops group?

scripts/iam-tools/aws_creds_check.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+
```
blentz, line 2: There should be a comment block at the top detailing what this script's purpose is so that future readers don't have to read the whole script just to get an idea of what it's supposed to be doing and why. It doesn't need to be a novel. Just a couple sentences should be fine.

Also, because this is in a public repo, it should have the ASL 2.0 boilerplate in it, too.

blentz (approved):
Overall, I think the script is fine. There's some stylistic issues I'd like to see considered, but nothing that's a show-stopper.

So, :+1:

scripts/iam-tools/aws_creds_check.sh
joelsmith: We put non-human users above UID 1000. Some of these are also part of libra_ops, e.g. opsmedic and autokeys. I don't ever want this stuff running for them. I'll add a comment to that effect.

scripts/iam-tools/aws_creds_check.sh
```diff
@@ -10,21 +16,30 @@ if [ -n "$BASH_VERSION" -a "$BASH_VERSINFO" -gt 3 -o "$ZSH_NAME" ]; then
...
         echo "~/.private not mounted. Skipping AWS credential check"
         return
     fi
 
     # if ~/.private/.aws doesn't exist, create it
-    test -d ~/.private/.aws || mkdir ~/.private/.aws
+    [ -d ~/.private/.aws ] || mkdir ~/.private/.aws
     # if ~/.private/.aws/credentials doesn't exist, create it
-    test -f ~/.private/.aws/credentials || { touch ~/.private/.aws/credentials; chmod 600 ~/.private/.aws/credentials; }
+    if  [ -f ~/.private/.aws/credentials ]; then
```
blentz: You reversed the logic here. I don't think that was the intent, was it?

joelsmith: good catch

PR #1682, opened 2016-12-01
Fix metrics route and image settings

dak1n1:
Use the proper CA cert instead of master's self-signed cert.
Updated path to template.
Use registry.ops for metrics image.


rharrison10 (approved):
Great catch

dak1n1:
Works with openshift-ansible-roles-3.3.55-1 and above.

PR #1684, opened 2016-12-01
Add aws_creds_check to openshift-tools-scripts-iam-tools RPM

joelsmith:
Also add ASL 2.0 licence file to repo root and ASL 2.0 header
to iam-tools scripts


blentz:
:+1:

PR #1688, opened 2016-12-01
add role to move PVs from one AZ to another

joelddiaz:
this includes an ansible module that handles the EC2 side of things
extend oc_edit module to allow editing PV objects


joelddiaz:
👍 #1668 in stg

PR #1690, opened 2016-12-01
switch ops-run-in-loop wrapper script from eval to running args directly

joelsmith:


joelsmith:
@twiest gave 👍 in stg PR #1689

PR #1694, opened 2016-12-02
Skip SSL verification step from SSO monitoring container

joelsmith:


rhdedgar:
need to disable the pylint warning on requests.packages.urllib3.exceptions

rhdedgar:
👍

PR #1696, opened 2016-12-02
Skip SSL verification step from SSO monitoring container

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1694

PR #1699, opened 2016-12-02
Print monitoring container errors also to stdout, prepend time stamp

joelsmith:


joelsmith:
@rhdedgar gave 👍 in #1698

PR #1714, opened 2016-12-05
Add authorized_keys file for SSO monitoring container

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1713

PR #1715, opened 2016-12-06
Adding new user ops_monitoring

mmahut:
Add new user ops_monitoring and generate its credential to a file.

### DIFF
--- a/ansible/roles/oso_host_monitoring/README.md
+++ b/ansible/roles/oso_host_monitoring/README.md
@@ -104,6 +104,14 @@ The AWS access key to use for the snapshotter
 
 The AWS secret access key to use for the snapshotter
 
+### osohm_ops_monitoring_aws_access_key_id
+
+The AWS access key to use for monitoring scripts
+
+### osohm_ops_monitoring_aws_secret_access_key
+
+The AWS secret access key to use for monitoring scripts
+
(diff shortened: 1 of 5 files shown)


mmahut:
Related https://github.com/openshift/openshift-ansible-ops/pull/1804

mwoodson:
I'm good with this.  I'd like other's to take a look.  +1 from me, but I'd like other +1. 

@twiest @ivanhorvath @kwoodson @joelddiaz Thoughts?

ivanhorvath:
So this user would have different access level than the account owner? I thought we already can use that with boto from host monitoring container to grab info from aws. Is that not so?

kwoodson:
@mmahut 
What's this user for?  I'm generally fine with adding him to the container if there is a good reason.  The location for the file seems fine.

I would definitely test and verify that the snapshotting continues to work when adding creds to this file.

mmahut:
@kwoodson @ivanhorvath In my case this user will be used to check the state of the ELBs and report if any of the nodes is missing from the ELB and report it to zabbix. For my case, this user will have only read-only access to the ELB information (elasticloadbalancing:DescribeListeners) for now, however the idea is to add new policies to this user, once Joel reviews them, in case we need other access to AWS from the monitoring container (for example to monitor security groups etc).

PR #1719, opened 2016-12-06
SSO monitoring: fix heartbeat/check race with short initial sleep

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1718

PR #1721, opened 2016-12-06
error check on docker build

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #1720

PR #1650, opened 2016-11-29
Tags in ops metric client

zgalor:
Add tags to metric-client-ops.
Depends on https://github.com/openshift/openshift-tools/pull/1640

PR #1728, opened 2016-12-07
Add docker & openshift version collection script

zgalor:
based on https://github.com/openshift/openshift-tools/pull/1650
Adding a script that collects docker and openshift versions and sends them to hawkular & zabbix


zgalor:
Rebased of prod

zgalor:
rebased after #1792 was merged.
@joelddiaz @twiest please review

scripts/monitoring/cron-send-rpms-versions.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Check OpenShift version
+OC_VERSION=$(oc version | sed -n '1{p;q}' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
+
+# Check Docker version
+DOCKER_VERSION=$(docker --version | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
```
twiest: The problem with this is that it's running inside of our container.

That means that you're checking the version of docker in the container and not on the host.

We bind in the docker socket, so you should be able to use docker-py to talk to the socket and get the version from there.

Does that sound right?

scripts/monitoring/cron-send-rpms-versions.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Check OpenShift version
+OC_VERSION=$(oc version | sed -n '1{p;q}' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
```
twiest: We currently bind in the oc binary, but we may not always do that.

I think this should probably talk to the openshift api and try to get the version from there.

I'm not 100% sure if it's there or not, but that might be best.

Failing that, this might be the only other reasonable way to get this value.

scripts/monitoring/cron-send-rpms-versions.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Check Docker version
+DOCKER_VERSION=$(docker --version | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
+
+HOSTNAME=hostname --fqdn
+
+# Send results to zabbix & hawkular
+ops-metrics-client -s $HOSTNAME -k "docker.version" -o "$DOCKER_VERSION" -t miq_metric="true"
+ops-metrics-client -s $HOSTNAME -k "oc.version" -o "$OC_VERSION" -t miq_metric="true"
```
twiest: What does miq_metric do?

Does it only send it to manageiq?

I think we could send this to both miq and zbx.

twiest (changes requested):

scripts/monitoring/cron-send-rpms-versions.sh
zgalor: The metrics are sent to zabbix and/or hawkular according to config.
the tag miq_metric="true" is used in manageiq to filter from other metrics in hawkular, for the ones we want to collect and place as node attributes.
This tag of course is not sent to zabbix.

In general I believe this tag should be a default for all metrics, in the metric_sender.yaml, but we can add this later if we decide.

twiest: Ok. As long as the metric itself is sent to both, then this is fine.

zgalor:
ok to test

zgalor:
@twiest I made the changes you requested: 
1. Getting docker version using docker api
2. Getting openshift version using REST

I tested in the container in my environment.
Please note the first commit fixes a bug in hawk_client of supporting unicode metrics and cherry-picked from a PR that should be merged first: https://github.com/openshift/openshift-tools/pull/1830

scripts/monitoring/cron-send-docker-oc-versions.py
```diff
@@ -0,0 +1,63 @@
...
+    mts.add_metric({"docker.version": docker_version}, key_tags={'miq_metric': 'true'})
+
+    # Get openshift version
+    if args.username:
+        creds = (args.username, args.password)
+    else:
+        creds = None
+
+    hostname = args.host or os.environ['HOSTNAME']
+    url = "https://{0}:8443/version/openshift".format(hostname)
```
twiest: I'd prefer not to hard code the entire URL here.

I'm fine with `/varsion/openshift`, but I'd prefer it if the rest of the URL came from a config file, or passed on the cli.

Here's how we did it to gather other openshift master data:
https://github.com/openshift/openshift-tools/blob/stg/scripts/monitoring/cron-send-os-master-metrics.py#L64-L68

You don't have to do it that way specifically, but that's an example.

The other thing you may want to consider is using our OpenshiftRestApi class like the example above shows. It's a pretty nice way to interact with the master api.

Here's an example:
https://github.com/openshift/openshift-tools/blob/stg/scripts/monitoring/cron-send-os-master-metrics.py#L160

You could do something like this (untested):
```python
response = ora.get('/version/openshift')
```

twiest (changes requested):
@zgalor This is really close.

I went ahead and merged the other PRs.

scripts/monitoring/cron-send-docker-oc-versions.py
zgalor: I will change to use the Rest API.
Is there a reason why I can't use it for local setup all the time?
`ora = OpenshiftRestApi()`

Will that fail it my node isn't the master node?
If it does fail, and I do need to get the master url from config file, then I see in your example that the config file path is hard coded. Is that ok?

twiest: @zgalor darn, I thought this was supposed to only be run on the masters.

Sorry, I think I led you astray.

If it needs to run on all hosts, then we're going to have to rely on the cli like you had before (again, sorry for leading you astray).

We'll just have to always ensure that the cli is bound into our docker container. We do this today, but I just didn't want to rely on that behavior when we could simply make an rest api call. But if it needs to run on every host, then we can't simply make a rest api call.

zgalor: @twiest Actually I am not quite sure if this needs to run on every host.
@simon3z maybe we should split this?
Collect docker version from each node, but openshift version only from master. It seems like a provider level attribute, no?

twiest: @zgalor I think it actually does make sense to run on each host.

That will allow us to do compliance related tasks like ensure that all hosts in a cluster are on the same version.

So I think your original thoughts on this are correct. I had just assumed masters incorrectly.

zgalor: @twiest Then I am probably confused regarding which openshift-version I should collect.
And if you prefer the CLI approach, do you want me to split the scripts so I can get docker version using the API?

twiest: We want to gather as much information from the APIs as posible. It's just not possible with openshift on every node.

So I'd say gather the docker version from the api like this PR is currently doing, and gather the openshift version from the cli which will have to be bound in.

I'm fine with both being in the same script, but splitting them would probably be ok as well. I don't have a strong opinion either way.

zgalor:
@twiest changed to get oc version using oc binary.
please approve.

twiest (approved):

scripts/openshift-tools-scripts.spec
```diff
@@ -46,6 +46,7 @@ cp -p monitoring/cron-send-docker-timer.py %{buildroot}/usr/bin/cron-send-docker
 cp -p monitoring/cron-send-docker-containers-usage.py %{buildroot}/usr/bin/cron-send-docker-containers-usage
 cp -p monitoring/cron-send-docker-dns-resolution.py %{buildroot}/usr/bin/cron-send-docker-dns-resolution
 cp -p monitoring/cron-send-docker-existing-dns-resolution.py %{buildroot}/usr/bin/cron-send-docker-existing-dns-resolution
+cp -p monitoring/cron-send-docker-oc-versions.py %{buildroot}/usr/bin/cron-send-docker-oc-versions
```
twiest, line 49: This copies the file into the rpm build area, but there isn't an rpm package (or subpackage) where it's being claimed as a file.

Look at cron-send-docker-existing-dns-resolution.py for an example of how to have the RPM spec file claim the file to be laid down:

https://github.com/zgalor/openshift-tools/blob/ed63ee62c9427bf51b94003542719dfcd4d9a457/scripts/openshift-tools-scripts.spec#L193

Basically it has to be added to the `%files` section of a sub package (`monitoring-docker`).

Also, before asking for review, please always test creating a test RPM using tito.

```bash
# cd openshift-tools/script
# time tito build --test --rpm --output "$HOME/.tito-build"
```

This is also documented here (up to and including step 4):
https://github.com/openshift/openshift-tools/blob/stg/docs/build.adoc#build-openshift-tools-scripts--rpms

twiest (changes requested):
Super close :)

scripts/openshift-tools-scripts.spec
zgalor: Should I add it to docker rpm or openshift? (It's related to both)

scripts/openshift-tools-scripts.spec
```diff
@@ -190,6 +191,7 @@ OpenShift Tools Docker Monitoring Scripts
 /usr/bin/cron-send-docker-containers-usage
 /usr/bin/cron-send-docker-dns-resolution
 /usr/bin/cron-send-docker-existing-dns-resolution
+/usr/bin/cron-send-docker-oc-versions
```
zgalor, line 197: @twiest Currently added to docker rpm, and tested locally with tito. If you want it move to openshift I'll move it.

twiest: @zgalor please put it in `monitoring-openshift`

scripts/monitoring/cron-send-docker-oc-versions.py
simon3z: > > Collect docker version from each node, but openshift version only from master. It seems like a provider level attribute, no?
> 
> @zgalor I think it actually does make sense to run on each host.

@zgalor if we're talking about RPMs yes indeed we want to collect the information from each host.

scripts/openshift-tools-scripts.spec
```diff
@@ -296,6 +297,7 @@ OpenShift Tools Openshift Product Scripts
 /usr/bin/cron-send-metrics-checks
 /usr/bin/cron-send-os-router-status
 /usr/bin/cron-send-logging-checks
+/usr/bin/cron-send-docker-oc-versions
```
zgalor, line 308: @twiest done. Script added to openshift rpms.

twiest (approved):
Looks good

PR #1734, opened 2016-12-07
Add a hawkular configuration, client and sender classes to the monitoring lib

twiest:


twiest:
@joelddiaz +1d in STG PR #1732

PR #1742, opened 2016-12-07
update git tools

mwoodson:


ansible/roles/lib_git/build/lib/ssh_agent.py
```diff
@@ -0,0 +1,130 @@
...
+                    actual key data, not the filename of a key)
+        Raises:
+            SshAgentException: when ssh-add does not immediately return (as in the
+                    case of a private key with a passphrase)
+        """
+        #if self.ssh_agent.poll() is None:
+        #    raise SshAgentException("Unable to add ssh key. Did agent die?")
+        named_pipe_path = os.path.join(self.ssh_auth_sock_dir, "keypipe." + str(random.getrandbits(64)))
+        try:
+            os.mkfifo(named_pipe_path, 0600)
```
kwoodson, line 112: is this right?  Does it need to be quoted or anything? Strange formatting in github.

kwoodson (approved):

ansible/roles/lib_git/build/lib/ssh_agent.py
joelsmith: no, python has the same number formatting prefixes as C and many languages since. numbers starting with 0 are octal, just like numbers starting with 0x are hexidecimal.

PR #1746, opened 2016-12-07
Fixed yedit when attribute did not exist.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1745

PR #1747, opened 2016-12-08
Generate metric_sender.yaml from template

zgalor:
Generate the metric_sender.yaml file from template and place it in /etc/openshift_tools/metric_sender.yaml

PR #1748, opened 2016-12-08
Change cloud scripts to use MetricSender instead of ZaggSender

zgalor:
Use MetricSender instead of ZaggSender in monitoring scripts

depends on #1747


zgalor:
@twiest I got some import errors from pylint.
openshift_tools.cloud.aws, 
openshift_tools.cloud.gcp

I don't know if i should disable those.

twiest:
@zgalor yes, it is fine to pylint disable those.

This is that rare case where we need to do it because the libs are not loaded on the box that runs jenkins / pylint.

We're in the process of fixing this, but for now, pylint disabling that check is fine.

zgalor:
@twiest Done. PTAL

zgalor:
@joelddiaz PTAL

twiest (approved):
@zgalor This looks great!

PR #1754, opened 2016-12-08
Change ops-runner and some cron scripts to Use MetricSender

zgalor:
Depends on #1747


zgalor:
@twiest PTAL

zgalor:
@joelddiaz PTAL

twiest (approved):
@zgalor This looks great!

PR #1755, opened 2016-12-08
Add tags to metric sender

twiest:
Add tags field to UniqueMetric. Send tags to hawkular per key.


twiest:
This was +1d in STG PR #1640.

@zgalor fyi, I created and approved this PR to get your changes into our production branch.

PR #1756, opened 2016-12-08
Skip accounts that are in error state.

kwoodson:


ansible/inventory/multi_inventory.py
```diff
@@ -88,20 +91,19 @@ def run_account_providers(accounts, force=False, validate_cache=True, account=No
...
-
-                    updated = True
-                    # The reason this exists here is that we
-                    # are using the futures pattern.  The account object
-                    # has run a subprocess and we are collecting the results.
-                    data = json.loads(out)
-                    acc.write_to_cache(data)
-                    acc.inventory = data
+                        err_msg = 'Account: %s Error_Code: %s StdErr: [%s] Stdout: [%s]\n' % \
+                                   (acc.name, proc.returncode, err, out)
```
joelsmith: err and out probably need to be err.replace('\n', '\\n') and out.replace('\n', '\\n') so that your one-per-line output doesn't get busted by any newlines they might have. or not, if you like that format.

joelsmith (approved):
👍 but you may want to follow my suggestion on log formatting.

PR #1764, opened 2016-12-08
Adding a log file and setting permissions.

kwoodson:


kwoodson:
@joelsmith :+1:  in #1762

PR #1765, opened 2016-12-08
Fixing variable bug.

kwoodson:

PR #1766, opened 2016-12-08
Fixing variable bug.

kwoodson:


kwoodson:
@joelsmith :+1:  in #1765

PR #1769, opened 2016-12-09
Fiz aws_creds_check for zsh, new accounts

joelsmith:


joelsmith:
@BlueShells gave 👍 in stg PR #1768

PR #1772, opened 2016-12-09
Fetch account and user using sts.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1771

PR #1774, opened 2016-12-12
reduce number of enabled yum repos

joelddiaz:
update to current GPG key files for zabbix
remove duplicate yum install packages


joelddiaz:
👍 #1770 in stg

PR #1776, opened 2016-12-12
added vars to start/enable oso-host-monitoring

mwoodson:


mwoodson:
@kwoodson approved #1775

PR #1778, opened 2016-12-12
made the lib_iam_accountid support aws_profiles

mwoodson:


mwoodson:
@kwoodson approved #1777

PR #1780, opened 2016-12-12
Allow AWS creds to be stored anywhere in ~/.private

joelsmith:


joelsmith:
@dak1n1 gave 👍 in stg PR #1779

PR #1782, opened 2016-12-12
Adding inventory check to zabbix.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1781

PR #1784, opened 2016-12-12
Adding multi-inventory checks to ansible-tower hosts.

kwoodson:


kwoodson:
@mwoodson :+1:  in #1783

PR #1788, opened 2016-12-12
Fix sso group sum aggregate item, add to Ops SSO App

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1787

PR #1790, opened 2016-12-12
adding pylint to osohm so we can use it also as a build platform in n…

ivanhorvath:
…ew jenkins


ivanhorvath:
Approved by @joelddiaz in #1789

PR #1791, opened 2016-12-12
Modify PR validation tests to run in a container on openshift

tiwillia:
When a pull request is submitted to the openshift-tools repository on github, a webhook starts a job on a persistent jenkins instance running in an openshift environment. The job is a simple jenkins pipeline using the `Jenkinsfile` in this directory. The pipeline determines the action to take based on the action in the webhook. When a test is necessary, the jenkins pipeline starts a build on the openshift environment titled `openshift-tools-test` and outputs the build logs to the pipeline's logs.

The `openshift-tools-test` build is defined in the `openshift-tools-pr-automation-template.json` template. The build mounts a secret containing the github bot user credentials, then starts a docker build using the Dockerfile `test/Dockerfile`. This build runs from the oso-host-monitoring container and clones the openshift-tools repository. Once the environment is set up, the build runs the `test/run_tests.py` script. This script merges in the changes defined in the github webhook payload and defines several environment variables for consumption by a suite of validators. Each `.py` or .sh` file in `test/validators` is run with these environment varibles.

The script will then submit a comment to the pull request indiciating whether all of the validators passed or if one or more of them failed. The build will exist successfully if all validators pass. When the build completes, so does the jenkins pipeline. The build in jenkins is also marked as a success only if all validators pass.


tiwillia:
If necessary, we can leave the files needed for the current tests instead of deleting them so that tests continue to run while we implement the changes.

jenkins/test/run_tests.py
```diff
@@ -0,0 +1,200 @@
...
+        if skip:
+            continue
+        validator_abs = os.path.join(VALIDATOR_PATH, validator)
+        executer = ""
+        _, ext = os.path.splitext(validator)
+        if ext == ".py":
+            executer = "/usr/bin/python"
+        elif ext == ".sh":
+            executer = "/bin/sh"
+	# If the ext is not recongized, try to just run the file
```
kwoodson, line 113: This makes me nervous and we should chat about it.  I won't block for now but I can see this being a black hole.  I believe this would allow someone to run code in our jenkins instance, is that the case?

jenkins/test/validators/yaml.sh
```diff
@@ -0,0 +1,24 @@
...
+# Overwrite sha's with provided command-line arugments
+if [ -n $1 ]; then
+    base_sha=$1
+fi
+if [ -n $2 ]; then
+    remote_sha=$2
+fi
+
+# Download the latest yaml validation script
+curl https://raw.githubusercontent.com/openshift/openshift-ansible/master/git/yaml_validation.py -o jenkins/test/validators/yaml_validation.py 2>&1
```
kwoodson, line 13: We recently changed this so that yaml_validation.py is in our repository instead of pulling from openshift-ansible.  We need to make sure that we are doing that new style instead of this as we found a bug.

kwoodson (commented):
Please review and discuss.

jenkins/test/run_tests.py
```diff
@@ -173,6 +186,16 @@ def main():
         print "Unable to load JSON data from $GITHUB_WEBHOOK_PAYLOAD"
         sys.exit(1)
     pull_request = payload["pull_request"]
+
+    # Check to ensure the user submitting the changes is on the whitelist
+    user = pull_request["user"]["login"]
+    whitelist = get_user_whitelist)
```
kwoodson: is this correct?  looks like a syntax error.

kwoodson (approved):

PR #1792, opened 2016-12-13
Tags in ops metric client

zgalor:
Add 2 clis ops-metric-client & ops-metric-pcp client (very similar to ops-zagg-client & ops-zagg-pcp-client).
ops-metric-client can set tags to key.
same commits from https://github.com/openshift/openshift-tools/pull/1650 - already merged to stg

cc @twiest


zgalor:
@joelddiaz PTAL

joelddiaz:
@twiest 👍 #1650 in stg

PR #1795, opened 2016-12-13
reverted the iam policy change

mwoodson:


mwoodson:
@kwoodson approved #1794

PR #1796, opened 2016-12-13
Generate metric_sender.yaml from template

zgalor:
A commit from https://github.com/openshift/openshift-tools/pull/1747. Already merged to stg.
cc @joelddiaz


joelddiaz:
👍 #1747 in stg

PR #1804, opened 2016-12-13
Fix the inventory wrapper to report correctly.

kwoodson:


kwoodson:
@joelsmith :+1:  in #1803

PR #1806, opened 2016-12-13
Undo addition of broken parameter to SSO monitoring

joelsmith:
This should fix the config loop on ops-aws masters


joelsmith:
@rhdedgar gave 👍 in stg PR #1805

PR #1808, opened 2016-12-14
change the way we test for pods and pod status; increase test duration by 50%

drewandersonnz:
change the way we test for pods and pod status
increase test duration by 50%


BlueShells:
:+1:

PR #1814, opened 2016-12-15
Monitoring of node participation in ELB

mmahut:
This proposed check will make sure we have the nodes in their respective ELBs and alert us if not.


gburges:
:+1:

PR #1818, opened 2016-12-15
Don't use oc import-image. It breaks some tags.

dak1n1:
This is the bug where we learned we've been importing images incorrectly. https://bugzilla.redhat.com/show_bug.cgi?id=1401794#c20
The oc import command is not needed here, and actually causes issues with importing some tags. Auto-import will handle this as soon as the image streams are defined via 'oc replace'.


mwoodson:
:+1:

PR #1731, opened 2016-12-07
[WIP] oc_image module implements image-import

ivanhorvath:

PR #1820, opened 2016-12-16
oc_image module implements image-import

ivanhorvath:


ivanhorvath:
Approved by @kwoodson in #1731

PR #1821, opened 2016-12-16
disabling CSP until workaround with Zabbix is found

rhdedgar:


joelsmith:
Dunno how many +1s you need, but here's mine. 👍

PR #1826, opened 2016-12-20
local development with 'oc cluster up'

joelddiaz:
new scripts to set up a local OpenShift cluster with CentOS container images for zabbix and host monitoring


docker/local_development/monitoring-secrets/opsmedic.yml
```diff
@@ -0,0 +1,11 @@
...
+g_user_opsmedic:
+  uid: "{{ g_user_opsmedic_uid }}"
+  gid: "{{ g_user_opsmedic_gid }}"
+  groups: "{{ g_user_opsmedic_groups }}"
+  username: opsmedic
+  authkey_opts: command="/usr/bin/remote-healer READ_SSH"
+  pub_key: ssh-rsa AAAAblahblahblah== opsmedic
+  priv_key: |-
+    -----BEGIN RSA PRIVATE KEY-----
+    -----END RSA PRIVATE KEY-----
```
twiest, line 11: Is this just to be used as an example?

Should this be a variable and set the default to this?

docker/local_development/monitoring-secrets/zabbix-server-vars.yml
```diff
@@ -0,0 +1,9 @@
+---
+g_ses_mail_from: libra-ops@redhat.com
+g_ses_smtp_domain: zabbix.ops.openshiftapps.com
+g_ses_smtp_server: email-smtp.us-east-1.amazonaws.com
```
twiest: None of these 3 values should be in openshift-tools.

docker/local_development/monitoring-secrets/zagg-server-vars.yml
```diff
@@ -0,0 +1,20 @@
+---
+g_default_zagg_server_config:
+  targets:
+  - name: local cluster zbx server
+    type: zabbix
+    trapper_server: oso-cent7-zabbix-server
+    trapper_port: 10051
+    api_url: http://oso-centos7-zabbix-web/zabbix/api_jsonrpc.php
+    api_user: Admin
+    api_password: 'zabbix'
```
twiest: Why is this quoted?

docker/local_development/monitoring-secrets/zagg-server-vars.yml
```diff
@@ -0,0 +1,20 @@
...
+    path: /tmp/zagg/data/cluster-zbx
+  templates:
+  - Template Heartbeat
+  - Template Zagg Server
+  hostgroups:
+  - local-dev
+
+g_default_zagg_server_user: admin
+g_default_zagg_server_password: supersecret
+g_default_zagg_server_api_password: 'password'
```
twiest: Why is this quoted?

docker/local_development/monitoring-secrets/zagg-server-vars.yml
```diff
@@ -0,0 +1,20 @@
...
+    api_password: 'zabbix'
+    path: /tmp/zagg/data/cluster-zbx
+  templates:
+  - Template Heartbeat
+  - Template Zagg Server
+  hostgroups:
+  - local-dev
+
+g_default_zagg_server_user: admin
+g_default_zagg_server_password: supersecret
```
twiest, line 19: Can / should this be set as part of the template?

Might not be worth it, just wondering what your thoughts are on it.

docker/local_development/setup-local-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+
+# Make sure 'oc' binary in path
+OC=$(which oc)
+if [ "$?" -ne "0" ]; then
+	echo "Could not find 'oc' binary in path"
+	exit 1
+fi
+
+echo "Flushing iptables rules (fix DNS issues)"
+sudo iptables -F
```
twiest: Won't this basically remove my firewall rules?

This seems bad.

docker/local_development/setup-local-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+${OC} process ${OS_TEMPLATE_NAME} | ${OC} create -f -
+
+echo "Deploying mysql pod"
+${OC} deploy --latest mysql --follow
+
+echo "Deploying zabbix-server pod"
+${OC} deploy --latest oso-cent7-zabbix-server --follow
+
+echo "Deploying zabbix-web pod"
+${OC} deploy --latest oso-centos7-zabbix-web --follow
```
twiest: consider changing this to cent7 to be consistent.

docker/local_development/setup-local-dev-env.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
twiest: Consider changing this to a `#!/bin/bash -e`

docker/local_development/setup-local-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+# sleep another 10 so zabbix-web is really up
+sleep 10
+
+echo "Config zabbix"
+PYTHONPATH=${OPENSHIFT_TOOLS_REPO}:${PYTHONPATH} ansible-playbook ../../ansible/playbooks/adhoc/zabbix_setup/oo-clean-zaio.yml -e zabbix_server=oso-centos7-zabbix-web
+PYTHONPATH=${OPENSHIFT_TOOLS_REPO}:${PYTHONPATH} ansible-playbook ../../ansible/playbooks/adhoc/zabbix_setup/oo-config-zaio.yml -e zabbix_server=oso-centos7-zabbix-web
+
+echo "Deploying zagg pod"
+${OC} deploy --latest oso-centos7-zagg-web --follow
+
```
twiest: We need to be pulling the latest images during this run, otherwise local dev will start using old images.

docker/local_development/setup-local-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+OC=$(which oc)
+if [ "$?" -ne "0" ]; then
+	echo "Could not find 'oc' binary in path"
+	exit 1
+fi
+
+echo "Flushing iptables rules (fix DNS issues)"
+sudo iptables -F
+
+sudo ${OC} cluster up
```
twiest: Make sure this script is re-runnable and that it either fails saying "instance already running" and points them cluster down script.

docker/local_development/setup-local-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+	-v /var/lib/docker:/var/lib/docker:ro \
+	-v /var/run/docker.sock:/var/run/docker.sock \
+	-v ${CONTAINER_SETUP_DIR}:/container_setup:ro \
+	--memory 512m \
+        docker.io/openshifttools/oso-centos7-host-monitoring:latest
+
+# for some reason the initial heartbeat fails
+echo "Sleep for a bit then send heartbeat for host-monitoring"
+sleep 20
+sudo docker exec oso-centos7-host-monitoring /usr/bin/ops-zagg-client --send-heartbeat
```
twiest: This already runs on container start up. Why is it needed here?

docker/local_development/setup-local-dev-env.sh
twiest: Consider renaming this script to "start-local-dev-env.sh" so that it's consistent with "stop-local-dev-env.sh"

docker/local_development/stop-local-dev-env.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+cd $(dirname "$0")
+
+OC=$(which oc)
+if [ "${OC}" -ne "0" ]; then
```
twiest: This should be comparing against `$?` right?

docker/local_development/zabbix_monitoring_centos_local_dev.yaml
```diff
@@ -0,0 +1,374 @@
+apiVersion: v1
```
twiest: Consider using the non-local development zabbix monitoring template using passed in vars.

twiest (changes requested):

docker/local_development/monitoring-secrets/opsmedic.yml
joelddiaz: Yes, just an example. In local development, there is no bastion host with an 'opsmedic' user defined.

docker/local_development/monitoring-secrets/zagg-server-vars.yml
joelddiaz: Just following how we do our secrets in production, but this could probably be done as a template (a secret template?).

tiwillia:
[test] again please!

tiwillia:
@joelddiaz the test failure appears to be legitimate, let me know if this is incorrect:
```
YAML validation failed on docker/local_development/zabbix_monitoring_cent7_local_dev.yaml: while scanning a simple key
  in "docker/local_development/zabbix_monitoring_cent7_local_dev.yaml", line 258, column 1
could not found expected ':'
  in "docker/local_development/zabbix_monitoring_cent7_local_dev.yaml", line 259, column 18
```

For some reason the build output was not included in the jenkins output, I'll have to look into that.

joelddiaz:
[test]

joelddiaz:
@twiest changes/feedback incorporated/addressed

twiest (changes requested):
Looks amazing!

But we need to update 

https://github.com/openshift/openshift-tools/blob/prod/docs/local_development_monitoring.adoc

docs/local_development_monitoring.adoc
```diff
@@ -40,286 +39,67 @@ sudo systemctl start docker
...
-. Set the python path to see our openshift_tools python package, as well as the system's site-packages
-[source]
 ----
-> export PYTHONPATH=$(pwd)/../../../..:$(ls /usr/lib/python2*/site-packages/ -d)
+cd openshift-tools/docker/local_development
+./start-local-dev-env.sh
 ----
 
+This will do the following:
+* Open a firewall port on port 53 so that OpenShift containers can do DNS resolution to other OpenShift containers. The change is ephemeral, so restarting firewalld should undo the change.
```
twiest: These are not seen as bullit points in the rendered doc here:
https://github.com/joelddiaz/openshift-tools/blob/3212168bde24602fd92d9e9f94a1a81872e74682/docs/local_development_monitoring.adoc#running-the-monitoring-containers-locally

docs/local_development_monitoring.adoc
```diff
@@ -40,286 +39,67 @@ sudo systemctl start docker
...
-[CTR][root@ ~]$ ops-zagg-processor
-----
-+
-. Exit the oso-centos7-zagg-web container:
-+
-----
-[CTR][root@ ~]$ exit
-----
+This will:
+* Stop the local OpenShift environemnt with `oc cluster down`
```
twiest: same

twiest (changes requested):

twiest (approved):

PR #1830, opened 2016-12-27
Fix hawk_client to properly send unicode string metrics

zgalor:
use isinstance instead of checking type to include unicode type


zgalor:
@twiest PTAL

twiest (approved):

PR #1836, opened 2017-01-03
added the openshift 3.4 vendored installer

mwoodson:


mwoodson:
@kwoodson approved #1835

PR #1845, opened 2017-01-03
Change scripts to use MetricSender instead of ZaggSender

zgalor:
same commits from https://github.com/openshift/openshift-tools/pull/1620 merged to stg


zgalor:
@twiest @joelddiaz

joelddiaz:
👍 #1620 in stg

PR #1847, opened 2017-01-03
Change cloud scripts to use MetricSender instead of ZaggSender

zgalor:
https://github.com/openshift/openshift-tools/pull/1748 from stg
@twiest @joelddiaz


joelddiaz:
👍 #1748 in stg

PR #1849, opened 2017-01-03
Change docker scripts to use MetricSender instead of ZaggSender

zgalor:
https://github.com/openshift/openshift-tools/pull/1749 from stg
@twiest @joelddiaz


joelddiaz:
👍 #1749 in stg

PR #1850, opened 2017-01-03
Change os scripts to use MetricSender instead of ZaggSender

zgalor:
#1750 from stg
@twiest @joelddiaz


joelddiaz:
👍 #1750 in stg

PR #1851, opened 2017-01-03
Change check sending scripts to use MetricSender instead of ZaggSender

zgalor:
https://github.com/openshift/openshift-tools/pull/1751 from stg
@twiest @joelddiaz


joelddiaz:
👍 #1751 in stg

PR #1852, opened 2017-01-03
Change metrics scripts to use MetricSender instead of ZaggSender

zgalor:
https://github.com/openshift/openshift-tools/pull/1752 from stg
@twiest @joelddiaz


joelddiaz:
👍 #1752 in stg

PR #1854, opened 2017-01-03
Change ops-runner and some cron scripts to Use MetricSender 

zgalor:
https://github.com/openshift/openshift-tools/pull/1754 from stg
@twiest @joelddiaz


joelddiaz:
👍 #1754 in stg

PR #1773, opened 2016-12-11
Change pcp script

zgalor:
Change pcp-ping script to use ops-metric-sender instead of ops-zagg-sender
Depends on #1650 and #1747

PR #1855, opened 2017-01-03
Change send-pcp-ping script to use ops-metric-client

zgalor:
https://github.com/openshift/openshift-tools/pull/1773 from stg
@twiest @joelddiaz


joelddiaz:
👍 #1773 in stg

PR #1857, opened 2017-01-03
Fix hawk_client to properly send unicode string metrics

zgalor:
use isinstance instead of checking type to include unicode type

https://github.com/openshift/openshift-tools/pull/1830 from stg

@twiest @joelddiaz


joelddiaz:
👍 #1830 in stg

PR #1858, opened 2017-01-03
add build count script, cronjob, zitem, ztrigger

stenwt:


stenwt:
ok to test

scripts/monitoring/cron-send-build-counts.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# hacky script to send build counts. Replace with something nicer soon. 
+
+oc get builds --all-namespaces -o yaml --config=/tmp/admin.kubeconfig \
+  | grep phase | sort | uniq -c \
```
joelsmith: this grep is way too general. It'll match on lines of projects with "phase" in their names and will cause all kinds of havoc. Also, you'll want to quote `[:upper:]` and `[:lower:]` because if the cwd has any files with single character filenames, the shell will think that `[:upper:]` is a glob and return the filenames that are `:`, `u`, `p`, etc. Try this experiment to see what I mean:

```
$ echo [:upper:] [:lower:]
[:upper:] [:lower:]
$ touch u p e r
$ echo [:upper:] [:lower:]
e p r u e r
$ echo '[:upper:]' '[:lower:]'
[:upper:] [:lower:]
$ rm -f u p e r
```
Also, when piping data to while loops, the shell automatically puts it in a subshell which is surprising if you try to set vars in the loop (you're not, but this is a point of habit). So I recommend that instead of

```
comannd1 | command 2 | while read ...; do ....; done
```
you do something like

```
while read ...; do ....; done < <(command 1 | command 2)
```
which will not use a subshell. 

If you do that, then it'll make it so that you can keep a running total along the way instead of re-running a get builds at the end.

Finally, if the tool returns any phases you weren't expecting, you'll be trying to send them. So it's probably worth checking to see if it's something you have a corresponding zabbix key for.

How about this instead:

```
declare -A buildstates=( \
 ['cancelled']=1 \
 ['complete']=1 \
 ['new']=1 \
 ['error']=1 \
 ['failed']=1 \
)

while read count state; do
  if [ -z "${buildstates["$state"]}" ]; then
    echo "Unrecognized build state: $state"
    exit 1
  fi
  ops-zagg-client -k "builds_running.${state}" -o "${count}"
  total=$((total+count))
done < <(oc get builds --all-namespaces -o jsonpath='{range .items[*]}{.status.phase}{"\n"}{end}' | sort | uniq -c | tr '[:upper:]' '[:lower:]')

ops-zagg-client -k "builds_running.total" -o "$total" 
```

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -430,6 +430,10 @@ host_monitoring_cron:
   minute: "1"
   job: ops-runner -f -s 30 -n ccexp.openshift.master.certificates /usr/bin/cron-certificate-expirations --cert-list=/etc/origin/master,/etc/origin/node
 
+- name: "Report running builds" 
+  minute: "*/5"
+  job: /usr/bin/cron-send-build-counts
```
joelsmith: Ideally, you should run this under ops-runner and have the script `set -e` somewhere so that if any of the sends break, we'll know about it. Otherwise, if something breaks, we probably won't notice until something else goes wrong.

joelsmith (changes requested):

stenwt:
ok to test

stenwt:
ok to test

stenwt:
ok to test

joelsmith (approved):
👍

PR #1863, opened 2017-01-04
rename openshift-tools-scripts-monitoring-zagg-client

joelddiaz:
...to openshift-tools-scripts-monitoring
fix up dependencies, use new RPM name everywhere, package up unpackaged files, and remove leftover unpackaged files


joelddiaz:
👍 #1864 in stg

PR #1865, opened 2017-01-05
move the node check to the primary master

BlueShells:
now the check are on all 3 masters , it will be better if only alert on one master 

https://trello.com/c/KaE1oEMK/414-change-zabbix-monitoring-for-hosts-not-ready-to-be-one-master-per-cluster


mmahut:
This looks good +1.

mmahut:
ok to test

PR #1871, opened 2017-01-06
catch missing executable file exception

joelddiaz:
add exception handing for when the actual Popen call fails (for example the executable is missing)


joelddiaz:
👍 #1870 in stg

PR #1872, opened 2017-01-06
remove unused kubeconfig check

joelddiaz:


stenwt (approved):
I think we're ok w/o this check

PR #1876, opened 2017-01-06
Monitor build count 2

stenwt:


scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,130 @@
...
+    if args.verbose:
+        logger.setLevel(logging.DEBUG)
+
+    try:
+        builds = count_builds()
+
+    except Exception as e:
+        logger.exception("error during test()")
+        exception = e
+
```
joelsmith: seems pointless to continue if `count_builds()` raised since `builds` won't be set. and if that happens, you'll lose the original exception when trying to iterate across an unset builds in `send_metrics()` below. IMO, if `count_builds()` raises, you should just bail out. I'd remove both of the `try`/`except` blocks in `main()` if it were me.

If you decide to keep one or both of the `try`/`except` blocks in main , keep in mind that re-rasing the exception later, like you're doing, will ruin the original backtrace showing where the error happened. Unless you re-raise from the `except` clause with a plain `raise` (no arguments), the backtrace will be lost. If I remember correctly, you can do gymnastics to save the original backtrace and show that in the re-raise if you want to do it the way you're doing it, but I don't think you gain enough to make it worth it.

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,130 @@
...
+        builds_yaml = runOCcmd_yaml("get builds --all-namespaces")
+        logger.debug(builds_yaml)
+    except Exception:
+        pass # don't want exception if builds not found
+
+    for build_item in builds_yaml["items"]:
+        build_state = build_item["status"]["phase"].lower()
+        logger.debug(build_state)
+        if build_state in valid_build_states:
+            build_counts["%s" % build_state] += 1
```
joelsmith: This reads cleaner as `build_counts[build_state] += 1`

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,130 @@
...
+    logger.info("Data sent to Zagg in %s seconds", str(time.time() - ms_time))
+
+def count_builds():
+    """ count build types """
+
+    build_counts = {}
+    count_build_time = time.time()
+
+    for build_state in valid_build_states:
+        build_counts["%s" % build_state] = 0
```
joelsmith: This reads cleaner as `build_counts[build_state] = 0`

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,130 @@
...
+
+    try:
+        builds_yaml = runOCcmd_yaml("get builds --all-namespaces")
+        logger.debug(builds_yaml)
+    except Exception:
```
joelsmith: your comment says that you're swallowing exceptions because builds might not be found, but what other exceptions could cause it to raise? it might be better to catch only the exception that happens when there are no builds, and verify it, and then `pass`, otherwise `raise` so that other causes of breakage can get noticed/fixed.

joelsmith (changes requested):
Your python version seems to be lacking the total count. Was that intentional? Also, if you take my exception-handling suggestions, you can probably remove several of the pylint disablings you have.

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,116 @@
...
+            build_counts[build_state] += 1
+
+    logger.info(build_counts)
+    logger.info("Count generated in %s seconds", str(time.time() - count_build_time))
+
+    return build_counts
+
+def main():
+    """ setup / test / teardown with exceptions to ensure teardown """
+    exception = None
```
joelsmith: this is a vestige of the old exception handling method. I'd remove it to make the code more readable.

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,116 @@
...
+    if args.verbose:
+        logger.setLevel(logging.DEBUG)
+
+    builds = count_builds()
+
+    ############# send data to zabbix #############
+    send_metrics(builds)
+
+    ############# raise any exceptions discovered #############
+    if exception:
```
joelsmith: another vestige

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,116 @@
...
+
+# We just want to see any exception that happens
+# don't want the script to die under any cicumstances
+# script must try to clean itself up
+# pylint: disable=broad-except
+
+# main() function has a lot of setup and error handling
+# pylint: disable=too-many-statements
+
+# main() function raises a captured exception if there is one
```
joelsmith: this can probably go away

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python
+""" Build status check for v3 """
+
+# We just want to see any exception that happens
+# don't want the script to die under any cicumstances
+# script must try to clean itself up
```
joelsmith: this can probably go away

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,116 @@
...
+# pylint: disable=broad-except
+
+# main() function has a lot of setup and error handling
```
joelsmith: can this go away?

scripts/monitoring/cron-send-build-counts.py
```diff
@@ -0,0 +1,116 @@
...
+    count_build_time = time.time()
+
+    for build_state in valid_build_states:
+        build_counts[build_state] = 0
+
+    get_builds = "get builds --all-namespaces -o jsonpath='{range .items[*]}{.status.phase}{\"\\n\"}{end}'"
+    try:
+        builds_list = runOCcmd(get_builds)
+        logger.debug(builds_list)
+    except Exception:
```
joelsmith: I'm still not a fan of the swallowing of all exceptions here. `ops-aws` currently has no builds, so I suggest you try your code there to see how it handles it. With any luck, it won't raise anything and everything will just work.

joelsmith (changes requested):
a couple of little cleanup items still, plus the total count is still missing

joelsmith (approved):
👍 looks great!

stenwt:
ok to test

PR #1881, opened 2017-01-09
Forgot to copy the script during the build

mmahut:


mmahut:
ok to check

mmahut:
ok to check

mmahut:
ok to test

PR #1906, opened 2017-01-11
Replace calls to ZaggSender with MetricSender in monitoring scripts

zgalor:
Some leftover script changes


zgalor:
@twiest PTAL

twiest (approved):
Looks good

PR #1907, opened 2017-01-11
Use ops-metric-client instead of zagg on image startup

zgalor:
use ops-metric-client to send heartbeat and set cron jobs when image starts


zgalor:
@twiest PTAL

twiest (approved):
Looks great!

PR #1908, opened 2017-01-11
use metric_sender.yaml as default config file for hawk_sender

zgalor:


zgalor:
@twiest PTAL

twiest (approved):
looks good

PR #1909, opened 2017-01-11
use metric_sender.yaml as default config file for zagg_sender

zgalor:
@twiest I think we still want zagg_sender to use the metric_sender config file by default.
Currently it is not called directly by any script and all scripts access it via metric_sender anyway.
I think it's safer to maintain the same file, even if we choose to leave the zagg_client.yaml for alternative configuration.


twiest (changes requested):
Looks good. This is going to require some changes to our config loop, but @joelddiaz said he'd pick that up.

twiest (approved):

PR #1916, opened 2017-01-11
Add a script that sends docker and openshift versions to metric_sender

zgalor:
#1728 based on prod

PR #1917, opened 2017-01-11
Add a script that sends docker and openshift versions to metric_sender

zgalor:
#1916 - stg
@joelddiaz


joelddiaz:
👍 #1916 in stg

PR #1923, opened 2017-01-11
[ci] Use correct SHA for git diff comparisons

tiwillia:
Previously, the incorrect SHA was used to create git diff file lists for
use by the linter and yaml validator. This change enhances the way
changes for a pull request are locally gathered and corrects the SHA
used by the validators

### DIFF
--- a/jenkins/test/run_tests.py
+++ b/jenkins/test/run_tests.py
@@ -16,16 +16,18 @@
 #    PRV_PULL_URL      URL of the pull request
 #
 #  Base info
-#    PRV_BASE_SHA      Current SHA of the target being merged into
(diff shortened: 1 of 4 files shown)


tiwillia:
@kwoodson when you have a chance, please review/merge. This should solve the pylint issues we were seeing.

The problem turned out being that I was comparing against the wrong SHA in the git diff commands. I've tested this fix in my own environment with success.

tiwillia:
Note that I accidentally marked this commit as pending testing while performing the testing for this change. No tests are actually pending, I'm not on the whitelist at the moment so it won't test my PRs.

kwoodson:
+1

PR #1924, opened 2017-01-12
Fix undefined variable error and pylint cleanup

tiwillia:
Remove an unused variable and perform some other various pylint cleanup


tiwillia:
@kwoodson One more error, hopefully this is the last one. Did some cleanup too.

kwoodson (approved):

PR #1931, opened 2017-01-13
Allow tests to be run for pull requests on-demand

tiwillia:
These changes add the following features:

1) The pull request that is commented on is gathered from the pull_request data, rather than being hard coded. This doesn't cause any changes now, but will make it easier to use this code for other repositories in the future.

2) The string [test] can be added to any pull request by a white-listed user to run tests on-demand for that pull request. 

This change requires that the github webhook also send 'issueComment' events.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -2,22 +2,49 @@
 
 node {
     stage "Check PR Action"
+        // Parse json payload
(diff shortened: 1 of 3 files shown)


twiest (approved):
I don't really know groovy, but it all looked reasonable to me. Groovy seems fairly readable.

The python looked good.

PR #1933, opened 2017-01-16
Fix docker oc version script to send rpm versions

zgalor:
Whoops!!!
somehow this got merged without actually sending the values.
I also added a console printout of the values being sent

### DIFF
--- a/scripts/monitoring/cron-send-docker-oc-versions.py
+++ b/scripts/monitoring/cron-send-docker-oc-versions.py
@@ -3,12 +3,14 @@
   Send openshift and docker versions with miq_metric tag to metric_sender
 
   Example:
-  ./cron-send-docker-oc-versions.py -u oc_username -p oc_password
+  ./cron-send-docker-oc-versions.py
 '''
 # Disabling invalid-name because pylint doesn't like the naming conention we have.
 # pylint: disable=invalid-name,import-error
 
-from docker import AutoVersionClient
+import json
+import os
+import sys
 import subprocess
 import argparse
 from openshift_tools.monitoring.metric_sender import MetricSender
@@ -24,6 +26,24 @@ def parse_args():
     return args
 
 
+def add_specific_rpm_version(package_name, rpm_db_path, keys, mts):
+    '''get rpm package version and add to metric sender and keys dictionary
+    '''
+    try:
+        return_value = subprocess.check_output("rpm --dbpath {rpm_db_path} -q {package}".format(rpm_db_path=rpm_db_path,
+                                                                                                package=package_name),
+                                               stderr=subprocess.STDOUT, shell=True)
+
+        if return_value.startswith(package_name):
+            package_version = return_value[len(package_name)+1:len(return_value)-1]
+            key = "{}.version".format(package_name)
+            mts.add_metric({key: package_version}, key_tags={'miq_metric': 'true'})
+            keys[key] = package_version
+            return True, None
+
+    except subprocess.CalledProcessError as err:
+        return False, err
+
 def main():
     '''get docker and openshift versions and send to metric sender
     '''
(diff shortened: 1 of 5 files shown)


zgalor:
ok to test

twiest:
ok to test

tiwillia:
[test] this please.

zgalor:
The reason I changed this PR to a WIP is that I realized I should be collecting docker and openshift rpm versions and not via API, so there is no reason to merge the fix before I do that.

@twiest I also plan to rename is accordingly to cron-send-rpm-versions.
My question is do we still want this packaged in openshift-tools-scripts-monitoring-openshift or in a different sub-package?

twiest:
@zgalor The problem is that if you collect just rpm versions, you'll be collecting the versions from inside the host monitoring container, which may or may not be the same as what's on the host.

zgalor:
@twiest easiest solution would be mounting /var/lib/rpm/ from the host. Question is, will we in anyway be needing the container rpm db we are overriding?

twiest:
I would prefer we mount the host rpm database to a different path like `/host/var/lib/rpm` and then use rpm's `--dbpath DIRECTORY` flag to point it to that path.

You should check for the existence of `/host/var/lib/rpm` and if it doesn't exist, fall back to the path `/var/lib/rpm`.

zgalor:
and in what package should this script be?

twiest:
`openshift-tools-scripts-monitoring-openshift` is the right package IMO.

zgalor:
Ok. I removed the wip. tested both on the node and on the master.
It's ready as far as i'm concerned.
@twiest PTAL

zgalor:
@twiest bump

scripts/monitoring/cron-send-docker-oc-versions.py
```diff
@@ -24,26 +25,60 @@ def parse_args():
...
+
+        if return_value.startswith(package_name):
+            package_version = return_value[len(package_name)+1:len(return_value)-1]
+            key = "{}.version".format(package_name)
+            mts.add_metric({key: package_version}, key_tags={'miq_metric': 'true'})
+            keys[key] = package_version
+            return True
+
+    except subprocess.CalledProcessError:
+        pass
```
twiest: Why are you swallowing this exception?

I think you should either handle it somehow (not sure how), or you should bubble up the error.

This could cause us to think that the versions are getting reported when they aren't. We hate silent failures. :)

scripts/monitoring/cron-send-docker-oc-versions.py
```diff
@@ -24,26 +25,60 @@ def parse_args():
...
+        rpm_db_path = "/var/lib/rpm"
+
+    keys = {}
+
     # Get docker version
-    cli = AutoVersionClient(base_url='unix://var/run/docker.sock', timeout=120)
-    docker_version = cli.version()["Version"]
-    mts.add_metric({"docker.version": docker_version}, key_tags={'miq_metric': 'true'})
+    if not add_specific_rpm_version("docker", rpm_db_path, keys, mts):
+        print "Failed to get docker rpm version."
```
twiest: This isn't enough of a failure IMO. The script itself needs to exit with a non-zero exit codeo (not necessarily right here, but somewhere in the script).

scripts/monitoring/cron-send-docker-oc-versions.py
```diff
@@ -24,26 +25,60 @@ def parse_args():
...
-        oc_version = return_value.split('\n')[0].split(' ')[1]
-        mts.add_metric({"oc.version": oc_version}, key_tags={'miq_metric': 'true'})
+    openshift_package_name = "origin"
+
+    # Get openshift node version (attempt upstream)
+    if not add_specific_rpm_version("{}-node".format(openshift_package_name), rpm_db_path, keys, mts):
+        # Get openshift version (attempt downstream)
+        openshift_package_name = "atomic-openshift"
+        if not add_specific_rpm_version("{}-node".format(openshift_package_name), rpm_db_path, keys, mts):
+            print "Failed to get openshift rpm version."
```
twiest: same

twiest (changes requested):

scripts/monitoring/cron-send-docker-oc-versions.py
zgalor: At first I printed the exception. Problem is we are collecting origin-node and origin-master, and origin-master might not exist and that's not necessarily an error.
Also I don't know if I have atomic-openshift or origin. (the only way to know that is if one fails).
There will be no silent failure because I am always checking return value for this function and if the system call succeeded it returns true. if we catch an exception we return false.
I will however bubble up the error.

scripts/monitoring/cron-send-docker-oc-versions.py
zgalor: I think that if for some reason one of the rpms doesn't exist we still need to send versions for the ones that do exist. 
But I can stop on first failure if you prefer it.

zgalor:
Ok, new logic:
1. Check for docker rpm version
2. Check for <origin/atomic-openshift>-node rpm version
3. Check for <origin/atomic-openshift>-master rpm version 

In any case of error print it.
If 1 or 2 fails - send versions we succeeded collecting but return none-zero code.
If 2 fails - don't attempt to collect 3.
If 3 fails - print notification, but return zero code. We don't know if this runs on master or not. I could query the **osohm_host_type** env variable but I don't want to count on it's existence. Also I believe openshift-master rpm version should be collected on best effort. This will be validated in ManageIQ policy for masters only.

@twiest PTAL

scripts/monitoring/cron-send-docker-oc-versions.py
```diff
@@ -24,26 +25,73 @@ def parse_args():
...
+        if not success:
+            # Print notification but don't count this as failure
+            print "Note: " + err.output
+
+    print "Sending these metrics:"
+    print json.dumps(keys, indent=4)
+    mts.send_metrics()
+    print "\nDone.\n"
+
+    return failures
```
twiest: I'm not seeing where the script exits with the number of failures (or at the very least, non-zero).

Am I missing something?

twiest (changes requested):

scripts/monitoring/cron-send-docker-oc-versions.py
zgalor: Sorry, to much C in my past. 
Better now?

twiest (approved):
Looks great!

PR #1934, opened 2017-01-16
Use base SHA when using git diff to list modified files

tiwillia:
Previously, the SHA of the working merge branch was comapred with the
SHA of the base branch to determine the list of files that were changed
in a particular pull request. This ended up also included any merged
changes to the base branch since the pull request was made.

Using the remote branch SHA will ensure that only the files with changes
that the remote branch includes will be returned from git diff commands.

### DIFF
--- a/jenkins/test/run_tests.py
+++ b/jenkins/test/run_tests.py
@@ -98,6 +98,14 @@ def merge_changes(pull_request):
 
 def run_validators():
     """ Run all test validators """
+    # First, add the validator direcotry to the python path to allow
+    # modules to be loaded by pylint
+    pypath = os.getenv("PYTHONPATH", "")
+    if pypath != "":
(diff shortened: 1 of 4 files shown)


tiwillia:
This is neato, the files modified are under jenkins/test/validators/, so the tests should run them to verify they work.

tiwillia:
I expect this to fail tests.

A new issue was revealed in this testing. If a file under jenkins/test/validators was modified, pylint is unable to properly import the `jenkins/test/validators/common.py` module for these files. The second commit on this PR should resolve that problem by adding the validators directory to the python path, but we won't see the change until this is merged.

@twiest are you okay with merging this? Chicken-and-egg scenario.

I've tested this in the oso-host-monitoring container and the fix is working appropriately.

tiwillia:
@twiest @kwoodson can I get some review & merge on this?

kwoodson (approved):

PR #1945, opened 2017-01-18
debug run time of oc commands, line up in-script timeouts so ops-runn…

stenwt:
…er timeout is less likely to fire

### DIFF
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -242,9 +242,9 @@ host_monitoring_cron:
 
 {% if osohm_host_type == 'master' %}
 {# Openshift Master checks #}
-- name: run create app every 5 minutes
-  minute: "*/5"
-  job: "ops-runner -f -t 180 -s 60 -n csca.openshift.master.app.create /usr/bin/cron-send-create-app -v --basename pull --source openshift/hello-openshift:v1.0.6 &>> /var/log/create_app.log"
+- name: run create app every 10 minutes
+  minute: "*/10"
+  job: "ops-runner -f -t 480 -s 60 -n csca.openshift.master.app.create /usr/bin/cron-send-create-app -v --basename pull --source openshift/hello-openshift:v1.0.6 &>> /var/log/create_app.log"
 
 - name: run Terminating status project every 5 minutes
   minute: "*/5"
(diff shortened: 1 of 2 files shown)


scripts/monitoring/cron-send-create-app.py
```diff
@@ -213,7 +222,7 @@ def test(config):
     noPodCount = 0
     http_code = 0
 
-    for _ in range(testLoopCountMax):
+    for _ in range(self.args.loopcount):
```
drewandersonnz, line 225: optional: after seeing your change I'm thinking this get passed into test() as a default parameter, along with all the other parameters
```
def test(loopCountMax=180, loopDelay=5)
``` 

this way all args (self.args.*) can stay contained to within main()

scripts/monitoring/cron-send-create-app.py
```diff
@@ -65,6 +72,8 @@ def parse_args():
     parser.add_argument('--basename', default="test", help='base name, added to via openshift')
     parser.add_argument('--namespace', default="ops-health-monitoring",
                         help='namespace (be careful of using existing namespaces)')
+    parser.add_argument('--loopcount', default="36",
```
drewandersonnz, line 75: optional: would it be worth introducing all top magic numbers as params here? loop max, loop delay, no pod count max

scripts/monitoring/cron-send-create-app.py
```diff
@@ -65,6 +72,8 @@ def parse_args():
     parser.add_argument('--basename', default="test", help='base name, added to via openshift')
     parser.add_argument('--namespace', default="ops-health-monitoring",
                         help='namespace (be careful of using existing namespaces)')
+    parser.add_argument('--loopcount', default="36", 
+                        help="how many 5 second loops before giving up on app creation")
```
drewandersonnz, line 76: optional: 5 (the number) may not be true in the future, and having it hardcoded can become an error in the future .. I'm thinking this would be more clearly described if we had all the params specified

scripts/monitoring/cron-send-create-app.py
```diff
@@ -40,19 +40,26 @@
 ocutil = OCUtil()
 
 commandDelay = 5 # seconds
```
drewandersonnz, line 42: optional: this is the 3x params I mentioned further down...

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -242,17 +242,17 @@ host_monitoring_cron:
 
 {% if osohm_host_type == 'master' %}
 {# Openshift Master checks #}
-- name: run create app every 5 minutes
-  minute: "*/5"
-  job: "ops-runner -f -t 180 -s 60 -n csca.openshift.master.app.create /usr/bin/cron-send-create-app -v --basename pull --source openshift/hello-openshift:v1.0.6 &>> /var/log/create_app.log"
+- name: run create app every 10 minutes
+  minute: "*/10"
+  job: "ops-runner -f -t 480 -s 60 -n csca.openshift.master.app.create /usr/bin/cron-send-create-app -v --basename pull --source openshift/hello-openshift:v1.0.6 &>> /var/log/create_app.log"
```
drewandersonnz, line 247: optional: I would include the loop count in this one also - not necessary but helps explain the reason for the new timeout.

drewandersonnz:
+1 

optional comments for possible improvements

PR #1952, opened 2017-01-20
Gather changed file list from github and add org whitelist

tiwillia:
This change implements a new way of gathering the files changed in a pull request. Instead of relying on git diff commands, whose output can change when new commits have been pushed to the base branch, we gather the list of changed files directly from Github using the pyGithub library to interface with the Github v3 api.

Additionally, this change implements an org whitelist. Previously the whitelist secret was only a comma-seperated list of github users which would be allowed to run tests. Now, a comma-seperated list of
organizations can be provided in addition to the user whitelist. PyGithub again is used to get the list of members for each organization from the v3 Github api.

Note that this change will require installing [python-PyGithub](https://admin.fedoraproject.org/pkgdb/package/rpms/python-PyGithub/) in the oso-host-monitoring container.


tiwillia:
Tested these changes in a 3.4 environment with no issues.

tiwillia:
@ivanhorvath when you have time, I'm ready to work on getting PyGtihub into the oso-host-monitoring container, which will be necessary for this pull request.

Alternatively, I can `pip install PyGithub` during the testing (which is how I was able to test this change), but I know we are pretty against using pip altogether.

tiwillia:
I expect tests to fail for this PR, due to adding a new environment variable PRV_CHANGED_FILES that is consumed by the validators. The run_tests.py that is currently in stg does not create this environment variable.

kwoodson (approved):

PR #1953, opened 2017-01-23
cron-send-stuck-builds: use the creationTimestamp for new builds

mmahut:
Use _.metadata.creationTimestamp_ instead of _.status.startTimestamp_, which doesn't always exists.


blrm:
Good catch, Marek. Thanks for posting a solution to this so quickly. 👍

PR #1957, opened 2017-01-23
Add OpenShift master audit config role.

dgoodwin:
Configures the settings in auditConfig section of master-config.yaml.

Variables left unset indicates property should be removed if present, we
assume that if calling this role, you intend to manage that section of
config and have passed appropriate vars.

### DIFF
--- a/ansible/roles/openshift_master_audit_config/tasks/main.yml
+++ b/ansible/roles/openshift_master_audit_config/tasks/main.yml
@@ -1,0 +1,25 @@
+---
+- name: "auditConfig.enabled"
+  yedit:
+    src: /etc/origin/master/master-config.yaml
+    key: auditConfig.enabled
+    value: "{{ omac_auditConfig_enabled }}"
+  notify: restart openshift master services
+
+- name: "deploy master audit config options"
+  yedit:
+    src: /etc/origin/master/master-config.yaml
+    key: "{{ item.key }}"
+    value: "{{ item.value }}"
+    state: "{{ 'present' if item.value != '' else 'absent' }}"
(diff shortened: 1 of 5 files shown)


ansible/roles/openshift_master_audit_config/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+  notify: restart openshift master services
+
+- name: "auditConfig.maximumFileSizeMegabytes"
+  yedit:
+    src: /etc/origin/master/master-config.yaml
+    key: auditConfig.maximumFileSizeMegabytes
+    value: "{{ omac_auditConfig_maximumFileSizeMegabytes | default(omit) }}"
+    state: "{{ 'present' if omac_auditConfig_maximumFileSizeMegabytes is defined else 'absent' }}"
+  notify: restart openshift master services
+
```
mwoodson: Can we combine these into a with_items passed into yedit?  Thoughts?

mwoodson (changes requested):

dgoodwin:
@mwoodson ok this is updated now, the empty string default is the only thing I can find that works correctly in terms of removing settings if you don't want them set anymore. Looking at upstream origin docs I don't think setting empty string has any value and is likely invalid.

dgoodwin:
@mwoodson ok this is updated now, the empty string default is the only thing I can find that works correctly in terms of removing settings if you don't want them set anymore. Looking at upstream origin docs I don't think setting empty string has any value and is likely invalid.

mwoodson (approved):

PR #1959, opened 2017-01-23
Fixing zbx_action to filter instead of wildcard search

kwoodson:


kwoodson:
@drewandersonnz :+1:  in #1948

PR #1960, opened 2017-01-24
Add python-hawkular-client.rpm to oso-centos7-host-monitoring Dockerfile

zgalor:
Added the copr repo with the python-hawkular-client rpm and installed the rpm.
Tested locally

### DIFF
--- a/docker/oso-host-monitoring/src/Dockerfile.j2
+++ b/docker/oso-host-monitoring/src/Dockerfile.j2
@@ -78,6 +78,14 @@ RUN yum clean metadata && \
 
 ADD urllib3-connectionpool-patch /root/
 RUN yum install -y patch && yum clean all && cd /usr/lib/python2.7/site-packages/ && patch -p1 < /root/urllib3-connectionpool-patch
+{% if base_os == "centos7" %}
+
+# Add copr repo for python-hawkular-client rpm
+RUN cd /etc/yum.repos.d && curl -O https://copr.fedorainfracloud.org/coprs/g/Hawkular/python-hawkular-client/repo/epel-7/group_Hawkular-python-hawkular-client-epel-7.repo
+{% endif %}
+
+# Install python-hawkular-client
+RUN yum install -y python-hawkular-client && yum clean all
 
 # Ansible startup configuration playbook
 ADD root /root
--- a/docker/oso-host-monitoring/centos7/Dockerfile
+++ b/docker/oso-host-monitoring/centos7/Dockerfile
@@ -76,6 +76,12 @@ RUN yum clean metadata && \
 ADD urllib3-connectionpool-patch /root/
 RUN yum install -y patch && yum clean all && cd /usr/lib/python2.7/site-packages/ && patch -p1 < /root/urllib3-connectionpool-patch
 
+# Add copr repo for python-hawkular-client rpm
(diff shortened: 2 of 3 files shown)


zgalor:
@twiest  PTAL

zgalor:
On second thought, it might be better to have the rpm in https://copr.fedorainfracloud.org/coprs/g/OpenShiftOnlineOps/openshift-tools/packages/

You don't wan't me to change the package externally and break your build.

@twiest @simon3z what do you think?

twiest:
@zgalor The file you modified is a generated file (from a template). So the next time it's generated, your changes will be removed.

Here's more information on why we generate them:

https://github.com/openshift/openshift-tools/blob/prod/docker/README.adoc#directory-structure

We want to change this process because it's tedious and non-intuitive, but, for now it is our process.

You need to modify this Dockerfile and make sure it's in the `{% if base_os == "centos7" %}` section so that it only applies to centos:

https://github.com/openshift/openshift-tools/blob/prod/docker/oso-host-monitoring/src/Dockerfile.j2





As far as the "break the build", that's a good point, however, that also relegates us to maintaining that RPM and making sure it's up to date with the one we use in RHEL.

Neither is a good situation, but I think having you maintain a COPR repo with that client (at least for now) is probably the better of the two.

We can always change this if it becomes a problem.

zgalor:
Whoops, my bad.
Fixed and generated from template.
@twiest PTAL

twiest (approved):
@zgalor Looks great!

Is this ready to merge?

zgalor:
@twiest yes please

docker/oso-host-monitoring/src/Dockerfile.j2
```diff
@@ -78,6 +78,12 @@ RUN yum clean metadata && \
 
 ADD urllib3-connectionpool-patch /root/
 RUN yum install -y patch && yum clean all && cd /usr/lib/python2.7/site-packages/ && patch -p1 < /root/urllib3-connectionpool-patch
+{% if base_os == "centos7" %}
+
+# Add copr repo for python-hawkular-client rpm and install it
+RUN cd /etc/yum.repos.d && curl -O https://copr.fedorainfracloud.org/coprs/zohargalor/python-hawkular-client/repo/epel-7/zohargalor-python-hawkular-client-epel-7.repo
+RUN yum install -y python-hawkular-client
```
twiest: In docker, you ALWAYS need to clean up after yum so that it's cached files aren't committed to the image layer.

So this line should be something like this (untested):
```
RUN yum install -y python-hawkular-client && yum clean all
```


Also, do we want to do this install only on centos?

I believe the package is available for rhel as well. 

The copr repo addition, however, should only be in the centos section.

twiest (changes requested):
Sorry, didn't think about this until after I approved the PR.

zgalor:
Done. 
@twiest PTAL

zgalor:
@twiest bump

zgalor:
Actually, let's wait till I can consume the Hawkular rpm from Hawkular repo in copr

twiest:
@zgalor ok, sounds good. Bump me again when that's ready.

zgalor:
@twiest let's merge this, and once we have a hawkular repo on copr I'll submit a new PR.

zgalor:
@twiest Now it's with the official Hawkular group repo.
Tested on my vm.

twiest (approved):
Looks great!

PR #1967, opened 2017-01-24
take into account when value returned is not 'Gi'

joelddiaz:
don't assume 'type' is defined as selector for all pods


joelddiaz:
👍 #1964 in stg

PR #1968, opened 2017-01-24
was changed inappropriately

drewandersonnz:
this should only happen if there is no exception, if exception the variables aren't even available :-(


blrm:
👍 Looks good! Thanks

PR #1971, opened 2017-01-25
Fix SAML authorization of users in YAML auth module

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #1970

PR #1974, opened 2017-01-25
get logging and metrics working on 3.4

mwoodson:


mwoodson:
@wshearn approved #1972

PR #1976, opened 2017-01-25
Fix bad push to upstream

joelsmith:
Bad push wiped out changes from PR #1973 and #1974


joelsmith:
already approved in #1974 (which my bad push to upstream wiped out)

PR #1969, opened 2017-01-25
If get fails, we exceeded API rate of amazon

mmahut:
From time to time, we exceed API rate of amazon and we need to wait for the next run. This is fine, we just don't want an exception here.

PR #1980, opened 2017-01-26
If get fails, we exceeded API rate of amazon

mmahut:
#1969 to master


mmahut:
```
OSError: [Errno 2] No such file or directory: 'jenkins/test/validators/'
```

Broken tests, should be fine.

PR #1983, opened 2017-01-26
gcloud RPM workaround

joelddiaz:
Iinstalling the gcloud RPM tries to pull in the releases-optional repo metadata which is over 1GB (even though gcloud doesn't depend on anything in that repo). Disable that repo when installing the gcloud RPM.


joelddiaz:
👍 #1982 in stg

PR #1987, opened 2017-01-26
made it possible to pass in default dm basesize for docker storage setup

mwoodson:


mwoodson:
@kwoodson approved #1986

PR #1995, opened 2017-01-27
remove non-existent RPM

joelddiaz:
...and redundant 'yum update' command (already do yum update in ops-base)


joelddiaz:
👍 #1994 in stg

PR #1996, opened 2017-01-27
Vendor openshift ansible 3.5.2

dgoodwin:
Despite the version mismatch, this version contains fixes we need for ops purposes to deploy 3.4 so for now this is correct. 

@mwoodson did I get this correct:

- download packages from brew for this build to ~/openshift-ansible-rpms/
- cd ~/git/openshift-tools/openshift/installer
- rm atomic-openshift-3.4 (otherwise I get permissions errors)
- ./vendor-rpms.yml -e openshift_version=3.4 -e installer_version=3.5.2 -e rpm_directory=/home/dgoodwin/openshift-ansible-rpms


dgoodwin:
PR is filed against stage as IIRC that is where you guys do this?

mwoodson:
lets discuss this.  We might want to go with a 3.5 version

dgoodwin:
Updated symlink to keep the 3.4, add a 3.5, but we will point to the 3.5 version specifically for 3.4 upgrades.

mwoodson:
This failed the bot because we are vendoring openshift-ansible.  This doesn't pass our standards and is allowed.

:+1:

PR #2000, opened 2017-01-30
fixing converter function

ivanhorvath:


ivanhorvath:
Approved by @kwoodson in #1998

PR #2004, opened 2017-01-30
Vendor in openshift-ansible 3.5.2.

dgoodwin:
We will use this for 3.4 upgrades with a hardcoded reference.


dgoodwin:
@mwoodson approved #1996

PR #1997, opened 2017-01-27
Push the CI updates to prod

tiwillia:
The CI updates have been pushed to and used in the stg branch. This PR pushes the changes to prod to allow for proper testing of PRs going into the prod branch.

PR #2005, opened 2017-01-30
Fix CI parent branch check

tiwillia:
Previously, the list of commits found by the parent check script was
incorrect, causing failures on pull requests submitted to the prod
branch. This change reaches out to github via the v3 api with an
authenticated request to gather the definitive list of commits for a
pull request and compare them against the commits in the stg branch
before merging to prod.

### DIFF
--- a/jenkins/test/github_helpers.py
+++ b/jenkins/test/github_helpers.py
@@ -72,3 +72,14 @@ def get_changed_files(repo, pr_num):
         filename = prfile.filename
         filename_list.append(filename)
     return filename_list
+
+# Takes a repo in the form 'org/repo-name' and a pr number string or int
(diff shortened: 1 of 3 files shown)


tiwillia:
@twiest @kwoodson can you please give this a look when you have a chance?

This, plus https://github.com/openshift/openshift-tools/pull/1997 should resolve the issues with testing on the prod branch.

tiwillia:
Tests are failing here because the change made to run_tests.py to set the new environment variable containing the list of commits that the parent check consumes is not present in stg yet.

All other tests are passing, this is ready to merge upon approval.

twiest (approved):
Looks great!

PR #2007, opened 2017-01-31
Adding 'unknown' build state to the count script

blrm:
@stenwt 👍 ed  in #2006


blrm:
I tested this script on a cluster with unknown builds and it worked as expected.

PR #2012, opened 2017-02-01
router DC changes for setting x_forwarded_for header

joelddiaz:
With OpenShift 3.3.1.11 and 3.4.1, the OpenShift router image accepts the environment var ROUTER_USE_PROXY_PROTOCOL to have the router pod use the PROXY protocol.

This allows propogation of the X-FORWARDED-FOR HTTP header showing the client IP address instead of the ELB's IP.

This will be the default setting going forward. This requires the AWS ELB to be configured to speak the PROXY protocol as well (https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html).

### DIFF
--- a/ansible/roles/openshift_secure_router/tasks/main.yml
+++ b/ansible/roles/openshift_secure_router/tasks/main.yml
@@ -8,6 +8,13 @@
   - "{{ ossr_default_router_keys }}"
   - "{{ ossr_default_router_cacert }}"
 
+- name: Include cloud-specific router edits
+  include_vars: "{{ ossr_cloud }}.yml"
+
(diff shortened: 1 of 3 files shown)


joelddiaz:
@mwoodson here's the changes to get the OpenShift routers to pass along the x-forwarded-for header. Another (non-merged) change in the AWS ELB port settings is required for all this stuff to work together.

kwoodson (approved):

dgoodwin (approved):
I have tested this in ded-int-aws, with Joel's help to verify it appears this is working. I don't know if my approval is enough but everything here looks good to me.

PR #2013, opened 2017-02-02
Build, install, and run unit tests on test rpms during CI

tiwillia:
This change adds two steps to the PR validation process. After validation scripts (pylint, yaml validation, etc) are successful, test versions of the openshift-tools rpms are built and installed. If successful, unit tests will then be run against the installed rpms.

This change should validate that rpms can be built and installed with the changes included in a pull request.

### DIFF
--- a/jenkins/test/run_tests.py
+++ b/jenkins/test/run_tests.py
@@ -36,11 +36,6 @@
 # TODO:
 # - Handle failures better. Just exiting is not a good option, as it will likely leave the PR
 #    commit status in pending forever.
-# - There are two packages that are not available in the centos version of the oso-host-monitoring image:
-#     oso-simplesamlphp https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=518089
-#     python-ruamel-yaml https://brewweb.engineering.redhat.com/brew/packageinfo?packageID=61628
-#   This will cause centos builds to fail. The python-ruamel-yaml rpm is needed for running the unit
-#    tests. The oso-simplesamlphp rpm is needed to build and install the test rpms
 
 import os
 import json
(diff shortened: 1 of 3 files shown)


tiwillia:
Note that tests run on this PR will use the run_tests.py currently in stg, which will not run through building rpms. I'm working on testing this separately to ensure the changes will work before merging.

tiwillia:
After speaking with @kwoodson about these changes, I need to move the funciton that runs tests to not run anything for now. The yedit tests and env setup script is currently runs are being moved to openshift-ansible. The function for now should call an empty script, so that we can modify the script with the unit tests down the road without changes to the CI.

jenkins/test/run_tests.py
```diff
@@ -116,11 +138,13 @@ def run_validators():
     """ Run all test validators """
     # First, add the validator direcotry to the python path to allow
     # modules to be loaded by pylint
+    # We also add the jenkins/test directory so that github_helpers can be properly loaded.
     pypath = os.getenv("PYTHONPATH", "")
+    tools_test_path = OPENSHIFT_TOOLS_PATH + "jenkins/test/"
     if pypath != "":
-        os.environ["PYTHONPATH"] = VALIDATOR_PATH + os.pathsep + pypath
+        os.environ["PYTHONPATH"] = VALIDATOR_PATH + os.pathsep + tools_test_path + os.pathsep + pypath
```
tiwillia, line 145: The test failures seen in this PR are fixed by this change to run_tests.py.

They will not take affect on the PR to stg though.

tiwillia:
@kwoodson @twiest This PR is ready for review. I've tested on my own instance with success.

The test failure we are seeing is expected for this PR only. The fixes in this PR should address the failure.

All together, building and installing the rpms adds about 30 seconds to the test process. On my own instance (which is about the same speed as ops-aws now that the iops issue is resolved), we are seeing tests take a total of ~4:30 compared to the current ~4:00.

twiest (approved):
Looks great!

PR #2015, opened 2017-02-03
upped the timeouts for logging/metrics

mwoodson:


mwoodson:
@dgoodwin approved #2014

PR #2018, opened 2017-02-03
Add tito and build deps to oso-host-monitoring container

tiwillia:
This commit modifies the Dockerifle of the oso-host-monitoring container
to install tito and the build dependencies necesary for building the
openshift tools rpms

### DIFF
--- a/docker/oso-host-monitoring/centos7/Dockerfile
+++ b/docker/oso-host-monitoring/centos7/Dockerfile
@@ -68,6 +68,11 @@ RUN yum clean metadata && \
         openvswitch \
         python-psutil \
         pylint \
+        tito \
+        python-devel \
+        libyaml-devel \
+        oso-simplesamlphp \
+        python-ruamel-yaml \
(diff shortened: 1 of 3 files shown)


tiwillia:
@kwoodson @twiest PTAL, these are the dependencies required to build & install the rpms into the oso-host-monitoring container. There are two dependencies here that aren't available in the current repositories for the centos version of the oso-host-monitoring container:

  oso-simplesamlphp and python-ruamel-yaml 

Both of these are tagged into li and are available in the rhel7 version. How can we ensure there will be available for the centos version?

twiest:
I talked with @tiwillia about our options and I've added the missing RPMs to our COPR repo.

We'll have to manually sync these from time to time, so not the cleanest of solutions, but it's best of a set of bad options.

twiest (approved):

tiwillia:
Tested the RHEL and CENTOS builds on tower and they are building fine, merging.

PR #2024, opened 2017-02-06
Add tito and build deps to oso-host-monitoring container

tiwillia:
This commit modifies the Dockerifle of the oso-host-monitoring container
to install tito and the build dependencies necesary for building the
openshift tools rpms


tiwillia:
This has already been approved, tested, and merged into stg.

PR #2025, opened 2017-02-07
added the check for the internal pod

zhiwliu:


zhiwliu:
https://github.com/openshift/openshift-tools/pull/1940

PR #2029, opened 2017-02-07
vendored latest 3.5.

kwoodson:


kwoodson:
@mwoodson :+1:  in #2028

PR #2030, opened 2017-02-08
the new version just change from Gi to G

BlueShells:
the new version of the online-volume-provisioner-3-5m0ot , change from Gi to G, like 1000Gi to 1000G , so this make the script can't work as hope


zhiwliu:
:+1:

PR #2035, opened 2017-02-08
defaultMode field on secret causing oadm_router to detect changes

joelddiaz:
when there were none

ignore defaultMode when trying to determine if there has been a configuration change

### DIFF
--- a/ansible/roles/lib_openshift_3.2/library/oc_obj.py
+++ b/ansible/roles/lib_openshift_3.2/library/oc_obj.py
@@ -650,7 +650,7 @@ class Yedit(object):
         tmp_filename = self.filename + '.yedit'
         try:
             with open(tmp_filename, 'w') as yfd:
-                # pylint: disable=no-member
+                # pylint: disable=no-member,maybe-no-member
                 if hasattr(self.yaml_dict, 'fa'):
                     self.yaml_dict.fa.set_block_style()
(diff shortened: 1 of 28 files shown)


ansible/roles/lib_openshift_3.2/library/oc_obj.py
```diff
@@ -934,7 +934,7 @@ def create(self, files=None, content=None):
         '''
            Create a config
 
-           NOTE: This creates the first file OR The first conent.
+           NOTE: This creates the first file OR the first conent.
```
kwoodson: This change is only in the library file.  Needs to move to source?

kwoodson (approved):

tiwillia:
[test]

tiwillia:
If all the stars are properly aligned, this should pass [test] now, or at least stop failing them because of the ruamel.yaml import.

tiwillia:
@joelddiaz the test failures appear to be actual failures now, although they look to have existed prior to your change. Sorry about the issues.

ansible/roles/lib_openshift_3.2/library/oc_obj.py
detiber, line 937: s/conent/content

kwoodson (approved):

PR #2038, opened 2017-02-08
Add optional redirect URL feature to SSO app's authorizeyaml module

joelsmith:
Also update example authdata.yaml file to show a role override and
a redirect url (relaystate_url) which translates into a RelayState
parameter for simpleSAMLphp.


rhdedgar:
👍

PR #2041, opened 2017-02-09
defaultMode field on secret causing oadm_router to detect changes

joelddiaz:
...when there were none

also clean up several new pylint issues


joelddiaz:
👍 #2035 in stg

PR #2042, opened 2017-02-09
adding stale credential entry removal option

rhdedgar:


scripts/iam-tools/aws_creds_check.sh
```diff
@@ -107,6 +107,8 @@ function _check_creds()
                     fi
                 elif [ "$line" != "default" ]; then
                     echo "AWS Credentials exist in $credfile for unrecognized (possibly stale?) account '$line'"
+		    echo "You can remove them with the following command:"
+		    echo "aws_api_key_manager -c -p '$line'"
```
joelsmith: These need to have spaces, not tabs, for their indenting so that they match the rest of the file.

scripts/iam-tools/aws_api_key_manager.py
```diff
@@ -265,6 +277,43 @@ def get_token(aws_account):
...
+            for aws_account in args.profile:
+                try:
+                    config.remove_section(aws_account)
+                    print('Successfully removed entry for %s' % aws_account)
+
+                except ConfigParser.NoSectionError:
+                    print('Account section could not be found, skipping')
+                    return True
+
+                else:
```
joelsmith: I advise against using an `else` with a `try`... it's an uncommon pattern and doesn't really accomplish anything.

```
try:
   do_something
   do_something_else
except ExceptionThrownBy_do_something:
   ...
```

is the same as (but more common) than this:

```
try:
   do_something
except ExceptionThrownBy_do_something:
   ...
else:
   do_something_else
```

The only time they would behave differently is if `do_something_else` threw `ExceptionThrownBy_do_something`, but usually the exception handler can be made to handle both throwers and the `else` can be avoided.

Also, I recommend only writing the file once. For example:

```
need_write = False
for aws_account in args.profile:
    try:
        config.remove_section(aws_account)
        print('Successfully removed entry for %s' % aws_account)
        need_write = True
    except ConfigParser.NoSectionError:
        print('Account section could not be found, skipping')
if need_write:
    with open(path, 'w') as configfile:
        config.write(configfile)    
```

scripts/iam-tools/aws_api_key_manager.py
```diff
@@ -265,6 +277,43 @@ def get_token(aws_account):
...
+            config.read(path)
+
+            for aws_account in args.profile:
+                try:
+                    config.remove_section(aws_account)
+                    print('Successfully removed entry for %s' % aws_account)
+
+                except ConfigParser.NoSectionError:
+                    print('Account section could not be found, skipping')
+                    return True
```
joelsmith: With this early return, any not found profile will get skipped, but so would any other profiles specified after it. Usually, `continue` instead of `return` is the right answer when you want to skip to the next. Of course, if the loop is about to come back around anyways, you can just leave the `continue` off.

scripts/iam-tools/aws_api_key_manager.py
```diff
@@ -422,6 +471,9 @@ def main(self):
         elif args.all and args.user:
             self.run_all(args, ops_accounts)
 
+        elif args.clean and args.profile:
```
joelsmith: This seems like it would need to be the first `if`, because `args.user` is almost always going to be present since you populate it for the user if they don't specify one, so the `if` at line 468 will always be true and this code will never be reached.

joelsmith (dismissed):
A few little things to fix, but otherwise looks great.

rhdedgar:
ok to test

joelsmith (approved):
👍

PR #2047, opened 2017-02-09
updated vendored 3.5 version

mwoodson:


mwoodson:
@dgoodwin approved #2046

PR #2052, opened 2017-02-09
change_iam_password: Capture a few of the common expected exceptions

joelsmith:
Give friendly errors with possible solutions for common exceptions

PR #2053, opened 2017-02-09
change_iam_password: Capture a few of the common expected exceptions

joelsmith:
Give friendly errors with possible solutions for common exceptions


joelsmith:
@rhdedgar gave 👍 in stg PR #2052

PR #2054, opened 2017-02-09
Support unicode characters in PR body and title

tiwillia:
This change ensures that the CI can properly handle unicode characters in both the pull request body and title.


tiwillia:
@jdiaz could you review this small change?

This should fix the issue you hit the other day where an ellipsis unicode character caused the CI testing to fail.

ivanhorvath (approved):

PR #2055, opened 2017-02-10
Support unicode characters in PR body and title

tiwillia:
This change ensures that the CI can properly handle unicode characters in both the pull request body and title.


tiwillia:
This was already approved when merged to stg

PR #2059, opened 2017-02-10
changed volume provisioner to use openshift-scripts-dedicated

mwoodson:


mwoodson:
@yocum137 approved #2058

PR #2068, opened 2017-02-13
added us-east-2 as a region to copy ami's to

mwoodson:


mwoodson:
@kwoodson approved #2060

PR #2081, opened 2017-02-14
Add job that reports docker oc rpm versions daily

zgalor:


zgalor:
@twiest @joelddiaz PTAL

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -149,6 +149,11 @@ host_monitoring_cron:
   minute: "*/5"
   job: ops-runner -f -s 15 -n csnm.network.int /usr/bin/cron-send-network-metrics
 
+- name: report rpm versions daily
+  hour: "1"
+  minute: "0"
+  job: ops-runner -f -s 120 -n csdov.rpm.versions /usr/bin/cron-send-docker-oc-versions
```
twiest, line 155: What does this do when run against zabbix?

I think we need zabbix items to catch these. Do you agree?

twiest (changes requested):

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
zgalor: What would happen on zabbix if the item isn't defined? Will it just be ignored?

I'm not sure we need it on zabbix. It's not a metric. We are simply using hawkular as a medium to transfer data that we can check against expected values in manageIQ policies.

twiest: Yeah, we have this for some things (like kernel version).

Here's an example:
https://github.com/openshift/openshift-tools/blob/prod/ansible/roles/os_zabbix/vars/template_os_linux.yml#L71-L74


Check some of the PCP items for more details.

zgalor: I can add that task in the zabbix ansible template, but I can't test it so I think it's best if someone from your end adds it, if you think you need that info on zabbix.

twiest: @zgalor You can't use the zaio to test this out?

My concern is that we're going to start having some checks in one and not the other. I think we need to keep these in sync.

zgalor:
@twiest I added zabbix keys for new metrics

ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -2,6 +2,18 @@
 g_template_openshift_master:
   name: Template Openshift Master
   zitems:
+  - key: origin-master.version
+    description: Show origin-master rpm version
+    value_type: string
+    applications:
+    - Openshift Cluster
+
+  - key: atomic-openshift-master.version
```
twiest: Hmm, this naming scheme of the key doesn't really fit in with the rest of the master keys.

I didn't notice it in the other PR (sorry).

What are your thoughts on changing the key to be something like `openshift.master.version.atomic-openshift` ?

And the one above would be something like `openshift.master.version.origin-master`.

Obviously we'd want to change the node keys as well. The `docker.version` is prefect as is, though.

The reason that keeping the scheme the same is that we display metrics in zabbix using the "latest data" and its search is quite limited.

So if this is a different pattern, it won't show up if we search for "openshift.master" metrics.

twiest (changes requested):

ansible/roles/os_zabbix/vars/template_openshift_master.yml
zgalor: I am sending rpm versions. So the pattern is the `<name of the rpm>.version`.
It also needs to be verified against expected values you push to manageIQ using ansible playbooks.
If hawkular that allows searching by tags, and manageIQ are the future, is it really worth complicating the keys to match a pattern in zabbix, we are planning to replace anyway?

zgalor:
@twiest renamed according to what we discussed and added descriptor_name tag.
Please review

twiest:
@zgalor please fix the branch conflicts and I'll re-review.

zgalor:
@twiest rebased

twiest (approved):

PR #2082, opened 2017-02-14
local development with 'oc cluster up'

joelddiaz:
The new script will set up a local OpenShift cluster, and install the CentOS versions of our zabbix/host monitoring.

start-local-dev-env.sh to start OpenShift, create SSL certificates, load templates, start containers, etc.
stop-local-dev-env.sh to stop OpenShift


joelddiaz:
👍 #1826 in stg

PR #2084, opened 2017-02-15
added cron-send-internal-pods-check.py

zhiwliu:


zhiwliu:
https://github.com/openshift/openshift-tools/pull/2083

Issue #2080, opened 2017-02-13
CI fails when pull request contains double-quotes


tiwillia:
When a pull request contains double-quotes in the title or in the description, the CI fails to run. No update form the CI is seen at all on the pull request, but a job is created on the jenkins server.

This appears to be due to the way the openshift-jenkins-plugin escapes characters in the environment variables passed into a job. The double quotes end up being escaped twice, creating unparse-able json:
```
..., "title": "This is a test PR with \\"double quotes\\"", ...
```

When it should end up like this:
```
..., "title": "This is a test PR with \"double quotes\"", ...
```

This issue can be worked around temporarily by removing the double-quotes from the PR title or description.

tiwillia:
After some testing, it looks like this is due to how openshift escapes env variables in builds. Running the below buildConfig will result in an improperly-escaped json string in the PAYLOAD env variable:
```
{
  "kind": "BuildConfig",
  "apiVersion": "v1",
  "metadata": {
    "name": "json-test",
  },
  "spec": {
    "source": {
        "type": "Dockerfile",
        "dockerfile": "FROM centos:7\nRUN env"
    },
    "strategy": {
      "type": "Docker",
        "dockerStrategy": {
        "noCache": true,
        "env": [
            {
              "name": "PAYLOAD",
              "value": ""
            }
          ]
       }
    }
  }
}

```
```
oc start-build json-test -e PAYLOAD='{"test":"This is a test with \"double-quotes\" in the value"}'
oc logs -f bc/json-test
...
PAYLOAD={"test":"This is a test with \\"double quotes\\" in a json value"}
...
```

The problem has nothing to do with the jenkins plugin

tiwillia:
Solved by https://github.com/openshift/openshift-tools/pull/2090

PR #2087, opened 2017-02-15
Implement add_dynamic_metric for HawkSender

zgalor:
Implementation will create tagging rules (similar to config file "metadata_rules").

### Example:
**cron-send-network-metrics**, calling:
metric_sender.**add_dynamic_metric**("disc.network", "#OSO_NET_INTERFACE", ["tun0", "vethcf88663d", "eth0"])

will create the following tagging rules:
[ {"regex": "disc.network.*\\[tun0\\]", "tags": {"OSO_NET_INTERFACE": "tun0"}},
  {"regex": "disc.network.\*\\[vethcf88663d\\]", "tags": {"OSO_NET_INTERFACE": "vethcf88663d"}}, 
  {"regex": "disc.network.\*\\[eth0\\]", "tags": {"OSO_NET_INTERFACE": "eth0"}}]

Then, 6 calls to **add_metric** with the following pairs:
disc.network.in.bytes[tun0]:  66085550
disc.network.in.bytes[vethcf88663d]:  30809517
disc.network.in.bytes[eth0]:  2174481627
disc.network.out.bytes[tun0]:  68700631
disc.network.out.bytes[vethcf88663d]:  41424582
disc.network.out.bytes[eth0]:  8585515645

Will produce the following metric tags:
key:  node/example.host.com/disc.network.in.bytes[tun0]
tags: {'OSO_NET_INTERFACE': 'tun0'}

key:  node/example.host.com/disc.network.in.bytes[vethcf88663d]
tags: {'OSO_NET_INTERFACE': 'vethcf88663d'}

key:  node/example.host.com/disc.network.in.bytes[eth0]
tags: {'OSO_NET_INTERFACE': 'eth0' }

key:  node/example.host.com/disc.network.out.bytes[tun0]
tags: {'OSO_NET_INTERFACE': 'tun0'}

key:  node/example.host.com/disc.network.out.bytes[vethcf88663d]
tags: {'OSO_NET_INTERFACE': 'vethcf88663d'}

key:  node/example.host.com/disc.network.out.bytes[eth0]
tags: {'OSO_NET_INTERFACE': 'eth0' }

### DIFF
--- a/openshift_tools/monitoring/hawk_sender.py
+++ b/openshift_tools/monitoring/hawk_sender.py
@@ -11,13 +11,11 @@ Examples:
     from openshift_tools.monitoring.hawk_sender import HawkSender
     HOSTNAME = 'host.example.com'
 
-    ZAGGCONN = HawkConnection(url='https://172.17.0.151', user='admin', password='pass')
-    ZAGGHEARTBEAT = HawkHeartbeat(templates=['template1', 'template2'], hostgroups=['hostgroup1', 'hostgroup2'])
+    HAWKCONN = HawkConnection(url='https://172.17.0.151', user='admin', password='pass')
 
-    zs = HawkSender(host=HOSTNAME, hawk_connection=ZAGGCONN)
-    zs.add_heartbeat(ZAGGHEARTBEAT)
-    zs.add_zabbix_keys({ 'test.key' : '1' })
-    zs.send_metrics()
+    hs = HawkSender(host=HOSTNAME, hawk_connection=HAWKCONN)
+    hs.add_metric({ 'test.key' : '1' })
+    hs.send_metrics()
 """
 import re
 from openshift_tools.monitoring.metricmanager import UniqueMetric
@@ -45,6 +43,7 @@ class HawkSender(GenericMetricSender):
         self.unique_metrics = []
         self.verbose = verbose
         self.debug = debug
+        self.dynamic_tag_rules = []
 
         if not host:
             host = self.get_default_host()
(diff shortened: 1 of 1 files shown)


zgalor:
@twiest PTAL. tests are not running for some reason...

twiest:
[test]

zgalor:
This is a rather ugly solution.
Refactoring all scripts that use add_dynamic_metric would be cleaner.
Question is does dynamic metric scripts contain crucial tagging data we need to filter by, or set triggers by in the near future.

@twiest @simon3z your thoughts on that?

twiest:
The more I see this, the more I think we're going to have to.

The regex thing you're relying on here is simply a pattern we use, but it's not enforced. Which means that somewhere in our code base, I'm sure we have dynamic items that don't use that pattern, which means the regex wouldn't work for them.

I don't have an example right now, though.

tiwillia:
It looks like the tests didn't run due to issue #2080 

PR #2090 should resolve the test failures once merged into stg.

tiwillia:
Please re [test]

tiwillia:
The test failures appear to be legitimate:
```
> /usr/bin/pylint --rcfile=jenkins/test/validators/.pylintrc openshift_tools/monitoring/hawk_sender.py
Pylint failed:
Pylint failed:
************* Module openshift_tools.monitoring.hawk_sender
I: 32, 0: Locally disabling too-many-arguments (R0913) (locally-disabled)
C: 92, 4: Invalid method name "update_tags_for_key_with_regex_rules" (invalid-name)
R: 92, 4: Method could be a function (no-self-use)
```

PR #2088, opened 2017-02-15
Carry restart hosts fix from upstream until next vendoring.

dgoodwin:
This is commit 781d2a1dc87bdc37f02ea6f1a3e83abb666f9c49 in
openshift-ansible.

It has been merged and will therefore be picked up next time we vendor
in a new version.


wshearn:
For easy reference https://github.com/openshift/openshift-ansible/pull/3333

wshearn (approved):

PR #2093, opened 2017-02-15
Fix issue with un-testable builds failing

tiwillia:
Builds initiated from comments lacking the "[test]" string were failing
due to incorrectly testing for an empty string rather than a null value.

Also, only post a failure status once, in the Jenkinsfile. It will be
the testing script's responsibility to post the success status, to
ensure it is posted as soon as possible.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -50,10 +50,8 @@ node {
         }
         echo "Trigger: ${trigger}"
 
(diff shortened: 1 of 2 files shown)


tiwillia:
@twiest this is a small follow-up to fix a few small issues, could you please review?

tiwillia:
Sorry, made a mistake, pushed the wrong branch.

PR #2095, opened 2017-02-15
Fix issue with un-testable builds failing

tiwillia:
Builds initiated from comments lacking the "[test]" string were failing
due to incorrectly testing for an empty string rather than a null value.

Also, only post a failure status once, in the Jenkinsfile. It will be
the testing script's responsibility to post the success status, to
ensure it is posted as soon as possible.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -50,6 +50,8 @@ node {
         }
         echo "Trigger: ${trigger}"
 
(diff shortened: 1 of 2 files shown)


tiwillia:
@twiest Sorry about the re-post, can you please review this small fix-up when you have a chance?

twiest (approved):

PR #2096, opened 2017-02-16
Updated metrics role

dak1n1:
* Run-once is applied at role level.
* Since `serial` cannot effectively be used with `run_once`, service restarts and config change was moved to a separate task.


dak1n1:
Approved in PR #2091

PR #2098, opened 2017-02-16
Add a git fetch module.

dgoodwin:
Assumes the repository is already checked out, and the remote exists,
otherwise we consider this an error.

If nothing is fetched, we return no change.


dgoodwin:
@mwoodson approved #2097.

PR #2106, opened 2017-02-20
Enable openshift-tools rpm building and pushing upon merge

tiwillia:
This change, when configured properly, will enable the building of rpms upon a pull request being merged. The base branch of the pull request is considered and the built rpms are only pushed to that branch's yum repository.

This will require the configuration of 3 yum repos in s3 buckets, one for int, stg, and prod. Pull requests merged to other base branches will be ignored.

Note that this change also refactors much of the code for testing to allow functions to be shared between the test and build process.

Once merged, the following changes will need to be made to enable builds:
  - The template will need to be re-deployed, with the new secrets specified
  - The new Jenkinsfile will need to be copied or pulled into the jenkins job
  - s3 buckets should be set up and ready with the user and accesskey specified in the template secrets
  - The oso-host-monitoring container will need to be re-built and pushed to all nodes, as new dependencies have been added

I expect this to fail CI testing until merged due to refactoring code out of the testing process.


tiwillia:
I'm blocked on this for a couple reasons:

1) Building without pushing to an s3 repository is going to make tags for the built rpms and push them to the openshift-tools repo. Unless we are okay with a lot of unused version bumps, we probably don't want to tag (and therefore build) until we have an s3 repo set up to push to.

2) The 'tito build' command that we usually use requires mock configuration that is only available on a particular system. The configuration does not appear to be something we can make public. If needed, I may be able to include the whole configuration as a secret, but I'd like to find out for sure if its necessary.

tiwillia:
This should be ready to go now. We aren't pushing tags and we are just going to do test builds for now until we can determine for sure whether we need the mock builder and its configuration to build the rpms properly.

I'm having some trouble testing in my environment though.

PR #2108, opened 2017-02-20
Dependencies for building rpms upon merging

tiwillia:
To allow for adequate testing, we need to merge in these dependency changes and rebuild the container.


tiwillia:
re-[test] please now that stg pr is merged.

PR #2122, opened 2017-02-22
Properly wait for metrics deploy to complete.

dgoodwin:
Includes some fixes for logging as well.


dgoodwin:
@wshearn approved #2115.

PR #2124, opened 2017-02-22
router DC changes for setting x_forwarded_for header

joelddiaz:
With OpenShift 3.3.1.11 and 3.4.1, the OpenShift router image accepts the environment var ROUTER_USE_PROXY_PROTOCOL to have the router pod use the PROXY protocol.

This allows propogation of the X-FORWARDED-FOR HTTP header showing the client IP address instead of the AWS ELB's IP.

This will be the default setting going forward. This requires the AWS ELB to be configured to speak the PROXY protocol as well (https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html).

For GCP ELBs, there is no need for this var, so only apply the extra env var when cloud is 'aws'.

### DIFF
--- a/ansible/roles/openshift_secure_router/tasks/main.yml
+++ b/ansible/roles/openshift_secure_router/tasks/main.yml
@@ -8,6 +8,13 @@
   - "{{ ossr_default_router_keys }}"
   - "{{ ossr_default_router_cacert }}"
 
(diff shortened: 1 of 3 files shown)


joelddiaz:
👍 #2012 in stg

PR #2134, opened 2017-02-23
Support numeric values from ops-metric-client cli

zgalor:
Fix a bug (hawkular only) that sent all values from cli as string metric.
The fix identifies if the value is a number and sends it as gauge.

# Before the fix:
`> ops-metric-client -k docker.ping -o 0 --debug`
`UniqueMetric('node.example.com', 'docker.ping', '0')`

# After the fix:

`> ops-metric-client -k docker.ping -o 0 --debug`
`UniqueMetric('node.example.com', 'docker.ping', 0.0)`

Also allowing to force string values for floats using CLI flag:
`> ops-metric-client -k docker.ping -o 0 -m string --debug`
`UniqueMetric('node.example.com', 'docker.ping', '0')`

### DIFF
--- a/scripts/monitoring/ops-metric-client.py
+++ b/scripts/monitoring/ops-metric-client.py
@@ -33,6 +33,7 @@ ops-metric-client -s --discovery-key filesys --macro-string #FILESYS --macro-nam
 
 """
 
+import json
 import argparse
 from openshift_tools.monitoring.metric_sender import MetricSender, MetricSenderHeartbeat
 import yaml
@@ -64,6 +65,21 @@ class OpsMetricClient(object):
 
         self.metric_sender.send_metrics()
 
+    @staticmethod
+    def adjust_type(val, metric_type):
+        """
+        :param val: string representing metric value
+        :param metric_type: string or numeric
+        :return:  if metric_type isn't string - value converted to float or int according to json parsing
+        """
+        if metric_type == 'string':
+            return val
+
+        try:
+            return json.loads(val)
+        except ValueError:
+            return val
+
     def parse_args(self):
         """ parse the args from the cli """
         parser = argparse.ArgumentParser(description='metric sender')
@@ -79,6 +95,9 @@ class OpsMetricClient(object):
         parser.add_argument('--debug', action='store_true', default=None, help='Debug?')
         parser.add_argument('-c', '--config-file', help='ops-metric-client config file',
                             default='/etc/openshift_tools/metric_sender.yaml')
+        parser.add_argument('-m', '--metric', choices=['numeric', 'string'],
+                            default='numeric',
+                            help='use specific metrics type [numeric, string]')
 
         key_value_group = parser.add_argument_group('Sending a Key-Value Pair')
         key_value_group.add_argument('-k', '--key', help='metric key')
@@ -133,7 +152,8 @@ class OpsMetricClient(object):
         # Get tags from command line args
         tags = dict([i.split("=")[0], i.split("=")[1]] for i in self.args.tags) if self.args.tags else {}
 
-        self.metric_sender.add_metric({self.args.key : self.args.value}, key_tags=tags)
+        self.metric_sender.add_metric({self.args.key : self.adjust_type(self.args.value, self.args.metric)},
+                                      key_tags=tags)
 
     def add_dynamic_metric(self):
         """ send zabbix low level discovery item to zagg """


zgalor:
@yaacov @twiest @simon3z please review

zgalor:
@tiwillia pylint test fails with reason that seems illogical:
`C: 68, 4: Missing method docstring (missing-docstring)`

Which I already fixed:
https://github.com/zgalor/openshift-tools/blob/e51961a8182192ca06429bca34017cd367792948/scripts/monitoring/ops-metric-client.py#L69

I'm afraid maybe it got stuck on the older output and didn't rerun.
Can you please take a look?

scripts/monitoring/ops-metric-client.py
```diff
@@ -64,6 +64,14 @@ def run(self):
 
         self.metric_sender.send_metrics()
 
+    @staticmethod
+    def adjust_type(val):
+        """ convert float values to float type. keep string as is """
+        try:
+            return float(val)
```
simon3z: @zgalor @twiest any way to know the type in advance instead of checking if something can be transformed into a float or not?

zgalor: I can add a cli parameter to indicate metric type. 
Problem is what will be the default? for backward compatibility with existing jobs?

simon3z: @zgalor @twiest IIUC zabbix knows the type of the metrics because you define them there in advance. On the other hand hawkular defines the type at collection-time.
Is it worth to keep this information (the type) in two places? Is there any way to unify it? (I can't think of a way to unify it).

zgalor: @simon3z the way I see it there are 3 ways to go:
1. What we are doing right now - for hawkular, we are deducing type from the metric itself, which can be both unsafe, and also limited: (for example: With the solution above, we can't send "1.3" as string metric)

2. A different solution would be to have the collector specifically state the metric type (with a default of gauge) and use that only for hawkular.

3. If we want these to be kept in one place, each collector could parse in runtime the vars template file that defines the type of this metric for zabbix. It's inefficient runtime-wise, but we will avoid hawk-zagg inconsistency.

twiest: My preference is for #1, but I could live with #2.  I don't really like #3.

Could we do #1, but allow for an override so that we could give "1.3" and set the type to string?

In other words, if we specify the type, use it. If we don't specify a type, then attempt to detect it.

twiest:
[test]

zgalor:
@twiest changed as requested. Also updated comment with usage example. Please review.

simon3z:
@zgalor @twiest WRT naming, do we really want to use "gauge"? In Hawkular they're a specific data-series. Since I assume we'll be pushing to any kind of numeric data-series: gauges and counters, maybe we should pick a different name? Maybe "numeric"?

zgalor:
@simon3z @twiest renamed 'gauge' to 'numeric'. Also updated PR title

scripts/monitoring/ops-metric-client.py
```diff
@@ -64,6 +64,21 @@ def run(self):
...
+        """
+        :param val: string representing metric value
+        :param metric_type: string or numeric
+        :return: value converted to float if metric_type isn't string
+        """
+        if metric_type == 'string':
+            return val
+
+        try:
+            return float(val)
```
twiest: Hmm, have you tested this against Zabbix?

Zabbix has int and float as 2 separate types. So I'm worried that all of our int based metrics will fail if this float conversion changes the value from 4 to 4.0 when parsed to json.


```python
$ python
Python 2.7.5 (default, Aug  2 2016, 04:20:16) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json.dumps({'a': 4})
'{"a": 4}'
>>> json.dumps({'a': 4.0})
'{"a": 4.0}'
>>> json.dumps({'a': float(4)})
'{"a": 4.0}'
>>> 

```

zgalor: @twiest I changed to `eval` that should convert to proper numeric type for both int and float.

scripts/monitoring/ops-metric-client.py
```diff
@@ -64,6 +64,21 @@ def run(self):
...
+        """
+        :param val: string representing metric value
+        :param metric_type: string or numeric
+        :return: value converted to float if metric_type isn't string
+        """
+        if metric_type == 'string':
+            return val
+
+        try:
+            return eval(val)
```
simon3z: @zgalor isn't this bad security-wise, and anyway a bad-practice?

zgalor: So it seems.
I could do the same with `json.loads` as we are trying to imitate it's behaviour

scripts/monitoring/ops-metric-client.py
```diff
@@ -64,6 +65,21 @@ def run(self):
...
+        """
+        :param val: string representing metric value
+        :param metric_type: string or numeric
+        :return: value converted to float if metric_type isn't string
+        """
+        if metric_type == 'string':
+            return val
+
+        try:
+            return json.loads(val)
```
zgalor, line 79: @cben do you have a better practice for this?

cben: LGTM.  Ideally ValueError would be an error, requiring either quotes or type 'string' for strings, but if you can't due to backwards compatibility, this is as good as it gets.

P.S. will you support other `metric_type`s?  E.g. 'float' with val == '1', should it parse as 1.0 or 1?

zgalor: Not planning on getting to that granularity. currently supporting string/numeric as metric_type

scripts/monitoring/ops-metric-client.py
```diff
@@ -64,6 +65,21 @@ def run(self):
 
         self.metric_sender.send_metrics()
 
+    @staticmethod
+    def adjust_type(val, metric_type):
+        """
+        :param val: string representing metric value
+        :param metric_type: string or numeric
+        :return: value converted to float if metric_type isn't string
```
twiest: This comment is no longer valid. It'll convert it to float or to int depending on parsing.

scripts/monitoring/ops-metric-client.py
twiest: We need to ensure compatibility with Zabbix. This is a core script and has the possibility to cause serious issues if these changes are incompatible with zabbix.

So, I'd like to see the following things tested:

1) send an int to a zabbix item that is also an int and ensure the value makes it into zabbix

2) send a float to a zabbix item that is also a float and ensure the value makes it into zabbix

3) send a string to zabbix item that is also a string and ensure the value makes it into zabbix

This should be testable with either the zaio or the new cluster up stuff (it's been merged and the [local development doc ](https://github.com/openshift/openshift-tools/blob/prod/docs/local_development_monitoring.adoc) shows how to use it).

twiest (changes requested):
Changing to request changes.

scripts/monitoring/ops-metric-client.py
zgalor: @twiest I can try setting up zabbix on Sunday, but it will probably take me a long time, and most likely I will need assistance from someone from your team.
It will probably go much faster if someone on your side could just pull the PR and test locally.

P.S.
running with `-m string` imitates old script behaviour

twiest: Go ahead with your test, I don't think anyone on our team will have time before then.

zgalor: Tested!!!!! Against my own local zabbix (with the help of @joelddiaz and @twiest )

float: docker.storage.data.space.available
int: docker.container.existing.dns.resolution.failed
string: kernel.uname.distro

All viewed as sent on the web.
@twiest ok to merge?

twiest (approved):

twiest:
@zgalor thanks for testing this.

Please let me know when the prod PR is ready.

PR #2135, opened 2017-02-23
raise exception when walking through object path

joelddiaz:
if we're given path a.b.c and the existing object is:
    a:
     b:
      - item1
    raise an exception due to unexpected objects found while traversing the path
    
also, add_entry assumes new dicts for each sub element when creating elements besides the final assignment value.
doing something like a.b.c[0] = 12 where 'c' doesn't exist raises an exception
    
add test cases to cover:
 access path that differs from existing object
 create new objects with an embedded list in the path
 create new object with a list at the end (define the end list in the passed in 'value' to avoid this exception)


joelddiaz:
[test]

kwoodson (approved):

PR #2137, opened 2017-02-23
Make ops-runner random sleeps deterministic by host & check name

joelsmith:
With this change, for a check that runs every hour on every host,
a random sleep of 3600 will evenly spread the run times out across
the environment while guaranteeing that each check will run with
the same interval for a single host.

Also, cheer up a sad pylint:
"Attribute 'args' defined outside __init__ (attribute-defined-outside-init)"


joelsmith:
@joelddiaz gave 👍 in stg PR #2136

PR #2142, opened 2017-02-24
Adding oadm_policy_group module

jupierce:


jupierce:
Module is being moved to openshift-ansible: openshift/openshift-ansible#3534

PR #2149, opened 2017-02-24
changing zagg target to service IP from skydns provided local dns name

ivanhorvath:


ivanhorvath:
Was pondering if we should use another check and make sure whatever is given is an ip address, but the ansible jinja filter ipaddr requires more packages to be installed, so I left it out for now. @twiest you wanted the container to not even start up if we can't get a connection? So we should check if it can connect to the zagg on host-monitoring container startup?

docker/oso-host-monitoring/centos7/root/config.yml
```diff
@@ -37,6 +37,25 @@
...
+      patterns: "system*kubeconfig"
+      use_regex: False
+    register: files
+
+  - name: get service IP for the service
+    command: >
+      /usr/bin/oc --config={{ files.files.0['path'] }}
+      get svc zagg-service
+      --template='{% raw %}{{.spec.clusterIP}}{% endraw %}'
+    register: svcipout
```
twiest: Will fail on tower. 

Also, please add a "FIXME: change to use openshift ansible module".

twiest (changes requested):

twiest (approved):

PR #2151, opened 2017-02-24
added the vendor_openshift_ansible_rpms tools role; update the vendor rpm to use it

mwoodson:


mwoodson:
@kwoodson approved #2150

PR #2157, opened 2017-02-27
Fix at-login check for AWS creds for zsh users

joelsmith:


joelsmith:
@kwoodson and @mwoodson gave 👍 in stg PR #2148

PR #2159, opened 2017-02-27
host-monitoring: Have docker mount needed paths for hostpkg checks

joelsmith:
Also remove double-added /host/var/lib/rpm and generate
rhel7 and centos7 versions of the host monitoring container


joelsmith:
@kwoodson gave 👍 in stg PR #2158

PR #2161, opened 2017-02-27
changing zagg target to service IP from skydns provided local dns name

ivanhorvath:


ivanhorvath:
Approved by @twiest in #2149

PR #2162, opened 2017-02-27
Enable cron-send-security-updates-count cron job

joelsmith:


joelsmith:
@dak1n1 gave 👍 in stg PR #2160

PR #2165, opened 2017-02-27
Disable rhsm while running hostpkg checks

joelsmith:
Also, don't forget to bind mount the host's /var/lib/rpm


joelsmith:
@rhdedgar gave 👍 in stg PR #2164

PR #2167, opened 2017-02-27
Add generated changes for rhel7 and centos7 in oho-host-monitoring

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #2166

PR #2169, opened 2017-02-27
new MetricSender/ZaggSender uses metric_sender.yaml for config info

joelddiaz:
generate that file from the intalled /etc/openshift_tools/monitoring-config.yml


joelddiaz:
👍 #2168 in stg

PR #2171, opened 2017-02-28
No sftp needed in SSO SSH endpoint config

joelsmith:


joelsmith:
@rhdedgar gave 👍 in stg PR #2170

PR #2175, opened 2017-03-01
updated oo_ami copy to have new sa-east-1 region

mwoodson:


mwoodson:
@kwoodson approved #2174

PR #2177, opened 2017-03-02
add cronjob for internal-pods-check

zhiwliu:


ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -357,6 +357,11 @@ host_monitoring_cron:
   minute: "*/10"
   job: /usr/bin/delete-stuck-projects &>> /var/log/delete-stuck-projects
 
+- name: Do a quick check for the internal pods status and location every 1 hour
+  minute: "15"
+  hour: "*/1"
```
drewandersonnz: for "every hour" we just use *:

```
hour: "*"
```

I think * is the default, so you may remove this line.

drewandersonnz (approved):
+1

PR #2183, opened 2017-03-02
don't apply PROXY env var changes to 3.3 clusters

joelddiaz:
moving to PROXY-by-default is for 3.4, so don't add this setting for pre-3.4 clusters

also update README to more accurately reflect variables to be passed into the role


joelddiaz:
👍 #2182 in stg

PR #2191, opened 2017-03-06
use centos-release-openshift-origin repo on CentOS

joelddiaz:
this allows us to use different ansible versions than what is found on EPEL
this allows us to use more up-to-date origin RPMs (for 'oc' and 'oadm' binaries)


kwoodson (approved):

joelddiaz:
thanks @andrewklau for #2186 ... we'll just use the OpenShift for Cent repo in general rather than just for Ansible

PR #2198, opened 2017-03-06
fixed spec file for config tag checks

mwoodson:


mwoodson:
@kwoodson approved #2197

PR #2200, opened 2017-03-07
adding a check to only look for zagg-service is zagg is active in config

ivanhorvath:


ivanhorvath:
@twiest this should take care of the problem with cmops,

joelddiaz (approved):

PR #2202, opened 2017-03-07
added the zabbix config loop client template

mwoodson:


mwoodson:
@kwoodson approved #2201

PR #2205, opened 2017-03-07
enabled the checks for the host monitoring to check the config loop tags

mwoodson:


mwoodson:
@ivanhorvath approved #2204

PR #2208, opened 2017-03-07
fixed the config loop client triggers

mwoodson:


mwoodson:
@joelddiaz approved #2207

PR #2211, opened 2017-03-07
adding a check to only look for zagg-service is zagg is active in config

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2200

PR #2213, opened 2017-03-09
vendored openshift ansible 3.5.28

mwoodson:

PR #2214, opened 2017-03-09
vendored openshift ansible 3.5.28

mwoodson:


mwoodson:
@jupierce approved #2213 

vendored rpms so will fail botcheck

PR #2216, opened 2017-03-09
Add check to see if Zabbix is running the most recent LTS version

joelsmith:


scripts/monitoring/cron-send-zabbix-too-old.py
```diff
@@ -0,0 +1,152 @@
...
+import gzip
+import io
+import os
+import rpm
+import lxml.etree
+import lxml.objectify
+import subprocess
+#pylint: disable=import-error
+from openshift_tools.monitoring.metric_sender import MetricSender
+#pylint: enable=import-error
```
kwoodson, line 26: Is this statement needed?

kwoodson (commented):
Small nit.  Everything else looks good.

scripts/monitoring/cron-send-zabbix-too-old.py
joelsmith: It's not needed, but I like re-enabling disabled checks so that the linter can find any other potential issues that do happen to be valid problems.

joelsmith:
I ran pylint manually and it passes. The check is exploding, so merging without passed checks.

PR #2217, opened 2017-03-09
Add check to see if Zabbix is running the most recent LTS version

joelsmith:


joelsmith:
@kwoodson gave 👍 in stg PR #2216

joelsmith:
I ran pylint manually and it passes. The check is exploding, so merging without passed checks.

PR #2218, opened 2017-03-09
Adding debug flag so we can peer into the docker container.

kwoodson:


tiwillia:
Thinking about how this is going to be used.

You can re-run a failed jenkins job. When you do, it allows you to modify the Jenkinsfile before re-running the build. You can add this env variable to the envs we pass to the build:
```
        openshiftBuild buildConfig: 'openshift-tools-test', env: [[ name: 'PULL_REQUEST', value: pull_request_string ], [ name: 'BUILD_URL', value: env.BUILD_URL ], [name: 'OO_PAUSE_ON_BUILD', value: "true"]], showBuildLogs: true
```

kwoodson:
[test]

PR #2221, opened 2017-03-10
Add generic pod template deployer role

themurph:


ansible/roles/openshift_template_deployer/README.md
```diff
@@ -0,0 +1,75 @@
+openshift_volume_provisioner
```
mwoodson: This file needs to be updated with the variables that are now being used, and the purpose of it.

ansible/roles/openshift_template_deployer/defaults/main.yml
```diff
@@ -0,0 +1,54 @@
+---
+ostd_rpm: False
+ostd_master_config_edits: False
+ostd_node_selector: type=infra
+ostd_project_display_name: "{{ ostd_project_name }}"
+
+#- name: SYNC_PERIOD
```
mwoodson: we should clean these up. I don't think they are needed.

ansible/roles/openshift_template_deployer/handlers/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: restart openshift master services
+  service:
+    name: "{{ item[1] }}"
+    state: restarted
+  delegate_to: "{{ item[0] }}"
+  with_nested:
+  - "{{ osvp_master_nodes }}"
```
mwoodson: this variable is wrong.  it needs the new name on it.

ansible/roles/openshift_template_deployer/meta/main.yml
```diff
@@ -0,0 +1,14 @@
...
+  company: Red Hat, Inc
+  license: ASL 2.0
+  min_ansible_version: 1.2
+  platforms:
+  - name: EL
+    versions:
+    - 7
+dependencies:
+- role: tools_roles/lib_yaml_editor
+- role: tools_roles/lib_openshift_3.2
```
mwoodson: we might want to switch this over to use the openshift-ansible lib_openshift and lib_yaml_editor.

mwoodson (changes requested):

ansible/roles/openshift_template_deployer/handlers/main.yml
themurph: We all waivered on just removing the handler, but I'll update the variable just in case we decide to keep this file.

ansible/roles/openshift_template_deployer/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
+---
+# Ensure needed variables are set
+# This is a a place to verify needed values
+- name: Ensure that osvp is set
+  fail:
+    msg:  "Please ensure that 'osvp_<variables>' are set!"
```
jupierce: Should be "Please ensure that **ostd**_variables are set!"

ansible/roles/openshift_template_deployer/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+- name: pause 10 seconds when services are restarted
+  pause:
+    seconds: 10
+  when: master_config_changes.changed
+
+# Enusre project is created
+- name: create project
+  oadm_project:
+    name: "{{ ostd_project }}"
+    node_selector: "{{ ostd_node_selector|default(omit)}}"
```
jupierce: Missing space before }} on several of these. @mwoodson suggested this as a style guideline.

ansible/roles/openshift_template_deployer/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+#- debug: var=templateout
+
+# TODO: What if we want to create the template in namespace A but create it's objects in namespace B?
+# This processes the template with variables
+- name: Deploy the app
+  oc_process:
+    namespace: "{{ ostd_project }}"
+    template_name: "{{ ostd_template_name }}"
+    create: True
+    params: "{{ ostd_template_params|default(omit)}}"
```
jupierce: Space before }}

ansible/roles/openshift_template_deployer/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+    name: "{{ ostd_rpm }}"
+    state: installed
+  when: ostd_rpm
+
+- name: Make template changes if needed
+  yedit:
+    src: "{{ ostd_template_file }}"
+    key: "{{ item.key }}"
+    value: "{{ item.value }}"
+  when: ostd_template_edits
```
jupierce, line 26: I think all of these  `when: X` statements need to be  `when: X is defined`    OR   they need to be initailized to False as you do for `ostd_master_config_edits` in defaults/main.yml . However, I think the `is defined` route is more expressive.

ansible/roles/openshift_template_deployer/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+  with_items: "{{ ostd_template_edits }}"
```
jupierce, line 27: In addition, if the task protected by the `when:` uses the variable, I needed to do something like "{{ ostd_template_edits | default(omit) }}" or else I would get an error that the variable was undefined.

ansible/roles/openshift_template_deployer/defaults/main.yml
```diff
@@ -0,0 +1,54 @@
+---
+ostd_rpm: False
+ostd_master_config_edits: False
+ostd_node_selector: type=infra
+ostd_project_display_name: "{{ ostd_project_name }}"
```
jupierce: I think this should be `{{ ostd_project }}`  instead.

ansible/roles/openshift_template_deployer/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+  when: master_config_changes.changed
+
+# pause if restarted
+- name: pause 10 seconds when services are restarted
+  pause:
+    seconds: 10
+  when: master_config_changes.changed
+
+# Enusre project is created
+- name: create project
```
jupierce: Received a pretty odd error on this task: 
```
TASK [tools_roles/openshift_template_deployer : create project] ****************
fatal: [cicd-master-14aa4]: FAILED! => {"changed": false, "failed": true, "msg": {"cmd": "/usr/bin/oc -n jupierce-test replace -f /tmp/jupierce-test", "results": {}, "returncode": 1, "stderr": "Error from server (BadRequest): error when replacing \"/tmp/jupierce-test\": Namespace in version \"v1\" cannot be handled as a Namespace: [pos 93]: json: expect char '\"' but got char 'n'\n", "stdout": ""}}
```
The project seemed to have been created the first time I ran the task, but this error occurs now.

ansible/roles/openshift_template_deployer/tasks/main.yml
themurph: Fixed

ansible/roles/openshift_template_deployer/tasks/main.yml
themurph: fixed

themurph:
@mwoodson and @jupierce - All updates we discussed this morning are in and all tests passed

mwoodson (approved):

jupierce:
@themurph There was a small change that needed to be made in your absence (lib_yaml_editor => lib_utils). #2240 has been created to introduce that change.

PR #2225, opened 2017-03-13
Add view role to hawkular SA before launching metrics deployer.

dgoodwin:


dgoodwin:
@wshearn approved #2222.

PR #2226, opened 2017-03-13
Pass in logging image version

wshearn:


ansible/roles/openshift_logging/defaults/main.yml
```diff
@@ -1,4 +1,5 @@
 osalog_fluentd_nodes:
 - type=compute
 osalog_image_prefix: registry.access.redhat.com/openshift3/
+osalog_image_version: v3.4
```
dgoodwin, line 4: This feels like it will get forgotten, I would suggest no default and error if not set, that indicates a problem IMO.

dgoodwin (changes requested):

dgoodwin (approved):

PR #2229, opened 2017-03-13
updated the subdomain variable in the byo generator

mwoodson:


mwoodson:
@wshearn approved #2228

PR #2231, opened 2017-03-13
Mount the entire host FS on the host monitoring container's /host

joelsmith:
The entirety of the hosts filesystem will now be visible (read-only)
in the host monitoring container so that security scans can be
run on the host from within the container.


joelsmith:
@kwoodson gave 👍 in stg PR #2230

PR #2232, opened 2017-03-14
Adding ansible roles to manage role, cluster-role, and scc policy for groups and users

jupierce:
@mwoodson ptal


mwoodson:
:+1:

PR #2101, opened 2017-02-20
cron-send-aws-instance-health.py

drewandersonnz:

PR #2236, opened 2017-03-14
cron-send-aws-instance-health.py uses the aws api check for known hos…

drewandersonnz:
…t issues across all regions

+1 from @BlueShells, and reviewd by @twiest on https://github.com/openshift/openshift-tools/pull/2101

cron additions are waiting still:
https://github.com/openshift/openshift-tools/pull/2235


drewandersonnz:
from #2101 : pylinted and tested on bastion. bot is missing pylint. email thread already started.

PR #2237, opened 2017-03-14
Fix for empty string getting converted to None and treated as None.

kwoodson:


mwoodson (approved):

kwoodson:
Bot failed because pylint is missing.

PR #2239, opened 2017-03-14
Start specifying v3.4 for the metrics image version.

dgoodwin:


dgoodwin:
@wshearn approved #2238.

PR #2247, opened 2017-03-16
fixed naming issue for storageclass

mwoodson:


mwoodson:
@wshearn approved #2247

PR #2254, opened 2017-03-20
logging container outbound TCP connections

rhdedgar:


ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+    - chain: oo-outbound-logging
+      line:  ":oo-outbound-logging - [0:0]"
+      after: "filter$"
+      regex: "Create oo-outbound-logging chain"
+      test: True
+      pos: 1
+
+    outbound_logging_rules:
+    - chain: oo-outbound-logging
+      line:  "-A oo-outbound-logging -m comment --comment \"Log outbound TCP connections with this chain\" -m limit --limit 50/second --limit-burst 50 -j LOG --log-prefix \"tcp_out_logging: \""
```
joelsmith: I don't think we need implementation comments in the iptables rules themselves. Since you do need an identifier to match your regexes on, I think you could just create an identifier for each rule that will for sure be unique. I'd rather see something like:
```
outbound_logging_rules:
# Log new outbound TCP connections (capped at 50/second)
- chain: oo-outbound-logging
  line:  "-A oo-outbound-logging -m comment --comment id_outbound_tcp_log_1_ -m limit --limit 50/second --limit-burst 50 -j LOG --log-prefix \"tcp_out_logging: \""
  after: "^-A FORWARD"
  ...
```

ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+      after: "^-A FORWARD"
+      regex: "Log outbound TCP connections with this chain"
+      test: True
+      pos: 1
+
+    - chain: oo-outbound-logging
+      line:  "-A oo-outbound-logging -m comment --comment \"We use this pattern since there is no LOG_AND_RETURN target\" -m limit --limit 50/second --limit-burst 50 -j RETURN"
```
joelsmith: Same thing goes here w.r.t. comments. This comment doesn't need to be in the iptables rule, but should be document this chain here in this YAML file. Maybe this one could be `-m comment --comment id_outbound_tcp_log_2_`

ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+      after: "Log outbound TCP connections with this chain"
+      regex: "We use this pattern since there is no LOG_AND_RETURN target"
+      test: True
+      pos: 2
+
+    - chain: oo-outbound-logging
+      line: "-A oo-outbound-logging -m comment --comment \"Drop 1 or more logs for 60 seconds after this message\" -m limit --limit 1/minute --limit-burst 1 -j LOG --log-prefix \"tcp_out_logging: logdrop \""
```
joelsmith: Same thing goes here w.r.t. comments. This comment doesn't need to be in the iptables rule, but should be document this chain here in this YAML file.

ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+      after: "filter$"
+      regex: "Create oo-outbound-logging chain"
+      test: True
+      pos: 1
+
+    outbound_logging_rules:
+    - chain: oo-outbound-logging
+      line:  "-A oo-outbound-logging -m comment --comment \"Log outbound TCP connections with this chain\" -m limit --limit 50/second --limit-burst 50 -j LOG --log-prefix \"tcp_out_logging: \""
+      after: "^-A FORWARD"
+      regex: "Log outbound TCP connections with this chain"
```
joelsmith: change to `id_outbound_tcp_log_1_`

ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+      test: True
+      pos: 1
+
+    - chain: oo-outbound-logging
+      line:  "-A oo-outbound-logging -m comment --comment \"We use this pattern since there is no LOG_AND_RETURN target\" -m limit --limit 50/second --limit-burst 50 -j RETURN"
+      after: "Log outbound TCP connections with this chain"
+      regex: "We use this pattern since there is no LOG_AND_RETURN target"
```
joelsmith: Change to `id_outbound_tcp_log_2_`

ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+    - chain: oo-outbound-logging
+      line:  "-A oo-outbound-logging -m comment --comment \"Log outbound TCP connections with this chain\" -m limit --limit 50/second --limit-burst 50 -j LOG --log-prefix \"tcp_out_logging: \""
+      after: "^-A FORWARD"
+      regex: "Log outbound TCP connections with this chain"
+      test: True
+      pos: 1
+
+    - chain: oo-outbound-logging
+      line:  "-A oo-outbound-logging -m comment --comment \"We use this pattern since there is no LOG_AND_RETURN target\" -m limit --limit 50/second --limit-burst 50 -j RETURN"
+      after: "Log outbound TCP connections with this chain"
```
joelsmith: Change to `id_outbound_tcp_log_1_`

joelsmith (changes requested):
So far I just have one thing that I think ought to be changed, but it's a pattern that you're using across the entire `openshift_outbound_tcp_logging/tasks/main.yml` file. I put comments on the first few, but IMO, we should put impl comments as YAML comments and use a consistent identifier naming convention for regex and after matching, and do that across the board.

joelsmith (approved):
Looks great!

PR #2256, opened 2017-03-21
fixed an issue  with the openshift default ns settings role

mwoodson:


mwoodson:
@kwoodson approves this change

PR #2258, opened 2017-03-21
Configure the OpenShift Online haproxy

themurph:
/cc @jupierce @mwoodson PTAL


mwoodson (approved):

ansible/roles/openshift_online_ha_proxy/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+
+- name: create the customrouter configmap
+  command: >
+    oc create configmap customrouter -n default --from-file=/etc/openshift-online/templates/haproxy-config.template
+  run_once: true
+
+# The oc_volume command does not currently support a configmap source. Falling back to command module.
+- name: Add a volume to the router
+  command: >
+    oc volume dc/router --add --overwrite --name=config-volume --mount-path=/var/lib/haproxy/conf/custom --source='{"configMap":{"name":"customrouter"}}'
```
0xmichalis: Use `oc set volume` - `oc volume` should be deprecated if it hasn't a notice already. Also instead of `--source` you can use `--configmap-name`

jupierce: Thanks @kargakis . 
@themurph I tested with:
`oc set volume dc/router --add --overwrite --name=config-volume --mount-path=/var/lib/haproxy/conf/custom --configmap-name=customrouter`
Looks good.

PR #2259, opened 2017-03-21
change local dev script to allow it to work on RHEL/Cent

joelddiaz:
also in the cleanup script, reload firewall to default (ie permanent) settings


twiest (approved):
Looks great!!!

PR #2263, opened 2017-03-21
Configure the OpenShift Online haproxy

themurph:


themurph:
@mwoodson approved #2258

PR #2269, opened 2017-03-22
added cicd to the openshift-tools-scripts spec file

mwoodson:


mwoodson:
@themurph approved #2268

PR #2276, opened 2017-03-22
Make simpleSAML authorizeyaml mod support users in multiple roles

joelsmith:


joelsmith:
@blentz gave 👍 in stg PR #2275

PR #2277, opened 2017-03-22
updated 3.5 vendored rpms

mwoodson:

PR #2278, opened 2017-03-22
updated 3.5 vendored rpms

mwoodson:


mwoodson:
@kwoodson approved #2277

PR #2281, opened 2017-03-23
Allow parameters to be passed to cicd-control.sh.

bbguimaraes:
@jupierce (@joelsmith?): PTAL.

The diff is a little convoluted because of all the whitespace changes.
`git diff --ignore-all-space` gives a better view of what's being changed.

### DIFF
--- a/scripts/cicd/verify-cicd-operation.py
+++ b/scripts/cicd/verify-cicd-operation.py
@@ -19,6 +19,16 @@ import socket
 import logging
 import logging.handlers
 
+OPTION = r'--openshift-ansible=\d+\.\d+'
+ALLOWED_COMMANDS = (
+    ['install'],
+    ['install', OPTION],
+    ['update'],
+    ['update', OPTION],
+    ['delete'],
+    ['delete', OPTION],
+)
+
 def runner(program, *args):
     ''' string '''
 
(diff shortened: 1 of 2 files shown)


scripts/cicd/verify-cicd-operation.py
```diff
@@ -30,21 +40,29 @@ def runner(program, *args):
...
+    logger.addHandler(logging.handlers.SysLogHandler('/dev/log'))
 
-if match:
-    logger.info("%s Restricted key '%s' running command: %s" % (os.path.basename(__file__), keyname, cmd))
-    runner("/home/opsmedic/aos-cd/git/aos-cd-jobs/bin/cicd-control.sh", keyname, match.group("operation"))
+    keyname = sys.argv[1] if len(sys.argv) >= 2 else "<none specified>"
+    cmd = os.environ.get("SSH_ORIGINAL_COMMAND", "").split()
+    if verify(ALLOWED_COMMANDS, cmd):
+        logger.info("%s Restricted key '%s' running command: %s" % (os.path.basename(__file__), keyname, cmd))
+        runner("/home/opsmedic/aos-cd/git/aos-cd-jobs/bin/cicd-control.sh", keyname, *cmd)
```
joelsmith, line 62: We still need cicd-control.sh to be installed by RPM, not just some script from a repo. Does that script really take keyname as an argument?

joelsmith (changes requested):
The verification stuff looks good. We just have the issue of the `cicd-control` script not being installed on the host in an approved way.

scripts/cicd/verify-cicd-operation.py
jupierce: @joelsmith The RPM work is underway (e.g. #2269 and others ).

keyname is actually a cluster name (e.g. 'dev-preview-int').

jupierce:
Thanks @bbguimaraes! 
@themurph @mwoodson ptal . 
@themurph once merged, this will need to drive a new RPM build.

PR #2287, opened 2017-03-27
Added yum-install-check.sh to catch when yum fails to install a package

twiest:
The purpose of this PR is to fix our build bot. EPEL stopped shipping pylint (temporary problem, they're working on fixing that).

This exposed another issue, which is that if yum is installing multiple packages, as long as 1 of the packages succeeds, then it will not error if the other packages are missing.

To fix that, I:
* Added yum-install-check.sh (written by Joel Smith) to catch when yum fails to install a package
* Fixed container builds because of missing packages
* Added ARG to containers (now needed for OO_PAUSE_ON_BUILD to work)


twiest:
+1d in STG PR #2285

PR #2290, opened 2017-03-27
adding conditionals

rhdedgar:
Cleaning up unused vars, and making sure the config loop won't write if the rules already exist in memory or /etc/sysconfig/iptables.


drewandersonnz (approved):
assuming already tested

PR #2293, opened 2017-03-28
vendored new openshift-ansible 3.5 to 3.5.41

mwoodson:


mwoodson:
vendored version, will fail bot

PR #2299, opened 2017-03-29
Adding debug flag so we can peer into the docker container.

kwoodson:


kwoodson:
@tiwillia and I agreed that this was necessary to debug pipeline tests.

kwoodson:
[test]

PR #2301, opened 2017-03-29
 Use oc_volume to add the router configmap volume. 

bbguimaraes:
@jupierce, @themurph: PTAL.  The command works on a local `oc cluster up`.

Is there documentation on how the `meta/main.yml` dependencies are supposed to
work?  For my tests I hard-coded the paths to the modules, I'd like to be sure
they work with this change.

### DIFF
--- a/ansible/roles/openshift_online_ha_proxy/tasks/main.yml
+++ b/ansible/roles/openshift_online_ha_proxy/tasks/main.yml
@@ -31,10 +31,15 @@
     oc create configmap customrouter -n default --from-file=/etc/openshift-online/templates/haproxy-config.template
   run_once: true
 
-# The oc_volume command does not currently support a configmap source. Falling back to command module.
 - name: Add a volume to the router
-  command: >
-    oc set volume dc/router --add --overwrite --name=config-volume --mount-path=/var/lib/haproxy/conf/custom --configmap-name=customrouter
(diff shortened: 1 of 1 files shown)


kwoodson:
@bbguimaraes, this looks good.

@mwoodson @jupierce The issue here is that the configmap support lives in openshift-ansible.  Please correct me if I'm wrong but I'm not sure we have a solution yet for module precedence as this is depending on lib_openshift_3.2.

themurph (approved):
👍  Yes this looks good. @bbguimaraes  @kwoodson @jupierce  We are using the openshift-ansible roles since it's been made clear that going forward, that is the correct way to go:

```
dependencies:
  - role: "aos_{{ g_play_openshift_version  }}_roles/lib_utils"
  - role: "aos_{{ g_play_openshift_version  }}_roles/lib_openshift"
```

@bbguimaraes in your test environment/directory, you can add another softlink back to your openshift-ansible repo with the correct name:

```
aos_3.5_roles -> /home/chmurphy/git/openshift-ansible/roles/
```

@bbguimaraes if this doesn't make sense, ping me offline and we can talk it over

bbguimaraes:
@themurph: it does, thanks.  Confirming the playbook works without any changes
to `meta/main.yml`.

PR #2304, opened 2017-03-29
create empty contiv role

joelddiaz:
openshift-ansible installer assumes an empty contiv ansible role. create one.


joelddiaz:
👍 #2302 in stg

PR #2308, opened 2017-03-31
Use new oc_configmap to pay back tech debt

themurph:


ansible/roles/openshift_online_ha_proxy/tasks/main.yml
```diff
@@ -14,21 +14,21 @@
...
-# when oc attempted to read the temporary file created by oc_obj.
-# We have documented the work necessary to address these problems long term. In the interim, the command
-# module will be used.
-
-- name: Delete the configmap if it exists
-  command: >
-    oc delete configmap customrouter -n default --ignore-not-found
+- name: Delete existing config map if necessary
+  oc_configmap:
+    state: absent
```
kwoodson: Why the delete here?

ansible/roles/openshift_online_ha_proxy/tasks/main.yml
```diff
@@ -14,21 +14,21 @@
...
-# of the complex content in the customrouter configmap.
-# An attempt was made to use oc_obj to create the configmap with the necessary content, but we ran into errors
-# when oc attempted to read the temporary file created by oc_obj.
-# We have documented the work necessary to address these problems long term. In the interim, the command
-# module will be used.
-
-- name: Delete the configmap if it exists
-  command: >
-    oc delete configmap customrouter -n default --ignore-not-found
+- name: Delete existing config map if necessary
```
jupierce: @themurph I'm assuming @kwoodson's module is idempotent and will update the configmap if the haproxy-config.template has changed. I don't think there is a reason to delete any longer.

ansible/roles/openshift_online_ha_proxy/tasks/main.yml
themurph: See above: I'm removing it and then I'll re-push /cc @jupierce

jupierce (approved):

PR #2309, opened 2017-03-31
Use new oc_configmap to pay back tech debt

themurph:


themurph:
@jupierce approved #2308

PR #2319, opened 2017-04-03
updated generate byo to support logging, metrics, router, regis, sdn and service networks

mwoodson:


mwoodson:
@kwoodson approved #2311

PR #2322, opened 2017-04-03
Enabling timeout for subscription-manager call

jupierce:
@kwoodson ptal

### DIFF
--- a/ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
+++ b/ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
@@ -56,6 +56,9 @@ EXAMPLES = '''
 import re
 
 RHSM_BIN = '/usr/sbin/subscription-manager'
+# Old versions of subscription-manager would hang idefinitely (BZ 1406491) in
+# certain conditions. This timeout wrapper ensures it will always terminate.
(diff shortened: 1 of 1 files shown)


ansible/roles/lib_aos_modules/library/aos_rhsm_repos.py
```diff
@@ -151,7 +154,7 @@ def needs_update(self):
 
     def check_identity(self):
         """verify that this system is entitled"""
-        rcode, _, _ = self.module.run_command([RHSM_BIN, 'identity'])
+        rcode, _, _ = self.module.run_command(RHSM_ENTRYPOINT + ['identity'])
```
kwoodson, line 157: Was this bracket intentional?

jupierce: Still coming up to speed on python, but yes. RHSM_BIN was just a string; the new RHSM_ENTRYPOINT is a list. So:  
list(RHSM_ENTRYPOINT) + list(['identity']) = list(a combination of both) 

It is this combination that is passed to run_command.

jupierce:
ptal @mwoodson

jupierce:
The ops team decided to avoid this change by forcing all clusters to upgrade to subscription manager > 1.17

PR #2323, opened 2017-04-03
updated the byo generate to have options passed in for logging and metrics

mwoodson:


mwoodson:
@joelddiaz approved #2321

PR #2332, opened 2017-04-05
added support for lib_vendored_deps

mwoodson:


mwoodson:
@kwoodson approved #2331

PR #2335, opened 2017-04-05
Adding ability to return hosts from hostgroup.

kwoodson:


kwoodson:
@mwoodson :+1:  in #2334

PR #2339, opened 2017-04-06
log pruning output to a file in the container

stenwt:


stenwt:
@ivanhorvath reviewed #2338

PR #2345, opened 2017-04-06
updated the byo for rencrypt on the registry

mwoodson:


mwoodson:
@kwoodson approved #2344

PR #2357, opened 2017-04-08
template deployer should update rpm before installing

jupierce:
ptal @mwoodson

### DIFF
--- a/ansible/roles/openshift_template_deployer/tasks/main.yml
+++ b/ansible/roles/openshift_template_deployer/tasks/main.yml
@@ -15,7 +15,8 @@
 - name: install the "{{ ostd_rpm }}"
   yum:
     name: "{{ ostd_rpm }}"
-    state: installed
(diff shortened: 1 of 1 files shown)


jupierce:
https://atomic-e2e-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/aos-cd/job/cluster%252Foperation/71/console

mwoodson:
I don't think we want this here. This allows packages to be updated during the config loop automatically update packages. This is generally not what we want, because this causes unpredictability.

We need to handle upgrades in a separate task.  I did include this same task as latest in my reconciler installer because it's not in the config loop.

jupierce:
@mwoodson ack. This upgrade logic can be outside the deployer role so as to not impact the config loop. However, the current upgrade script did not trigger an actual RPM update for openshift-scripts-devpreivew, so I think it needs update_cache: true added.

PR #2362, opened 2017-04-10
Make Ops SSO app auto-regenerate its config when secrets change

joelsmith:


joelsmith:
@drewandersonnz gave 👍 in stg PR #2361

PR #2364, opened 2017-04-10
Don't restart SSO pods on secrets change

joelsmith:
This change is only fine because now the SSO pods monitor for secrets
changes and reconfig themselves when the secrets change
(which they can do thanks to the OpenShift feature that refreshes secrets
without a pod restart)

(Also, add autogenerated files of oso-saml-sso container)


joelsmith:
@drewandersonnz gave 👍 in stg PR #2363

PR #2366, opened 2017-04-11
Restart Apache httpd if secrets change

joelsmith:
Necessary when the cert/key pair that comes as a secret are changed

Also fix outdated comments w.r.t. pod restarts on secrets changes


joelsmith:
@drewandersonnz gave 👍 in stg PR #2365

PR #2367, opened 2017-04-12
Disable restricted name check for 3.5

jupierce:
Until https://github.com/openshift/origin/pull/13673/files is merged and installed, nothing prevents users from creating these restricted namespaces after the upgrade. A pointless check that would otherwise cause use to need to delete user data in dev-preview-prod was deemed removable by the technical team.


mwoodson (approved):

enj:
The API check is active in 3.5.1 so 3.5.0 is the last version this patch is valid for.

PR #2377, opened 2017-04-13
added httpd-tools to the os_utils to get installed during image creating and bootstrap

mwoodson:


mwoodson:
@kwoodson approved #2376

PR #2382, opened 2017-04-13
add cron job that cycles router pods if a cluster requires this

joelddiaz:
pass osohm_enable_dirty_router_restarts as True for this functionality


joelddiaz:
👍 #2381 in stg

PR #2384, opened 2017-04-13
add openjdk rpm for systems that we initiate installations from

joelddiaz:
required for the metrics installation piece of the openshift installer


joelddiaz:
👍 #2383 in stg

PR #2389, opened 2017-04-17
added the aws pre warm role

mwoodson:


mwoodson:
@kwoodson approved #2388

PR #2392, opened 2017-04-18
catch and ignore epipe errors

joelddiaz:
when doing something like:
  ossh --list | head

'head' will close the pipe and python will raise an IOError. It is okay for the program receiving the output of ossh to close the pipe once it is complete before processing all of the output.

Catch and ignore these EPIPE errors so that we can build things like:
  ossh --list | grep --max-count=1 clustername | awk '{ print $3 }'

...to get the environment for a particular cluster.


joelddiaz:
👍 #2391 in stg

PR #2393, opened 2017-04-18
changing interval for heartbeat to 1min and logging changes

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2393

PR #2394, opened 2017-04-18
Add new role binding for logging for 3.4

wshearn:


ansible/roles/openshift_logging/tasks/main.yml
```diff
@@ -166,6 +166,16 @@
   when: podout.results.results[0]['items'][0].status.phase != 'Succeeded'
   run_once: true
 
+- name: add cluster role binding to aggregated-logging-elasticsearch service account
+  oadm_policy_user:
```
kwoodson, line 170: My only concern here is that the `oadm_policy_user` module is now out of date and we should be deprecating them in favor of the upstream code in the future.  That should happen once we arrive at 3.5.

That being said this should work.

kwoodson (approved):

ansible/roles/openshift_logging/tasks/main.yml
wshearn: Once we arrive on 3.5 we will not be using this code. So that should be fine.

PR #2396, opened 2017-04-18
fixed the template deployer role

mwoodson:


mwoodson:
@jupierce approved #2395

PR #2401, opened 2017-04-18
added some debug to the zagg deployer

mwoodson:


mwoodson:
@ivanhorvath approved #2400

PR #2408, opened 2017-04-19
Added container usage regexes for zabbix and ops registry pods.

twiest:


twiest:
[test]

joelddiaz (approved):

twiest:
Merging without bot because the bot is down and this is needed urgently.

PR #2428, opened 2017-04-21
Zagg optimizations

twiest:

PR #2429, opened 2017-04-21
Zagg optimizations

twiest:


twiest:
Ivan +1d this in STG PR #2428

twiest:
Merging on a Friday because Ivan is on call and he +1d this.

PR #2433, opened 2017-04-25
added check for the project creation and deletion

zhiwliu:


scripts/monitoring/cron-send-project-operation.py
```diff
@@ -0,0 +1,130 @@
...
+    " delete the project "
+    try:
+        runOCcmd("delete project {}".format(config.namespace), base_cmd='oc')
+        time.sleep(commandDelay)
+    except Exception:
+        logger.exception('error delete project')
+def main():
+    """ check the project operation status """
+
+    logger.info('Starting project Create - %s', datetime.datetime.now().strftime("%Y-%m-%d %H:%M"))
```
drewandersonnz: logger will add date for you

scripts/monitoring/cron-send-project-operation.py
```diff
@@ -0,0 +1,130 @@
...
+        logger.exception('error creating new project')
+
+def delete_project(config):
+    " delete the project "
+    try:
+        runOCcmd("delete project {}".format(config.namespace), base_cmd='oc')
+        time.sleep(commandDelay)
+    except Exception:
+        logger.exception('error delete project')
+def main():
```
drewandersonnz: code flair - please add a single newline between the end of one function and the beginning of another

scripts/monitoring/cron-send-project-operation.py
```diff
@@ -0,0 +1,130 @@
...
+
+    ocutil.namespace = args.namespace
+    project_exists = check_project(args)
+    # project does not exists.
+    create_project_code = 0
+    delete_project_code = 0
+
+    if project_exists == 1:
+        logger.info("project does not exists, going to create it")
+        create_project(args)
```
drewandersonnz, line 108: the exception raised in create_project is captured in the function and hidden from this call .. it means any exceptions will be logged but ignored .. is this what you want to do?

scripts/monitoring/cron-send-project-operation.py
```diff
@@ -0,0 +1,130 @@
...
+            logger.info('project creation failed')
+        else:
+            # project creation succeed, then delete the project
+            delete_project(args)
+            delete_project_code = check_project(args)
+            if delete_project_code == 1:
+                # 1 means project deletion failed
+                logger.info('project deletion failed')
+
+    print create_project_code, delete_project_code
```
drewandersonnz: please don't use print, please use logger

scripts/monitoring/cron-send-project-operation.py
```diff
@@ -0,0 +1,130 @@
...
+        project = runOCcmd("get project {}".format(config.namespace))
+    except Exception:
+        logger.exception('error creating new project')
+
+def delete_project(config):
+    " delete the project "
+    try:
+        runOCcmd("delete project {}".format(config.namespace), base_cmd='oc')
+        time.sleep(commandDelay)
+    except Exception:
```
drewandersonnz, line 89: this exception is handled here, logged, then ignored .. is this what we want?

scripts/monitoring/cron-send-project-operation.py
```diff
@@ -0,0 +1,130 @@
...
+
+    return project_exist
+
+def create_project(config):
+    " create the project "
+    try:
+        runOCcmd("new-project {}".format(config.namespace), base_cmd='oadm')
+        time.sleep(commandDelay)
+        project = runOCcmd("get project {}".format(config.namespace))
+    except Exception:
```
drewandersonnz, line 81: this exception is logged and ignored, is this what we want?

scripts/monitoring/cron-send-project-operation.py
```diff
@@ -0,0 +1,130 @@
...
+    oc_time = time.time()
+    oc_result = ocutil.run_user_cmd(cmd, base_cmd=base_cmd, )
+    logger.info("oc command took %s seconds", str(time.time() - oc_time))
+    return oc_result
+
+def parse_args():
+    """ parse the args from the cli """
+    logger.debug("parse_args()")
+
+    parser = argparse.ArgumentParser(description='OpenShift app create end-to-end test')
```
drewandersonnz: this script does not to app create end-to-end test

scripts/monitoring/cron-send-project-operation.py
drewandersonnz: after logging, I would `raise exception`

scripts/monitoring/cron-send-project-operation.py
drewandersonnz: after logging, I would `raise exception`

drewandersonnz (changes requested):
a few changes @zhiwliu, can you please check?

drewandersonnz (approved):

scripts/monitoring/cron-send-project-operation.py
drewandersonnz: after talking about this, the code which calls this handles the error cases (project exists, not exists)

scripts/monitoring/cron-send-project-operation.py
drewandersonnz: after talking about this, the code which calls this handles the error cases (project exists, not exists)

scripts/monitoring/cron-send-project-operation.py
drewandersonnz: this comment is no longer required

drewandersonnz (approved):
+1

PR #2438, opened 2017-04-25
allow dig ansible lookups

joelddiaz:
trying an ansible filter of lookup('dig'...) fails without python-dns RPM


joelddiaz:
👍 #2437 in stg

PR #2439, opened 2017-04-25
ansible suggests this is how we should run on one host only with serial

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2436

PR #2444, opened 2017-04-26
Ensure only kernel messages considered for tcp out logging

portante:


ansible/roles/openshift_outbound_tcp_logging/files/tcp_out_logging.conf
```diff
@@ -1,5 +1,5 @@
 # Ansible-managed
-if $msg startswith "tcp_out_logging:" then {
+if ($programname == "kernel") and ($msg startswith "tcp_out_logging:") then {
```
joelsmith, line 2: It seems like a container could still make a program called "kernel" and spam us, no? What about this instead?
```
if $syslogfacility-text == 'kern' and $msg startswith 'tcp_out_logging:' then {
```
As I understand it, nothing can spoof the facility with `kern` since anything in userspace that tries will have its facility flipped back to a userspace facility.

joelsmith (changes requested):
Sorry I didn't review sooner. I didn't notice the GH notification that you wanted a review.

PR #2447, opened 2017-04-26
matching the resource limits in zabbix template to the live system

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2446

PR #2449, opened 2017-04-26
Fixed zabbix web route to redirect when it gets traffic to port 80 for local development.

twiest:


twiest:
PR +1d in STG PR #2448

PR #2452, opened 2017-04-26
revendored to 3.5.62

mwoodson:


mwoodson:
@kwoodson approved #2451

PR #2456, opened 2017-04-27
as per bign, we are doing rslave every cluster and no more /var/lib/d…

ivanhorvath:
…ocker mounting


ivanhorvath:
Approved by @joelddiaz in #2455

PR #2470, opened 2017-05-01
Added logging to list of allowed operations to verify-cicd-operation.py

themurph:


themurph:
@jupierce approved #2469

PR #2477, opened 2017-05-02
Revert "Changed base OS level checks to run every minute so that Jeremy Eder's team can use our zabbix data."

twiest:
This reverts commit 3271a0051128141f84cfd26ba68dab03f075628a.


twiest:
This PR was +1d in STG PR #2476

PR #2483, opened 2017-05-03
Change the zabbix history from the default of 90 days to a default of 14 days.

twiest:


twiest:
This PR was +1d in STG PR #2481.

@blrm has approved this to go to PROD during the Summit change freeze.

PR #2486, opened 2017-05-04
[WIP] Initial ansible-in-docker script using Fedora:25

drewandersonnz:


docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -vx
+
+function buildAnsible {
+    dockerTag=oso-ansible:${ansibleVersion}
+
+    bashScript=./oso-ansible-playbook-${ansibleVersion}
+
+    docker build --tag ${dockerTag} - << EndOfDockerfile
+FROM fedora:25
```
twiest: This needs to support building both rhel7 and centos7 containers.

See our other containers for the patter.

docker/ansible-in-docker/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+* build.bash
+* use built scripts in `#!` line in ansible script
+
+test.yaml:
+
+```
+#!./oso-ansible-playbook-2.3.0.0-3
+# - or -
+#!./oso-ansible-playbook-2.2.2.0-1
```
twiest: Love this use case

docker/ansible-in-docker/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+- hosts: my-host
+  remote_user: root
+  gather_facts: false
+
+  tasks:
+
+  - name: "date -u"
+    command: /bin/bash -c "date -u"
```
twiest: Consider having the example be something like this:

```
- name: show the ansible version
  debug: var=ansible_version
```

docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -vx
+
+function buildAnsible {
+    dockerTag=oso-ansible:${ansibleVersion}
```
twiest: This should include the distro it's built for as well (like our other containers).

e.g. `oso-rhel7-ansible` and `oso-centos7-ansible`

docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
...
+EndOfDockerfile
+
+    cat > ${bashScript} << EndOfBashScript
+#!/bin/bash
+#set -vx
+
+sudo docker run --rm -it --privileged \\
+  -v /etc/ansible:/etc/ansible:ro \\
+  -v /usr/share/ansible/inventory/multi_inventory.py:/usr/share/ansible/inventory/multi_inventory.py:ro \\
+  -v /:/host:ro \\
```
twiest: We should do that rslave thing here like we do with the host monitoring container (avoid any EBS mount problems).

docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -vx
+
+function buildAnsible {
+    dockerTag=oso-ansible:${ansibleVersion}
+
+    bashScript=./oso-ansible-playbook-${ansibleVersion}
```
twiest: We need to figure out the best way to get this installed into /usr/bin.  rpm?

docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
...
+
+sudo docker run --rm -it --privileged \\
+  -v /etc/ansible:/etc/ansible:ro \\
+  -v /usr/share/ansible/inventory/multi_inventory.py:/usr/share/ansible/inventory/multi_inventory.py:ro \\
+  -v /:/host:ro \\
+  -v /dev/shm/.ansible:/dev/shm/.ansible:ro \\
+  -v \$SSH_AUTH_SOCK:/tmp/sshauth.sock \\
+  -e SSH_AUTH_SOCK=/tmp/sshauth.sock \\
+  -w /host/\`pwd\` \\
+  ${dockerTag} \\
```
twiest: This should probably be a registry.ops.openshift.com URL so that both tower machines are using the exact same images.

docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
...
+  /usr/bin/ansible-playbook \$@
+EndOfBashScript
+
+    chmod 755 ${bashScript}
+}
+
+export ansibleRPM=https://kojipkgs.fedoraproject.org/packages/ansible/2.2.2.0/1.fc25/noarch/ansible-2.2.2.0-1.fc25.noarch.rpm
```
twiest: Need to change to epel / el7 specific versions.

docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -vx
+
+function buildAnsible {
+    dockerTag=oso-ansible:${ansibleVersion}
+
+    bashScript=./oso-ansible-playbook-${ansibleVersion}
+
+    docker build --tag ${dockerTag} - << EndOfDockerfile
```
twiest: We need to build and push to registry.ops.openshift.com

twiest (changes requested):

docker/ansible-in-docker/README.md
```diff
@@ -0,0 +1,26 @@
+ansible-in-docker
```
twiest: Our pattern for directory names is to have it be the same as the container name.

So in this case `oso-ansible`

docker/ansible-in-docker/build.bash
```diff
@@ -0,0 +1,42 @@
...
+FROM fedora:25
+
+RUN yum -y update \
+ && yum -y install ${ansibleRPM} \
+ && yum clean all
+EndOfDockerfile
+
+    cat > ${bashScript} << EndOfBashScript
+#!/bin/bash
+#set -vx
```
twiest: Add deprecation section to alert the user when the version of ansible being used has been deprecated.

joelsmith:
:+1: from a security stand-point as long as people aren't using versions of Ansible with security issues. Perhaps anyone wishing to use a version of Ansible with security issues should seek security team approval for their specific use case prior to use.

docker/ansible-in-docker/build.bash
drewandersonnz: As the container is short-lived (life-of-script, as opposed to host-monitoring-daemon), I'm not 100% sure this will be a problem. I'll try add it.

docker/ansible-in-docker/build.bash
drewandersonnz: I'm not sure RPM would work. With deprecating containers based on ansible versions it would need to match whatever containers are available. We could create these seperately to the build and do registry scanning and/or fixed (in repo) versions without scanning?

joelsmith:
Depending upon the outcome of the "where should we get Ansible from" discussion from libra-ops (subj: Ansible update for security fix), it might make sense to get the packages from OpenShift repos instead of from EPEL

PR #2492, opened 2017-05-04
Fix zabbix config loop problems.

twiest:
Fix zabbix config loop problems:
* zabbix_config wasn't actually configuring the zabbix_config template.
* two triggers for docker memory usage were named the same thing, causing one to overwrite the other. Depending on timeing, this was also causing zabbix-web to return a 504 on either the first or second attempt to define this trigger. This was meant to be 2 different triggers and as such needs to have 2 different names. Since these are related, there also needed to be a dependency setup. This PR adds those things.


twiest:
This PR was +1d in STG PR #2491.

@blrm has signed off on merging this during the summit change freeze.

twiest:
[test]

PR #2493, opened 2017-05-05
3.5 Upgrade Cherry Picks and Logging fix

dgoodwin:
Main addition is a fix for the installation of latest openshift-scripts-dedicated every installation. Recently the 3.5 version of this package became incompatible for 3.4. Now we enforce use of an openshift-scripts-dedicated that matches the major OpenShift version. 

WARNING: The repos are not yet ready for this, only one version of the package can exist at any time, so every 3.5 build will break new 3.4 installs, every 3.4 build will break 3.5 installs. Have emailed about this but no one has solutions at this time.

Also included are two commits that are cherry picks from openshift-ansible release-1.5 branch, they will be picked up automatically if 3.5 is vendored again. Both fix issues in logging.

### DIFF
--- a/ansible/roles/openshift_dedicated_admin/tasks/main.yml
+++ b/ansible/roles/openshift_dedicated_admin/tasks/main.yml
@@ -1,9 +1,4 @@
 ---
-- name: Install the OpenShift Dedicated Admin scripts
-  yum:
-    name: openshift-scripts-dedicated
-    state: present
-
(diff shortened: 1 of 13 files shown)


yocum137 (approved):

yocum137:
If this breaks everything, can I emigrate to Canada?

dgoodwin:
Never bet against my PRs @yocum137 :)

yocum137 (approved):

PR #2495, opened 2017-05-05
updated the ovs check to fire if not 2 or 4 nodes are there

mwoodson:


mwoodson:
@robotmaxtron and @twiest approved #2494

PR #2499, opened 2017-05-08
zabbix: removed config loop client data from the config loop template (already in config loop client template)

mwoodson:


mwoodson:
@twiest approved #2498

PR #2500, opened 2017-05-08
put more detailed check for whether zabbix is ready to be configured

joelddiaz:
attempt to access the web UI results in HTTP response '200' even when mysql isn't yet configured.

keep accessing the web UI until the HTML output includes the text 'sign in as guest' (which is some of the text you would find when the web UI was able to successfully talk to the mysql backend


joelddiaz:
👍 #2487 in stg

PR #2503, opened 2017-05-08
add some defaults for httptests

joelddiaz:
so that we can set up local zabbix without having to comment these peices out


joelddiaz:
👍 #2502 in stg

PR #2505, opened 2017-05-08
remove assumption about only 2 pods

joelddiaz:
check whether every pod (not just the first two pods) is running on a separate node


joelddiaz:
👍 #2504 in stg

PR #2509, opened 2017-05-10
Add ability to exclude pkgs from EPEL repos

joelsmith:


joelsmith:
@mwoodson gave 👍 in STG pr #2508

PR #2510, opened 2017-05-11
Push back metrics check by one minute

mwringe:
This should hopefully help with the intermittent test failures we are seeing here which pass when the tests are rerun.

We could be running into a situation where the timing of the Heapster gathering period, along with the time it takes to process and store metrics into Hawkular is causing this test to fail.


mwringe:
@wshearn what needs to be done here?

wshearn:
ok to test

wshearn:
[test]

wshearn:
Ok, so what is happening is you need to merge to stage first, then you can merge to prod.

mwringe:
OK, the stg PR is available here: https://github.com/openshift/openshift-tools/pull/2565

Lets wait and see what the results are from that one before proceeding with this PR.

PR #2516, opened 2017-05-16
track global_exception, introduce and use logging, use logging exceptions

drewandersonnz:
+1 from @BlueShells on https://github.com/openshift/openshift-tools/pull/2515


gburges:
ok to test

PR #2519, opened 2017-05-16
add zabbix key for cron-send-project-operation.py

zhiwliu:


zhiwliu:
https://github.com/openshift/openshift-tools/pull/2518

PR #2520, opened 2017-05-16
removed the debug for deploying the zagg service

mwoodson:


mwoodson:
@ivanhorvath approves

PR #2525, opened 2017-05-16
fixed oc_Version to have results instead of result. This is upstream compatible now

mwoodson:


mwoodson:
@ivanhorvath approved #2524

PR #2529, opened 2017-05-16
update all links to ops-sop

blentz:


blentz:
:+1: from @gburges in #2523

PR #2533, opened 2017-05-17
upped the host monitoring service timeout to 10 minutes

mwoodson:


mwoodson:
@jupierce approved #2532

PR #2539, opened 2017-05-19
Adding necessary config to run and report rkhunter scans from host mo…

rhdedgar:
…nitoring container

### DIFF
--- a/docker/oso-host-monitoring/centos7/root/config.yml
+++ b/docker/oso-host-monitoring/centos7/root/config.yml
@@ -11,6 +11,17 @@
     - /var/lib/yum
     - /var/cache/yum
 
+    rkhunter_cron:
+    - name: check_rkhunter
+      minute: "30"
+      hour: 21
+      job: "/bin/cron-send-rkhunter-checks"
+
+    - name: rkhunter_scan
+      minute: "0"
+      hour: "21"
+      job: "chroot /var/local/rkhunter_chroot /tmp/rkhunter/bin/rkhunter --sk --check --configfile /tmp/rkhunter/etc/rkhunter.conf --dbdir /tmp/rkhunter/db --tmpdir /tmp/rkhunter >/dev/null 2>&1"
+
   pre_tasks:
   - stat:
       path: /secrets/host_monitoring_vars.yml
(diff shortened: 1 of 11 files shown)


docker/oso-host-monitoring/centos7/root/config.yml
```diff
@@ -169,3 +182,54 @@
...
+    with_items:
+    - /var/local/rkhunter_chroot
+    - /var/local/rkhunter_tmp
+    - /var/local/rkhunter_tmp/rkhunter
+    - /var/local/rkhunter_tmp/rkhunter/bin
+    - /var/local/rkhunter_tmp/rkhunter/db
+    - /var/local/rkhunter_tmp/rkhunter/etc
+    - /var/local/rkhunter_tmp/rkhunter/scripts
+
+  # Running as commands because mount module doesn't seem to recognize rbind
```
kwoodson: Maybe only run the ones that aren't supported under command.

scripts/monitoring/cron-send-rkhunter-checks.py
```diff
@@ -0,0 +1,77 @@
...
+    @staticmethod
+    def check_rkhunter(log_message, logfile):
+        """ Check number of occurrences of the provided string in the specified logfile. """
+
+        total_issues = 0
+
+        if os.path.isfile(logfile):
+            with open(logfile) as open_file:
+                stripped_line = list([line.rstrip() for line in open_file.readlines()])
+                for line in stripped_line:
```
kwoodson: Looks like a double loop in your code here.  First loop walks the lines and strips and then directly after you proceed to loop again.  Is this the case?  I'd prefer to walk the loop once if we can.

kwoodson (dismissed):
Few questions here.  Please respond but the overall idea looks sound.  I didn't not test or syntax check.

scripts/monitoring/cron-send-rkhunter-checks.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @staticmethod
+    def check_rkhunter(log_message, logfile):
+        """ Check number of occurrences of the provided string in the specified logfile. """
+
+        total_issues = 0
+
+        if os.path.isfile(logfile):
+            with open(logfile) as open_file:
+                stripped_line = list([line.rstrip() for line in open_file.readlines()])
```
kwoodson: the `[ x for x in z ]` is a list comprehension that produces a list.  No need to cast to a list.

kwoodson (approved):

PR #2541, opened 2017-05-22
Updating cron-send-usage-pv to support SI units

blrm:
This adds unit handling to this check so that the proper usage values are reported. This will fix some of the ops-runner alerts for this check that are currently on the zabbix board.

### DIFF
--- a/scripts/monitoring/cron-send-usage-pv.py
+++ b/scripts/monitoring/cron-send-usage-pv.py
@@ -18,6 +18,7 @@
 
 import argparse
 import datetime
+import logging
 import time
 import re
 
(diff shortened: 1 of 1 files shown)


blrm:
[test]

blentz:
:-1: 

Rather than doing unit coversions in the script, I'd prefer to see the raw number submitted to zabbix and use the item's Units, Custom Multiplier, or Store Value As configuration settings to tell Zabbix how it should be transformed, whichever is appropriate.

blentz:
Chatted with @blrm and he made a convincing argument. So, :+1: for the change given that we're responding to unfortunate behavior from the underlying call we're making to collect data.

PR #2542, opened 2017-05-22
moving FORWARD logic to OPENSHIFT-OUTPUT-FILTERING

rhdedgar:


ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -43,42 +43,48 @@
...
   command: '/usr/sbin/iptables -w -nL oo-outbound-logging'
   changed_when: False
   register: loggingchain
   ignore_errors: True
 
-- name: add in-memory iptables chain
+- name: add in-memory filtering chain
+  command: "/usr/sbin/iptables -N OPENSHIFT-OUTBOUND-FILTERING"
+  when: not "OPENSHIFT-OUTBOUND-FILTERING" in filteringchain.stdout
+  with_items: "{{ outbound_logging_chain }}"
```
joelsmith: This `with_items` seems vestigial. You never refer to {{ item }} in the rule, so I'm guessing you don't want the `with_items`.

ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -43,42 +43,48 @@
...
+
+- name: add in-memory logging chain
   command: "/usr/sbin/iptables -N oo-outbound-logging"
   when: not "oo-outbound-logging" in loggingchain.stdout
   with_items: "{{ outbound_logging_chain }}"
 
 - name: modify in-memory iptables rules
   command: "/usr/sbin/iptables {{ item.line }}"
-  when: not item.regex in forwardchain.stdout and not item.regex in loggingchain.stdout
+  when: not item.regex in filteringchain.stdout and not item.regex in loggingchain.stdout
```
joelsmith: even though this works, I don't like the idea of a regex being treated as a string. What if somebody comes along later and updates the "regex" which only supports a fixed-string? I suggest something like this instead:

`when: not (filteringchain.stdout | search(item.regex)) and not (loggingchain.stdout | search(item.regex))`

joelsmith (changes requested):
Mostly good... a couple of code cleanliness issues.

joelsmith (approved):
👍

ansible/roles/openshift_outbound_tcp_logging/tasks/main.yml
```diff
@@ -1,5 +1,11 @@
 ---
 - set_fact:
+    outbound_filtering_chain:
+    - chain: OPENSHIFT-OUTPUT-FILTERING
+      line:  ":OPENSHIFT-OUTPUT-FILTERING - [0:0]"
+      after: "filter$"
```
joelsmith: I'd prefer something that more tightly matches so that we're not matching something from another table that happens to end with filter. Maybe something like:

`after: "^\\*filter$"`

joelsmith (changes requested):
Sorry about the OUTBOUND/OUTPUT thing. Just one more minor regex tightening and I think it'll be good.

joelsmith (approved):
👍 assuming it works when tested. Thanks for making these fixes!

PR #2543, opened 2017-05-22
Udp outbound rules

rhdedgar:


ansible/roles/openshift_udp_limits/files/iptables_udp.conf
```diff
@@ -0,0 +1,5 @@
+# Ansible-managed
+:msg, startswith, "UDP:" -/var/log/iptables.log
```
joelsmith: So as @portante pointed out with the TCP logging, with the way you have it we'd grab any log message that started with the prefix. What we really want is only to grab kernel log messages that start with the prefix, so something like this:
```
if $syslogfacility-text == 'kern' and $msg startswith 'UDP: ' then -/var/log/iptables.log
& stop
```

Also, we should only need that one stanza, not the one that regex one that matches the numbers and stuff. I think that one came from me being lazy in V2 and not bothering to figure out which flavor I needed.

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+      {%- for item in hostsubnets['results']['results'][0]['items'] -%}
+       {%- if item.host == kubernetes_node_name -%}
+        {{ item.subnet }}
+       {%- endif -%}
+      {%- endfor -%}
+
+- set_fact:
+    oo_udp_chain:
+    - chain: oo-udp
+      after: "filter$"
```
joelsmith: Like the other PR:
`"^\\*filter$"`

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+      regex: :oo-udp - \[0:0\]
+      block:  ":oo-udp - [0:0]"
+
+    oo_udp_limit_chain:
+    - chain: oo-udp-limit
+      after: "filter$"
```
joelsmith: Like the other PR:
`"^\\*filter$"`

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+      regex: :oo-udp-limit - \[0:0\]
+      block:  ":oo-udp-limit - [0:0]"
+
+    filtering_udp_rules:
+    # allow all outbound UDP for privileged pods
+    # send all other traffic to the oo-udp chain so that rules that allow
+    # traffic can call -j RETURN instead of -j ACCEPT, and thus permit us
+    # to do later accounting
+    - chain: OPENSHIFT-OUTBOUND-FILTERING
```
joelsmith: Reminder, replace all `OPENSHIFT-OUTBOUND-FILTERING` with `OPENSHIFT-OUTPUT-FILTERING`. Sorry 'bout that.

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+      after: "OPENSHIFT-OUTBOUND-FILTERING"
+      regex: "id_outbound_udp_filtering_2_"
+      line: "-I OPENSHIFT-OUTBOUND-FILTERING -m comment --comment id_outbound_udp_filtering_2_ -s {{ node_cidr }} -o eth0 -p udp -j oo-udp"
+
+    - chain: OPENSHIFT-OUTBOUND-FILTERING
+      after: "id_outbound_udp_filtering_2_"
+      regex: "id_outbound_udp_filtering_1_"
+      line: "-I OPENSHIFT-OUTBOUND-FILTERING -m comment --comment id_outbound_udp_filtering_1_ -s 172.0.0.0/8 -o eth0 -p udp -j ACCEPT"
```
joelsmith: I believe that this rule is no longer needed due to the interface checking in the jump rule in the FORWARD chain.

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+    # Limit any remaining permitted UDP traffic to 30 datagrams per second,
+    # and define permitted UDP traffic for un-privileged pods
+    # Finally, drop all other UDP traffic. Log dropped traffic up to once per minute.
+    # Because all dropped traffic is crammed into this single limited log rule,
+    # log entries will only be a sampling of actual dropped traffic.
+    - chain: oo-udp
+      after: "id_outbound_udp_filtering"
+      regex: "id_outbound_udp_rule_1_"
+      block: |-
+        -A oo-udp -d 172.16.0.0/12 -o eth0 -p udp -m udp --dport 53 -j RETURN
```
joelsmith: Looks like we also need rules for each of the reserved private address spaces, so:
```
-A oo-udp -d 10.0.0.0/8 -o eth0 -p udp -m udp --dport 53 -j RETURN
-A oo-udp -d 192.168.0.0/16 -o eth0 -p udp -m udp --dport 53 -j RETURN
```

joelsmith (changes requested):
I'm not finished with my review yet, but here's the first batch of feedback comments.

ansible/roles/openshift_udp_limits/files/iptables_udp.conf
portante: Is there a way to get a more unique prefix via the iptables configuration?  Seems like lots of other logs might begin with `UDP:` as well.

joelsmith: @portante are you concerned about other kernel messages starting with `UDP: `? I looked and our current kernel doesn't have any log messages that begin with `UDP: `. Even so, if we think it's a real concern, we could do `UDP outlog: ` or something like that. What do you think @rhdedgar?

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+  run_once: true
+
+- oc_obj:
+    state: list
+    kind: hostsubnet
+  register: hostsubnets
+  delegate_to: "{{ cluster_master }}"
+  run_once: true
+
+- set_fact:
```
joelsmith: I recommend that you combine this and the following 2 `set_fact` tasks into a single `set_fact` task... makes things go a little faster (but not the above `set_fact` since it's a `run_once` task).

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+    - chain: oo-udp
+      after: "id_outbound_udp_filtering"
+      regex: "id_outbound_udp_rule_1_"
+      block: |-
+        -A oo-udp -d 172.16.0.0/12 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 204.13.250.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 204.13.251.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 208.78.70.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 208.78.71.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -s 10.0.0.0/8 -o eth0 -p udp -j oo-udp-limit
```
joelsmith: I would rather we make this dynamic so it supports clusters with different network configs, so instead of `10.0.0.0/8`, it would say `{{ node_cidr }}`, which I think is what we really want (packets coming from this node's pod, only).

ansible/roles/openshift_udp_limits/tasks/main.yml
joelsmith: Nevermind on this, I think it would be better if we have a single rule that allows DNS traffic to the service network, like this:

First, above, you'd fetch the clusternetworks config:
```
  - oc_obj:
      state: list
      kind: clusternetworks
      name: default
    register: clusternetworks
    delegate_to: "{{ cluster_master }}"
    run_once: true
```
then you'd set a fact for the service network:
```
  - set_fact:
      servicenetwork: "{{clusternetworks.results.results[0].serviceNetwork}}"
```
You could combine that new fact with the other `set_fact` task if you want.

Then, your rule here would use that instead of `172.16.0.0/12`:
```
-A oo-udp -d {{servicenetwork}} -o eth0 -p udp -m udp --dport 53 -j RETURN
```

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+    # allow all outbound UDP for privileged pods
+    # send all other traffic to the oo-udp chain so that rules that allow
+    # traffic can call -j RETURN instead of -j ACCEPT, and thus permit us
+    # to do later accounting
+    - chain: OPENSHIFT-OUTPUT-FILTERING
+      after: "OPENSHIFT-OUTPUT-FILTERING"
+      regex: "id_outbound_udp_filtering_2_"
+      line: "-I OPENSHIFT-OUTPUT-FILTERING -m comment --comment id_outbound_udp_filtering_2_ -s {{ node_cidr }} -o eth0 -p udp -j oo-udp"
+
+    - chain: OPENSHIFT-OUTPUT-FILTERING
```
joelsmith: Thanks to the above rule being specific enough (with `{{ node_cidr }}`, we shouldn't need this rule at all anymore.

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+      after: "id_outbound_udp_filtering_2_"
+      regex: "id_outbound_udp_filtering_1_"
+      line: "-I OPENSHIFT-OUTPUT-FILTERING -m comment --comment id_outbound_udp_filtering_1_ -s 172.0.0.0/8 -o eth0 -p udp -j ACCEPT"
+
+    filtering_udp_memory_rules:
+    - chain: OPENSHIFT-OUTPUT-FILTERING
+      regex: "id_outbound_udp_filtering_2_"
+      line: "-I OPENSHIFT-OUTPUT-FILTERING -m comment --comment id_outbound_udp_filtering_2_ -s {{ node_cidr }} -o eth0 -p udp -j oo-udp"
+
+    - chain: OPENSHIFT-OUTPUT-FILTERING
```
joelsmith: Thanks to the above rule being specific enough (with `{{ node_cidr }}`, we shouldn't need this rule at all anymore.

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
+---
+- name: Choose a master on which to run
+  set_fact:
+    cluster_master: "{{ groups['oo_hosttype_master'] | intersect(groups['oo_clusterid_' ~ oo_clusterid]) | first }}"
+  run_once: true
+
+- oc_obj:
```
joelsmith: It would be nice to have a `name` here. Maybe something like "Get the host's pod network subnet".

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+      block:  ":oo-udp - [0:0]"
+
+    oo_udp_limit_chain:
+    - chain: oo-udp-limit
+      after: "filter$"
+      regex: :oo-udp-limit - \[0:0\]
+      block:  ":oo-udp-limit - [0:0]"
+
+    filtering_udp_rules:
+    # allow all outbound UDP for privileged pods
```
joelsmith: This comment is a little out of date now that the positioning of the chains has been shuffled around. Perhaps something like:
```
# Send network-namespaced pod traffic to the oo-udp chain.
```
I don't think you need any of the rest of this comment.

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+    # log entries will only be a sampling of actual dropped traffic.
+    - chain: oo-udp
+      after: "id_outbound_udp_filtering"
+      regex: "id_outbound_udp_rule_1_"
+      block: |-
+        -A oo-udp -d 172.16.0.0/12 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 204.13.250.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 204.13.251.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 208.78.70.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
+        -A oo-udp -d 208.78.71.23/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
```
joelsmith: I'm not sure that we need to specially whitelist dyn DNS servers here. I think it's fine to remove these 4 lines since by default, pods use our in-cluster DNS servers.

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+        -A oo-udp -s 10.0.0.0/8 -o eth0 -p udp -j oo-udp-limit
+        -A oo-udp -o eth0 -p udp -m udp --dport 123 -j RETURN
+        -A oo-udp -o eth0 -p udp -m udp --dport 67 -j RETURN
+        -A oo-udp -o eth0 -p udp -m udp --dport 161 -j RETURN
+        -A oo-udp -o eth0 -p udp -m udp --dport 500 -j RETURN
+        -A oo-udp -o eth0 -p udp -m udp --dport 1514 -j RETURN
+        -A oo-udp -o eth0 -p udp -m udp --dport 4500 -j RETURN
+        -A oo-udp -o eth0 -p udp -m udp --dport 25826 -j RETURN
```
joelsmith, line 81: I think it might be a good idea to allow some minimal DNS traffic to other DNS servers. The local DNS traffic isn't throttled, but this would be, so I don't think it's dangerous:
```
-A oo-udp -o eth0 -p udp -m udp --dport 53 -j RETURN
```

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+    name: oo-udp
+    rules: "{{ (outbound_udp_rules|map(attribute='block')|join('')).splitlines() }}"
+
+- name: List iptables rules filtering chain
+  command: '/usr/sbin/iptables -w -nL OPENSHIFT-OUTPUT-FILTERING'
+  changed_when: False
+  register: filtering_chain_check
+
+- name: Modify in-memory filtering chain
+  command: "/usr/sbin/iptables {{ item.line }}"
```
joelsmith: should probably be `"/usr/sbin/iptables -w {{ item.line }}"` so that it'll wait for the lock if something else has it.

joelsmith (changes requested):
Changes, round 2!

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -4,63 +4,60 @@
...
     filtering_udp_rules:
-    # allow all outbound UDP for privileged pods
-    # send all other traffic to the oo-udp chain so that rules that allow
-    # traffic can call -j RETURN instead of -j ACCEPT, and thus permit us
-    # to do later accounting
+    # Send network-namespaced pod traffic to the oo-udp chain. 
     - chain: OPENSHIFT-OUTPUT-FILTERING
       after: "OPENSHIFT-OUTPUT-FILTERING"
-      regex: "id_outbound_udp_filtering_2_"
-      line: "-I OPENSHIFT-OUTPUT-FILTERING -m comment --comment id_outbound_udp_filtering_2_ -s {{ node_cidr }} -o eth0 -p udp -j oo-udp"
```
joelsmith: looks like you got rid of the wrong one. We want to keep the -j oo-udp rule and lose the -j ACCEPT rule. Sorry if I made that confusing.

ansible/roles/openshift_udp_limits/files/iptables_udp.conf
```diff
@@ -1,5 +1,3 @@
 # Ansible-managed
-:msg, startswith, "UDP:" -/var/log/iptables.log
-& stop
-:msg, regex, "^\[ *[0-9]*\.[0-9]*\] UDP:" -/var/log/iptables.log
+:msg, startswith, "UDP outlog:" -/var/log/iptables.log
```
joelsmith: If we switch from the Docker json log driver to the journald log driver (but keep rsyslog), this setup would allow containers to spam our log file by starting their message with `UDP outlog:`. That's why last time I suggested that you change to an expression that would also filter on whether the message came via the kernel log facility, which would make it impossible for a container to spam the log file. If you get rid of this `:msg` line and replace it with this, it should do the trick:
`if $syslogfacility-text == 'kern' and $msg startswith 'UDP outlog: ' then -/var/log/iptables.log`

joelsmith (changes requested):
Getting close

ansible/roles/openshift_udp_limits/files/iptables_udp.conf
```diff
@@ -0,0 +1,2 @@
+# Ansible-managed
+if $syslogfacility-text == 'kern' and $msg startswith 'UDP outlog: ' then -/var/log/iptables.log
```
joelsmith, line 2: we still want the `& stop` or the logs will end up in `/var/log/messages` as well as `/var/log/iptables.log`.

joelsmith (changes requested):
So close!

joelsmith (approved):
Looks good, 👍 after testing proves out

PR #2553, opened 2017-05-23
Prevent forward chain band-aid from flushing the jump chain every minute

joelsmith:


joelsmith:
Approved by @rhdedgar in stg PR #2552

PR #2556, opened 2017-05-24
Re-add mistakenly-removed section to openshift_outbound_tcp_logging

joelsmith:
Also make sure we wait for iptables locks when adding rules

This is to fix a bug introduced in #2554 that broke the config loop.


BlueShells:
:+1:

PR #2562, opened 2017-05-25
Patching comparison operation breaking with iptables lock message is …

rhdedgar:
…present in output


joelsmith:
This breaks the logic by not calling .get() twice. The right place to filter it if we wanted to is in `get()`, but we don't want to. The correct fix is to uninstall the version of iptables that @eparis installed on our nodes that prints this garbage message into iptables's stdout. If he really needs that message, then he needs to rebuild it to write to stderr instead of stdout.

PR #2570, opened 2017-05-30
Adding verbose output to metrics check to assist in debugging

blrm:


scripts/monitoring/cron-send-metrics-checks.py
```diff
@@ -179,9 +179,13 @@ def check_node_metrics(self):
...
                 res = yaml.load(resp.read())
                 if res[0]['empty']:
+                    if self.args.verbose:
+                        print "WARN - Node not reporting metrics: %s" % item['metadata']['name']
                     result = 0
 
-            except urllib2.URLError:
+            except urllib2.URLError as e:
+                if self.args.verbose:
+                    print "WARN - Request returned HTTP %s : %s" % (e.code, e.url)
```
wshearn: Since this is an exception, wouldn't this be better served as ERROR instead of WARN?

wshearn (changes requested):

scripts/monitoring/cron-send-metrics-checks.py
blrm: Yes, this makes more sense as an ERROR. changed

wshearn (approved):

PR #2571, opened 2017-05-31
cron changes based on arch meeting discussions

ivanhorvath:


ivanhorvath:
[test]

ansible/roles/os_zabbix/vars/template_os_linux.yml
```diff
@@ -337,3 +337,8 @@ g_template_os_linux:
     url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/check_yum_update.asciidoc"
     priority: info
     description: "One or more security updates need to be applied to the host"
+
+  - name: "pcp items send failed on {HOST.NAME}"
+    expression: "{Template OS Linux:kernel.uname.sysname.max(#3)}<1"
```
twiest: I believe uname.sysname is a string, therefore a numeric operation (`max`) won't work on it.

Consider using `nodata` instead.

Also, make sure you test this to make sure it's having the desired effect.

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -356,10 +345,10 @@ host_monitoring_cron:
   job: ops-runner -f -s 300 -n ozc.send.synthetic.heartbeat.full /usr/bin/ops-metric-client --send-heartbeat --synthetic
 
 - name: Do a quick heartbeat for synthetic host
-  minute: "*"
+  minute: "*/2"
```
twiest: This should be in sync with the normal heartbeat, which is set to every 3 minutes.

Sorry, I probably screwed up and said 2 here.

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -424,7 +413,7 @@ host_monitoring_cron:
 
 - name: "Check EBS stuck volume attachment state"
   minute: "*/2"
-  job: "ops-runner -f -s 15 -n cseeviss.ebs.stuck.state cron-send-ec2-ebs-volumes-in-stuck-state --region {{ osohm_region }} --stuck-after 120 --aws-creds-profile snapshotter --verbose"
+  job: "ops-runner -f -s 15 -n cseeviss.ebs.stuck.state cron-send-ec2-ebs-volumes-in-stuck-state --region {{ osohm_region }} --stuck-after 120 --aws-creds-profile snapshotter --verbose"- name: "Check EBS stuck volume attachment state"
```
twiest: This change looks wrong. Was this intentional?

twiest (changes requested):
Looking good other than my comments.

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
ivanhorvath: Ahh yes, this is the result of tower not responding while clicking around the screen, probably selected and copied stuff. Good catch.

ansible/roles/os_zabbix/vars/template_os_linux.yml
ivanhorvath: Good call, tested it in local dev, it works with nodata.

ansible/roles/os_zabbix/vars/template_os_linux.yml
```diff
@@ -337,3 +337,8 @@ g_template_os_linux:
     url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/check_yum_update.asciidoc"
     priority: info
     description: "One or more security updates need to be applied to the host"
+
+  - name: "pcp items send failed on {HOST.NAME}"
+    expression: "{Template OS Linux:kernel.uname.sysname.nodata(15m)}=1"
+    url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/check_pcp_ping.asciidoc"
+    priority: average
```
twiest: Ok, I spent a lot of time last night thinking about this.

For posterity, here are the tradeoffs:
Good:
* this is reusing an item we're already sending

Bad:
* nodata is going to cause this trigger to fire for all cases where the heartbeat trigger will fire (doubling them up)
* nodata is more expensive than max


Given the above tradeoffs, I think I prefer keeping it the way it was, so I suggest getting rid of these changes.

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -110,15 +110,12 @@ metric_sender_config:
 
 # Generic Linux Checks
 host_monitoring_cron:
-- name: send pcp ping every 5 mintues
-  minute: "*/5"
-  job: ops-runner -f -s 60 -n cspp.pcp.ping /usr/bin/cron-send-pcp-ping
```
twiest: Keep this section (i.e. undo this change), but have it to run every 10 minutes.

twiest (changes requested):
This is very close to ready.

ansible/roles/os_zabbix/vars/template_os_linux.yml
ivanhorvath: Changed this back.

twiest (approved):
Looks amazing!

PR #2574, opened 2017-05-31
fixing check on removed files

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2573

PR #115, opened 2015-09-15
Adding SSL to v3 monitoring

kwoodson:

PR #2582, opened 2017-06-03
Touch SSO container's reconfig touchfile via ansible

joelsmith:
Under high load situations, "ansible-playbook" was exiting with a 0 return
status (true) despite its never having run any of the tasks in the
playbook. By moving the
"I'm done!" touchfile to the Ansible playbook, we can make sure that the
playbook ran to completion.

### DIFF
--- a/docker/oso-saml-sso/centos7/root/config.yml
+++ b/docker/oso-saml-sso/centos7/root/config.yml
@@ -105,3 +105,11 @@
         content: "{{ saml2_sso_configdata.saml2_idp_cert_data.cert }}"
       - name: "{{ saml2_sso_configdata.saml_privatekey_file }}"
         content: "{{ saml2_sso_configdata.saml2_idp_cert_data.key }}"
+
+  # NOTE: this must be the final task because it indicates to
+  #       the start-up script that this playbook successfully
+  #       ran to completion.
+  - name: touch file to indicate that the reconfigure was successful
+    copy:
+      content: ""
+      dest: "{{ config_success_touchfile }}"
--- a/docker/oso-saml-sso/centos7/start.sh
+++ b/docker/oso-saml-sso/centos7/start.sh
@@ -1,5 +1,13 @@
 #!/bin/bash -e
 
+mypid=$$
+
+if [ $mypid -eq 1 ]; then
+  # allow somebody to "kill" us even though we're running as pid 1 in a Docker container
+  trap "echo 'PID 1 received SIGTERM (from liveness or readiness probe?), exiting!'; exit 1" SIGTERM
+  trap "echo 'PID 1 received SIGINT (failed reconfigure?), exiting!'; exit 1" SIGINT
+fi
+
 # This is useful so we can debug containers running inside of OpenShift that are
 # failing to start properly.
 if [ "$OO_PAUSE_ON_START" = "true" ] ; then
@@ -13,6 +21,7 @@ fi
 
 echo user:x:$(id -u):0:USER:/root:/bin/bash >> /etc/passwd
 echo group:x:$(id -G | awk '{print $2}'):user >> /etc/group
+reconf_touchfile="/configdata/reconfigure_successful"
 
 # This backgrounded block will run in a loop forever.
 # Its job is to monitor for secrets changes and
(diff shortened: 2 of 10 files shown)


docker/oso-saml-sso/centos7/root/config.yml
```diff
@@ -105,3 +105,11 @@
         content: "{{ saml2_sso_configdata.saml2_idp_cert_data.cert }}"
       - name: "{{ saml2_sso_configdata.saml_privatekey_file }}"
         content: "{{ saml2_sso_configdata.saml2_idp_cert_data.key }}"
+
+  # NOTE: this must be the final task because it indicates to
+  #       the start-up script that this playbook successfully
+  #       ran to completion.
```
dbaker-rh: Is it worth an additional comment to say "ansible exits with successful return code even if it times out"?

The code is clear in what it does, but it takes institutional knowledge to know WHY it's here.

docker/oso-saml-sso/centos7/start.sh
```diff
@@ -13,6 +13,7 @@ fi
 
 echo user:x:$(id -u):0:USER:/root:/bin/bash >> /etc/passwd
 echo group:x:$(id -G | awk '{print $2}'):user >> /etc/group
+reconf_touchfile="/configdata/reconfigure_successful"
```
dbaker-rh: You have hard coded values in two places.  The /path/file is here and also above in docker/oso-saml-sso/centos7/root/config.yml line #115.

Can this be parameterized across files?

docker/oso-saml-sso/centos7/start.sh
```diff
@@ -23,7 +24,9 @@ echo group:x:$(id -G | awk '{print $2}'):user >> /etc/group
     config_version_dir=$(readlink -f /secrets/..data)
     echo "Running config playbook"
     for attempt_number in {1..3}; do
-      if ansible-playbook /root/config.yml; then
+      rm -f "$reconf_touchfile"
+      # if the playbook is successful, it will touch $reconf_touchfile as its final task
```
dbaker-rh: Suggestion: comment should indicate that ansible-playbook returns success even if it fails.

docker/oso-saml-sso/centos7/start.sh
```diff
@@ -23,7 +24,9 @@ echo group:x:$(id -G | awk '{print $2}'):user >> /etc/group
...
+      if ansible-playbook /root/config.yml && [ -f "$reconf_touchfile" ]; then
         break
       else
         echo "Pod configuration attempt #$attempt_number failed"
```
dbaker-rh: This "else" doesn't get run on the new failure mode.  Should it?

docker/oso-saml-sso/src/start.sh
```diff
@@ -23,7 +24,9 @@ echo group:x:$(id -G | awk '{print $2}'):user >> /etc/group
     config_version_dir=$(readlink -f /secrets/..data)
     echo "Running config playbook"
     for attempt_number in {1..3}; do
-      if ansible-playbook /root/config.yml; then
+      rm -f "$reconf_touchfile"
+      # if the playbook is successful, it will touch $reconf_touchfile as its final task
+      if ansible-playbook /root/config.yml && [ -f "$reconf_touchfile" ]; then
```
dbaker-arch: suggest:  # ansible-playbook will return 0 on some error conditions (Such as timeout), so we cannot use it's return code to determine success.

ansible-playbook /root/config.yml
if [ -f "$reconf_touchfile" ]; then
  ....

docker/oso-saml-sso/src/start.sh
```diff
@@ -13,6 +13,7 @@ fi
 
 echo user:x:$(id -u):0:USER:/root:/bin/bash >> /etc/passwd
 echo group:x:$(id -G | awk '{print $2}'):user >> /etc/group
+reconf_touchfile="/configdata/reconfigure_successful"
```
dbaker-arch, line 24: Duplicate hard-coded data  -- can it be parameterized across files?

dbaker-arch (changes requested):
Would like to discuss my comments briefly before approving the PR.

docker/oso-saml-sso/src/start.sh
joelsmith: The && requires that both Ansible return success and the touchfile exist. So if it returns false but somehow the touchfile lands, it would be counted a failure (as currently written).

docker/oso-saml-sso/src/start.sh
joelsmith: Probably better to just have the bash script pass in the file name and have the ansible playbook use the passed in one, if provided, and its configured default, if not.

dbaker-arch (approved):
+1 -- I like the new changes.  Thanks.

PR #2584, opened 2017-06-06
added the skip_if_unavailable option to our epel repo

mwoodson:


mwoodson:
@twiest approved #2583

PR #2586, opened 2017-06-06
Add module for managing the "exclude" line of a yum repo config

joelsmith:
Also use it to set epel.exclude=ansible


mwoodson:
:+1:

PR #2589, opened 2017-06-06
fixed epel yum exclude dependency

mwoodson:


mwoodson:
@ivanhorvath approved #2588

PR #2591, opened 2017-06-06
Fix faux-HTTP client of SSO SSH endpoint to send \r\n as per spec

joelsmith:
The new Apache httpd requires \r\n as a line ending where it
previously allowed \n.


joelsmith:
@robotmaxtron gave 👍 in stg PR #2590

PR #2592, opened 2017-06-06
Touch SSO container's reconfig touchfile via ansible

joelsmith:
Under high load situations, "ansible-playbook" was exiting with a 0 return
status (true) despite its never having run any of the tasks in the
playbook. By moving the
"I'm done!" touchfile to the Ansible playbook, we can make sure that the
playbook ran to completion.

Also, add a signal handler for SIGTERM and SIGINT since it seems that our
error path didn't work well. It appears that the SIGKILL was not delivered
to PID 1 when there were reconfigure errors. (Dockerism?)


joelsmith:
@dbaker-rh approved in stg PR #2582

PR #2604, opened 2017-06-08
Limit ssh proxy connections to bastion hosts.

dbaker-arch:


dbaker-arch:
@joelsmith approved stg PR #2603

PR #2606, opened 2017-06-08
Add generated versions of ssl.conf.j2

dbaker-arch:


dbaker-arch:
@joelsmith approved stg PR #2605

PR #2608, opened 2017-06-09
WIP - tech debt logging addition

rhdedgar:


ansible/roles/openshift_logging/templates/logging-deployer.yaml.j2
```diff
@@ -11,3 +11,4 @@ data:
   kibana-hostname: logs.{{ osalog_clusterid }}.openshift.com
   public-master-url: https://api.{{ osalog_clusterid  }}.openshift.com
   master-url: https://internal.api.{{ osalog_clusterid  }}.openshift.com
+  use-journal: use-journal=true
```
portante, line 14: This won't force the use of `journald` for the actual containers, this just tells `fluentd` to ONLY collect using `journald`, which means when docker is configured to use `json-file` driver you won't get any logs.

ansible/roles/openshift_logging/defaults/main.yml
```diff
@@ -1,5 +1,7 @@
 osalog_fluentd_nodes:
 - type=compute
+- type=master
```
portante, line 3: Shouldn't we have this by default with the openshift-ansible playbook/roles for installing logging?

ansible/roles/openshift_logging/templates/logging-fluentd.yaml.j2
```diff
@@ -29,7 +28,6 @@ data:
       @include configs.d/openshift/filter-kibana-transform.conf
       @include configs.d/openshift/filter-k8s-flatten-hash.conf
```
portante, line 29: This file is totally out of date, and should dropped for the openshift-ansible method of installing logging.

portante (changes requested):

PR #2609, opened 2017-06-09
Fix DNS whitelist destination, clean up node name

joelsmith:


ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -12,18 +12,8 @@
...
-    kind: clusternetworks
-    name: default
-  register: clusternetworks
-  delegate_to: "{{ cluster_master }}"
-  run_once: true
-
 - set_fact:
-    kubernetes_node_name: "{{ (hostvars[inventory_hostname].ec2_private_dns_name if hostvars[inventory_hostname].oo_location == 'aws' else inventory_hostname) }}"
-    servicenetwork: "{{ clusternetworks.results.results[0].serviceNetwork }}"
+    kubernetes_node_name: "{{ ec2_private_dns_name if oo_location == 'aws' else inventory_hostname }}"
```
joelsmith: @twiest could you review? I think it was probably a mistake for me to have us to derive `kubernetes_node_name` from `oo_location` and `ec2_private_dns_name` here in the role. Should we have done that in the tech profile and passed the node name in to the role via a role variable?

ansible/roles/openshift_udp_limits/tasks/main.yml
```diff
@@ -75,7 +65,9 @@
       after: "id_outbound_udp_filtering"
       regex: "id_outbound_udp_rule_1_"
       block: |-
-        -A oo-udp -d {{ servicenetwork }} -o eth0 -p udp -m udp --dport 53 -j RETURN
+        {# Note: sdodson says that pod DNS resolution happens via dnsmasq running on the node. That's
+                 why we only have DNS traffic to the node whitelisted prior to rate limiting. #}
+        -A oo-udp -d {{ oo_private_ip }}/32 -o eth0 -p udp -m udp --dport 53 -j RETURN
```
joelsmith: @twiest likewise, is it a mistake to use `oo_private_ip` here in the role, and should we instead pass it in from the tech profile?

ansible/roles/openshift_udp_limits/tasks/main.yml
twiest: Yes, we should never reference global variables (aka direct hostvars) in our roles.

Instead, our roles should have role variables and we should pass in the value to the role.

Here's an example:
https://github.com/openshift/openshift-ansible-ops/blob/prod/ops_tech_profiles/openshift_master/3.5/meta/main.yml#L18-L19

ansible/roles/openshift_udp_limits/tasks/main.yml
twiest: Yes, it should be passed into the role using a role variable.

twiest (dismissed):

dak1n1 (approved):
:+1:

joelsmith:
[test]

joelsmith:
merging despite failure since bot is broken

PR #2610, opened 2017-06-12
initial version of developer access wrapper tool

joelddiaz:
only allows for:
oc get nodes
oc get nodes -o json
rpm -qa

Allow limited remote access to certain ssh public keys by making /home/<common user>/.ssh/authorized_keys entries like:

command="/usr/bin/devaccess_wrap READ_SSH <username>" ssh-rsa .....

Any commands run through this unix user will be caught and processed by the devaccess_wrap command.

It will reference the ACL in /etc/openshift_tools/devaccess_users.yaml to check against which commands the user is allowed to run, and if allowed, run the commandand return the output.

### DIFF
--- a/scripts/devaccess/devaccess_users.yaml
+++ b/scripts/devaccess/devaccess_users.yaml
@@ -4,11 +4,11 @@ roles:
   - oc get nodes -ndefault
   - oc get nodes -ndefault -ojson
   - rpm -qa
-- name: super_role
+- name: nobody_role
   commands:
   - uname
 
 users:
-- username: superperson
+- username: nobody
   roles:
-  - super_role
+  - nobody_role
--- a/scripts/devaccess/devaccess_users.yaml
+++ b/scripts/devaccess/devaccess_users.yaml
@@ -1,0 +1,14 @@
+roles:
+- name: ALL
+  commands:
+  - oc get nodes -ndefault
+  - oc get nodes -ndefault -ojson
+  - rpm -qa
+- name: super_role
+  commands:
+  - uname
+
+users:
+- username: superperson
+  roles:
+  - super_role
--- a/scripts/devaccess/devaccess_wrap.py
+++ b/scripts/devaccess/devaccess_wrap.py
@@ -413,7 +413,7 @@ class DevGet(object):
 
         return cmd in self._allowed_commands
 
-    def cmd_not_allowed(self, user_cmd):
+    def cmd_not_allowed(self):
         ''' Print generic info when user isnt' able to
             run a command.
         '''
@@ -441,13 +441,13 @@ class DevGet(object):
                 results = self._command_dict[generic_cmd](self._oc_cmd)
                 print results
             else:
-                self.cmd_not_allowed(full_cmd)
+                self.cmd_not_allowed()
         # non-oc command run handling
         elif self.can_run_cmd(cmd):
             results = self._command_dict[cmd](cmd)
             print results
         else:
-            self.cmd_not_allowed(cmd)
+            self.cmd_not_allowed()
 
 
 if __name__ == '__main__':
--- a/scripts/devaccess/devaccess_wrap.py
+++ b/scripts/devaccess/devaccess_wrap.py
@@ -1,0 +1,454 @@
+#!/usr/bin/python
+# vim: expandtab:tabstop=4:shiftwidth=4
+'''
+    Tool to provide developers with ability to remotely run a limited
+    set of commands.
+'''
+# Disabling invalid-name because pylint doesn't like the naming conention we have.
+# pylint: disable=invalid-name
+
+import logging
+import os
+import re
+import subprocess
+import sys
+import yaml
+
+class OCCmd(object):
+    ''' Class to hold and standardize building of /usr/bin/oc commands
+        out of raw string command line text '''
+    def __init__(self, raw_cmd):
+        # original command line text
+        self._raw_cmd = raw_cmd
+
+        self._params = {}
+        # always define a namespace
+        self._params['namespace'] = 'default'
+        # whether special output formatting is requested
+        self._params['output_format'] = None
+        # --follow or -f
+        self._params['follow'] = None
+        # -c for 'oc get logs -c <container>
+        self._params['container'] = None
+
+        # the actual command to run without the params
+        # (ie oc get pods or oc get secrets my-secret)
+        self._params['verb'] = None
+        self._params['type'] = None
+        self._params['subject'] = None
+        self.parse_cmd()
+
+    def get_namespace(self, cmd):
+        ''' find a namespace (if passed in) and return cmd
+            without the namespace-related parameters '''
+
+        cmd_split = cmd.split()
+        delete_tokens = []
+
+        for x in range(0, len(cmd_split)):
+            if cmd_split[x] == '-n':
+                # token was exactly '-n' so next token is the namespace
+                self._params['namespace'] = cmd_split[x+1]
+
+                # make sure to mark these items for removal from list
+                # before returning the command
+                delete_tokens.append(cmd_split[x])
+                delete_tokens.append(cmd_split[x+1])
+
+                # skip next token since we already processed it
+                x = x + 1
+            elif cmd_split[x].startswith('-n'):
+                # we have a namespace in the format of -n<namespace>
+                self._params['namespace'] = re.sub('-n', '', cmd_split[x])
+
+                delete_tokens.append(cmd_split[x])
+            # elif match --namespace and --namespace=<namespace> params
+
+        # now clean up namespace-related tokens before returning
+        # cmd without the namespace-related parameters
+        for token in delete_tokens:
+            cmd_split.remove(token)
+
+        new_cmd = " ".join(cmd_split)
+        return new_cmd
+
+    def get_follow(self, cmd):
+        ''' find -f --follow (if passed in) and return cmd without
+            the param '''
+
+        cmd_split = cmd.split()
+        delete_tokens = []
+
+        for x in range(0, len(cmd_split)):
+            if cmd_split[x] == '-f' or cmd_split[x] == '--follow':
+                self._params['follow'] = '--follow'
+                delete_tokens.append(cmd_split[x])
+
+        for token in delete_tokens:
+            cmd_split.remove(token)
+
+        new_cmd = " ".join(cmd_split)
+        return new_cmd
+
+    def get_output_format(self, cmd):
+        ''' find output format parameters (if passed in) and return cmd
+            without the format parameters '''
+
+        cmd_split = cmd.split()
+        delete_tokens = []
+
+        for x in range(0, len(cmd_split)):
+            if cmd_split[x] == '-o':
+            # token was exactly '-o' so next token is output format
+                self._params['output_format'] = cmd_split[x+1]
+
+                # mark tokens for removal
+                delete_tokens.append(cmd_split[x])
+                delete_tokens.append(cmd_split[x+1])
+
+                # skip next token since we already processed it
+                x = x + 1
+            elif cmd_split[x].startswith('-o'):
+                # we have an output in format -o<output_format>
+                self._params['output_format'] = re.sub('-o', '', cmd_split[x])
+
+                delete_tokens.append(cmd_split[x])
+
+        # clean up output-related tokens and return resulting string
+        for token in delete_tokens:
+            cmd_split.remove(token)
+
+        new_cmd = " ".join(cmd_split)
+        return new_cmd
+
+    def get_container(self, cmd):
+        ''' Take '-c' command line param (for oc logs)
+            Return remaining string '''
+        cmd_split = cmd.split()
+        delete_tokens = []
+
+        for x in range(0, len(cmd_split)):
+            if cmd_split[x] == '-c':
+                self._params['container'] = cmd_split[x+1]
+
+                # mark tokens for removal
+                delete_tokens.append(cmd_split[x])
+                delete_tokens.append(cmd_split[x+1])
+
+                # skip next token since we already processed it
+                x = x + 1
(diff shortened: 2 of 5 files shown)


scripts/devaccess/devaccess_users.yaml
```diff
@@ -0,0 +1,14 @@
...
+  commands:
+  - oc get nodes -ndefault
+  - oc get nodes -ndefault -ojson
+  - rpm -qa
+- name: super_role
+  commands:
+  - uname
+
+users:
+- username: superperson
```
dbaker-arch: This is granting access to a randomly named user.  Possible to include this block commented out, or inactive to demonstrate format without granting privs to a random user?

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,454 @@
...
+        # all that is left should be: 'oc <verb> <type> <optional-subject>'
+        #
+        cmd = self.get_verb_type_subject(cmd_no_container)
+
+
+        # should be nothing left after we parsed all the tokens
+        if cmd != "":
+            raise Exception("Unprocessed command tokens left.")
+
+    def normalized_cmd(self, generic=False):
```
dbaker-arch, line 207: This function looks to be "oc" specific.  I suggest it has "oc" in the function name to make that clear, and keep namespace tidy for future commands that may need similar processing..

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def setup_logging(self):
+        ''' Configure logging '''
+
+        if os.environ.has_key("DEVACC_DEBUG") or self._debug is True:
+            log_level = logging.DEBUG
+        else:
+            log_level = logging.INFO
+
+        logging.basicConfig(filename=self._config['logfile_path'],
```
dbaker-arch, line 402: Does this need file permissions to be defined?

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,454 @@
...
+        '''
+
+        return cmd in self._allowed_commands
+
+    def cmd_not_allowed(self, user_cmd):
+        ''' Print generic info when user isnt' able to
+            run a command.
+        '''
+
+        print "Command not supported/allowed: {}".format(user_cmd)
```
dbaker-arch: Printing user input - esp on a "permission denied" - is how XSS started.  Suggest you do not print that input.

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,454 @@
...
+        print "Allowed commands:"
```
dbaker-arch: This is command enumeration -- suggest you do not print what the user can do in response to their requesting access they don't have.

scripts/devaccess/devaccess_users.yaml
joelddiaz: would it be better to grant it to user 'root'?

scripts/devaccess/devaccess_wrap.py
joelddiaz: normalized_cmd() is a method of the OCCmd object. is that not enough to indicate it's oc-specific?

scripts/devaccess/devaccess_wrap.py
joelddiaz: Yes. The way I see it, the role that installs this RPM (openshift-tools-scripts-devaccess) will have to place a file with correct permissions and populate the config file pointing to the file.

scripts/devaccess/devaccess_users.yaml
dbaker-arch: After further review, I see this is a sample config, and won't be used as a live/initial config.  Granting priv escalation to root seems incorrect.  Perhaps "nobody" or "guest" would be better example users?

scripts/devaccess/devaccess_wrap.py
joelddiaz: It only prints what the specific user can do. Not every possible command the wrapper implements. The thought here was that if the user used something like 'oc logs <pod-name-here> --follow', but we only supported the format 'oc logs <pod-name-here> -f', they could see to use '-f'.

scripts/devaccess/devaccess_wrap.py
dbaker-arch: Good point - yes, this is enough.

scripts/devaccess/devaccess_wrap.py
dbaker-arch: Understood, but IMHO still too much information.  Separately, I feel it's not appropriate to display this list on this type of error.

For comparison, sudo does have the "-l" list option which lists valid commands.  Perhaps a "-l / list" option is acceptable, but might have the option to enable/disable per user, and can be discussed separately.

joelddiaz:
[test]

joelddiaz:
[test]

joelddiaz:
[test]

joelddiaz:
[test]

joelddiaz:
@dbaker-rh @joelsmith added test cases, addressed feedback, and dealt with pylint (now that jenkins is back in a working state). This should be good to re-review now.

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,455 @@
...
+
+        for x in range(0, len(cmd_split)):
+            if cmd_split[x] == '-n':
+                # token was exactly '-n' so next token is the namespace
+                self._params['namespace'] = cmd_split[x+1]
+
+                # make sure to mark these items for removal from list
+                # before returning the command
+                delete_tokens.append(cmd_split[x])
+                delete_tokens.append(cmd_split[x+1])
```
dbaker-arch: This doesn't look safe, and will behave oddly if tokens are not unique.
For consideration: "command -a foo -b foo".  If this code fragment were used here, it would delete both occurrences of "foo".  This might apply here if namespace and output format are the same string -- can you have a namespace called "text" or "json"?
Can you use "import argparse" for processing instead?

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,455 @@
...
+        ''' find a namespace (if passed in) and return cmd
+            without the namespace-related parameters '''
+
+        cmd_split = cmd.split()
+        delete_tokens = []
+
+        for x in range(0, len(cmd_split)):
+            if cmd_split[x] == '-n':
+                # token was exactly '-n' so next token is the namespace
+                self._params['namespace'] = cmd_split[x+1]
```
dbaker-arch: For consideration: "command -n -n foo".  This will take the second "-n" as the namespace which is probably invalid.  We should validate the format of the next token before using it as a namespace.

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,455 @@
...
+        self.parse_cmd()
+
+    def get_namespace(self, cmd):
+        ''' find a namespace (if passed in) and return cmd
+            without the namespace-related parameters '''
+
+        cmd_split = cmd.split()
+        delete_tokens = []
+
+        for x in range(0, len(cmd_split)):
```
dbaker-arch: For consideration: "command foo bar baz -n"
Doesn't look like this correctly handles -n as the last argument as it blindly expects a token to follow.

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,455 @@
...
+                # make sure to mark these items for removal from list
+                # before returning the command
+                delete_tokens.append(cmd_split[x])
+                delete_tokens.append(cmd_split[x+1])
+
+                # skip next token since we already processed it
+                x = x + 1
+            elif cmd_split[x].startswith('-n'):
+                # we have a namespace in the format of -n<namespace>
+                self._params['namespace'] = re.sub('-n', '', cmd_split[x])
```
dbaker-arch: Same as above: need to validate the string fragment left behind to make sure it's an appropriate namespace format.

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,455 @@
...
+            cmd = os.environ.get("SSH_ORIGINAL_COMMAND", "")
+
+            # SSH_ORIGINAL_COMMAND will include the whole command
+            # as a continuous string
+            args = cmd
+
+            # The user's authorized_keys will force the command run to be:
+            # <path to devaccess_wrap> READ_SSH <username>
+            # Save the username for permission lookups later.
+            user = sys.argv[2]
```
dbaker-arch: Validate input.  Is there a way to run this with a hostile environment?

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,455 @@
...
+                            level=log_level)
+
+    def can_run_cmd(self, cmd):
+        ''' Return True/False for whether user is allowed to run the command
+        '''
+
+        return cmd in self._allowed_commands
+
+    def cmd_not_allowed(self):
+        ''' Print generic info when user isnt' able to
```
dbaker-arch: Typo: isn't

scripts/devaccess/test_devget.py
```diff
@@ -0,0 +1,96 @@
...
+    # No 'self' use is intended here
+    # pylint: disable=no-self-use
+    def setUp(self):
+        ''' steps to run before every test run '''
+        DevGet.CONFIG_FILE = './test_devaccess_config.yaml'
+        sys.argv = ['devaccess_wrap', 'READ_SSH', 'nobody']
+
+    def test_unsupported_command(self):
+        ''' Test trying to run a command that isn't supported '''
+        os.environ['SSH_ORIGINAL_COMMAND'] = '/usr/bin/blah'
```
dbaker-arch, line 44: more test cases:

same token in two arguments:  -n json -o json
no namespace given:  oc get foo -n
invalid namespace given:  oc get foo -n (long string, string with % chars, string with path traversal, etc)

scripts/devaccess/devaccess_wrap.py
joelddiaz: If argparse is capable of handling the situation where the namespace parameter can show up just about anywhere in the argument list, I'm unaware of how to handle it:

oc get pods -n namespace podname
oc get pods podname -n namespace

Both of these are allowed and do the exact same thing.

As far as the -a foo -b foo, I'll look into redoing this by removing items by list index position so it removes the exact item.

scripts/devaccess/devaccess_wrap.py
joelddiaz: This is an interesting case. The way this would work out (with the way the argument parsing is currently written) is that the -n param would have a value of -n, and there would be this dangling unconsumed parameter/token 'foo'. There is a check after all the argument parsing that ensures that all parameters/tokens have been consumed, and raises and exception if there are spare parameters/tokens.

I'll add some regex handling to make sure that parameter values don't start with '-'.

scripts/devaccess/devaccess_wrap.py
joelddiaz: Two ways I can see to fake this.
1) User modifies .ssh/authorized_keys to replace the user argument. This implies that they have write access to .ssh/authorized_keys, so they can just remove the 'command=...' wrapper.
2) User 'bob' tries to run this wrapper script and sets the appropriate variables to get the script to read 'bob'-controlled argv. At this point, the script is running with the 'bob' permissions, so 'bob' wont be able to do anything that 'bob' couldn't already do.

I can't think of another way this could be attacked right now...

dbaker-arch: Since we're writing a wrapper to manage the escalation of privs, I suggest we validate ALL input regardless of whether it's reasonable and expected that it might be exploitable.

In this case, a test as simple as $user =~ /[^a-zA-Z0-9]/ ... would suffice.  Does the input have anything that's not alphanumeric?  If so, abort.

scripts/devaccess/devaccess_wrap.py
joelddiaz: Added test case (and code) to handle this.

scripts/devaccess/test_devget.py
joelddiaz: Added test case, and changed token removal to be by-index position.

joelddiaz:
@dbaker-rh @joelsmith This version should address previous feedback.

joelddiaz:
[test]

scripts/devaccess/devaccess_wrap.py
```diff
@@ -0,0 +1,451 @@
...
+            cmd = os.environ.get("SSH_ORIGINAL_COMMAND", "")
+
+            # SSH_ORIGINAL_COMMAND will include the whole command
+            # as a continuous string
+            args = cmd
+
+            # The user's authorized_keys will force the command run to be:
+            # <path to devaccess_wrap> READ_SSH <username>
+            # Save the username for permission lookups later.
+            if re.match("^[A-Za-z0-9]+$", sys.argv[2]) is None:
```
dbaker-arch: Username cannot start with a digit, which is permitted here (this is the prohibitive constraint that prevents a username from being all numeric, thus indistinguishable from a uid).

Technically, username can contain an underscore even though ours never do.

Also, username should be length limited.  I think 32 chars is the technical limit.

joelddiaz:
@dbaker-rh should have the username regex update with this last change

dbaker-arch:
There are two minutae that come up re username regex (technically, usernames can contain underscores, are 32 chars or less in length, and may optionally have a literal $ as the last character.  It's all, now I discover, in the man page for useradd(8) if you want the gory details).  What you have will cover every use case that I can foresee coming anytime soon so there's no need to make an additional change just to handle that update.

That aside, looks good to me.  Let's move forward.

dbaker-arch (approved):
+1 Thanks.

PR #2613, opened 2017-06-13
metrics every 30m, 2 fails in a row to trigger

drewandersonnz:
Change to have 2x failed checks in a row, but check twice as fast.
I assume the original intention was to run every 30m, not once per 60m
(on :30).


BlueShells:
:+1:

PR #2615, opened 2017-06-13
items don't exist at the moment, causing broken zabbix config loop

drewandersonnz:
also, lines with excess whitespace


BlueShells:
:+1:

PR #2617, opened 2017-06-13
Fix DNS whitelist destination, clean up node name

joelsmith:
Also, avoid use of globals in UDP limits role


joelsmith:
@dak1n1 reviewed and approved stg PR #2609. Merging despite check failures due to broken bot.

PR #2618, opened 2017-06-14
Cert add comments

stenwt:


stenwt:
@joelddiaz reviewed #2612

PR #2619, opened 2017-06-15
changed project operation check severity from high to avg

zhiwliu:
since there is a bug in the script and we can not build the image succeed, so for now we changed the severity from high to avg, once the bug fixed, changed it back to high


BlueShells:
:+1:

PR #2624, opened 2017-06-15
missed disc.

stenwt:
already ran through and added so syntax is confirmed, this just cements it in config management


gburges:
👍

PR #2626, opened 2017-06-15
back iops alert down to average, it's too flappy and generates false …

stenwt:
…alerts as is


robotmaxtron (approved):
👍 pending checks passing.

PR #2629, opened 2017-06-16
Vendoring github 3.6

kwoodson:


kwoodson:
@joelddiaz :+1:  in #2628

PR #2632, opened 2017-06-19
remove pylint workaround and add /usr/bin/oc workaround

joelddiaz:
pylint is in EPEL again

Also, just install the RPM package that provides /usr/bin/oc for OpenShift (for the RHEL7 container).


joelddiaz:
👍 #2630 in stg

PR #2634, opened 2017-06-19
Add wrapper script to gather cluster logs

codificat:
openshift/aos-cd-jobs#318 sets the base for developers to collect cluster logs by using an SSH key that will invoke a log gathering script for the cluster of choice:
```sh
ssh -i log_gathering_key use-tower2.ops.rhcloud.com <clusterName>
```
This is the entry script associated with the log gathering SSH key, i.e. what should go into the `command=...` part in `authorized_keys` for the entries corresponding to `log_gathering_key` 

The script validates that a valid clusterName has been provided and invokes the log gathering script with it as an argument.

@jupierce FYI

PR #2638, opened 2017-06-20

codificat:
Re-sending #2634 on the stg branch, copying over @jupierce 's comment with the background:

@joelsmith FYR. This relates to the work we have previously discussed to allow non-ops engineering team member to grab (non-application) logs from a select set of clusters. This script (modeled on the one you provided for verifying cluster operations initiated by Jenkins) will be the entrypoint for an ssh key in tower2's authorized_keys file.

Ultimately, an engineer should be able to down an ssh key and run:
```
ssh -i logkey.pem tower2 free-int > /tmp/free-int-logs.tgz
```
And receive a tarball with atomic-master journal entries.

The associated keys will be rotated weekly and made available to the engineering team in a private github repo which is already locked down to openshift developers and regularly culled.

- Our most recent conversations about enabling engineering log access focused on a web application front end for log gathering, but I've backed off that idea until we have some spare cycles.
- This script will be loaded onto tower using an RPM.

### DIFF
--- a/scripts/cicd/verify-gather-logs-operations.py
+++ b/scripts/cicd/verify-gather-logs-operations.py
@@ -1,0 +1,91 @@
+#!/usr/bin/python
+
+'''
+Validates command to gather logs from a cluster
+
+Developer usage:
+
+   ssh -i log_gathering_key use-tower2.ops.rhcloud.com -- -c <clusterName> -u <kerberosID>
+
+clusterName is checked against a list of known/valid cluster names.
+
+kerberosID is the Kerberos ID of the developer requesting logs, for tracking purposes
+'''
+
+# pylint: disable=invalid-name
+# pylint: disable=logging-not-lazy
+# pylint: disable=broad-except
+
+# The log_gathering_key is kept in the shared-secrets repo. It is generated and
+# rotated weekly by the Jenkins job cluster/rotate-log-access-key. The public
+# key in .ssh/authorized_keys looks like this:
+#
+#  command="verify-gather-logs-operations.py" ssh-rsa ThePublicKey logs_access_key_week_22-YMD-H:M:S
+
+import argparse
+import os
+import re
+import sys
+import socket
+import logging
+import logging.handlers
+
+# The list of valid cluster names to request logs from
+VALID_CLUSTER_NAMES = [
+    'free-int',
+    'free-stg',
+    'starter-us-east-1',
+    'starter-us-east-2',
+    'starter-us-west-2'
+]
+
+# The command that is invoked to perform the actual log collection.  This
+# command should expect one argument, the cluster name on which to operate, and
+# produce the logs as a tarball in stdout, which gets passed directly as the
+# output of this script.
+LOG_GATHER_CMD = '/home/opsmedic/aos-cd/git/aos-cd-jobs/tower-scripts/bin/gather-logs.sh'
+
+HOSTNAME = socket.gethostname()
+
+logger = logging.getLogger('verify_command_logger')
+logger.setLevel(logging.INFO)
+logger.addHandler(logging.handlers.SysLogHandler('/dev/log'))
+
+def valid_krbid(username):
+    '''Sanity check that the username looks like valid according to the description
+    of valid usernames from useradd(8)
+    '''
+    if re.match(r'^[a-z_][a-z0-9_-]*[$]?$', username) and len(username) <= 32:
+        return username
+    else:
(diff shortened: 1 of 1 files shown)


jupierce:
@codificat We've gotten through an initial review, but we've been asked to prompt for and log the developer's kerberos id (no validation necessary) in the verify script. 

So the interaction would look something like:
```
sh -i logkey.pem opsmedic@tower2 free-int > /tmp/free-int-logs.tgz
Enter your kerberos id:  jupierce
```
The prompt would obviously need to be piped to stderr in order to not corrupt resultant tarball. Once you have this in place, can you schedule some time with @dbaker-rh to do a final review/test of this script from a security perspective?

dbaker-arch:
I had more in mind putting kerb id on the command line:

ssh -i logkey.pem opsmedic@tower2 -u dbaker -c free-int

To recap the pertinent part of the conversation, having shared keys removes the ability to audit access.  Absent doing something high-cost like provisioning rotating keys for each developer with access, a procedural policy to self-identify goes a long way to closing that hole.

jupierce:
@dbaker-rh An argument based method works for me. 
@codificat With multiple parameters, we will also need help text, -h, --help to describe those arguments out stderr.

codificat:
@dbaker-rh @jupierce just pushed a new version that uses arguments. BTW, note that we need to separate ssh arguments from the command itself with `--`:

```
ssh -i logkey.pem opsmedic@tower2 -- -u dbaker -c free-int
```

Argument parsing is done with `argparse` which among other things provides the requested `--help` for free... albeit not via stderr. This is meant to be run more or less interactively, so if someone explicitly invokes with `-h/--help` I hope it's ok to get that help via stdout and a non-error exit status?

On the other hand though, one thing we lose in this version with `argparse` is logging of incorrect invocations, e.g. an invocation referencing an invalid cluster name or without a username will get an error, but this won't get logged via syslog.

Wanted to get your thoughts on this: If we need to handle the latter (i.e. log incorrect invocations) this version will have to be adjusted.

codificat:
To clarify: we can log every invocation - it's just that argparse will handle incorrect arguments itself and won't explicitly syslog them. The question here is if we need an explicit log entry that explains incorrect arguments.

@dbaker-rh let me know if we can schedule a short meeting to review this (and the rest) thanks!

scripts/cicd/verify-gather-logs-operations.py
```diff
@@ -0,0 +1,90 @@
...
+
+HOSTNAME = socket.gethostname()
+
+logger = logging.getLogger('verify_command_logger')
+logger.setLevel(logging.INFO)
+logger.addHandler(logging.handlers.SysLogHandler('/dev/log'))
+
+def valid_krbid(username):
+    ''' Sanity check that the username looks like a valid kerberos id '''
+    if re.match(r'^[a-z]+$', username) and len(username) <= 8:
```
dbaker-arch: *OUR* kerb usernames are restricted to 8 chars, but this will break in other environments.  The username cannot be all numeric, but can contain numbers.  Doing some poking about, it seems it's ok to contain underscores as well, and can be up to 31 or 32 chars long.

Consider: ^([A-Za-z_][A-Za-z0-9_]{0,30})$

scripts/cicd/verify-gather-logs-operations.py
```diff
@@ -0,0 +1,90 @@
...
+        return username
+    else:
+        msg = '%s is not a valid Kerberos ID' % username
```
dbaker-arch: This error is misleading.  We're not checking if the input is a valid kerb id, but we're specifically checking if it's an acceptable format for a kerb id.

More importantly, if the input does not pass input validation it should NEVER be displayed back to the user.  We already know it's in an unknown, and perhaps unsafe format.

Consider: msg = "Kerberos ID was not provided in acceptable formal"

scripts/cicd/verify-gather-logs-operations.py
codificat: Sure! I was just assuming that we should aim at validating (to some extent) that it's one of *our* krb users here, hence why the check for 1-8 lowercase letters only. But I'm happy to replace it with a more generic regexp if that's preferable.

Combining your suggestion with the info from `useradd(8)`, I'd go for something like `^[a-z_][a-z0-9_-]{0,30}[$]?$`

scripts/cicd/verify-gather-logs-operations.py
codificat: Ok

dbaker-arch:
+1

PR #2647, opened 2017-06-22
Fixing empty inventory on reboot.

kwoodson:


kwoodson:
@joelsmith 👍  in #2646

PR #2654, opened 2017-06-26
made generate byo have capability to do local ip addresses

mwoodson:


mwoodson:
@jupierce approved #2651

PR #2656, opened 2017-06-27
put in a default for the us_private_ips

mwoodson:


mwoodson:
@kwoodson approved #2655

PR #2657, opened 2017-06-27
Pod count fix

stenwt:
current check alerts if we have e.g. 3 registry pods on 2 hosts. Fix that so larger clusters don't alert.


stenwt:
@joelddiaz reviewed #2658

PR #2660, opened 2017-06-27

codificat:
This is the entry point script associated with the log gathering SSH key.

@dbaker-rh approved in #2638

### DIFF
--- a/scripts/cicd/verify-gather-logs-operations.py
+++ b/scripts/cicd/verify-gather-logs-operations.py
@@ -1,0 +1,91 @@
+#!/usr/bin/python
+
+'''
+Validates command to gather logs from a cluster
+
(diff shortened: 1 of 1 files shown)


codificat:
I'm afraid I did it wrong - cherry-picked the change in stg. Closing, will do it properly.

codificat:
Actually what I did wrong was the original branch off from stg... by now though, cherry-pick seems to be the way to get the change into `int`, so reopening.

codificat:
Really sorry for the noise here, but no: a better way to fix this was to redo the original PR but branching off `prod` as it should have been, so further PRs to `prod` and `int` work fine.

The proper PR to `int` following that approach is #2684 so closing this one for good.

PR #2661, opened 2017-06-27

codificat:
This is the entry point script associated with the log gathering SSH key.

@dbaker-rh approved in #2638


codificat:
I'm afraid I did it wrong - cherry-picked the change in stg. Closing, will do it properly.

PR #2663, opened 2017-06-27
initial version of developer access wrapper tool

joelddiaz:
only allows for:
oc get nodes
oc get nodes -o json
rpm -qa

Allow limited remote access to certain ssh public keys by making /home//.ssh/authorized_keys entries like:

command="/usr/bin/devaccess_wrap READ_SSH " ssh-rsa .....

Any commands run through this unix user will be caught and processed by the devaccess_wrap command.

It will reference the ACL in /etc/openshift_tools/devaccess_users.yaml to check against which commands the user is allowed to run, and if allowed, run the commandand return the output.


joelddiaz:
👍 #2610 in stg

PR #2668, opened 2017-06-28
Adding disk encryption feature along with byo options.

kwoodson:


kwoodson:
@mwoodson 👍  in #2667

PR #2675, opened 2017-06-29
vendored openshift-ansible-3.5.91

mwoodson:


mwoodson:
@dyocum approved #2673

PR #2677, opened 2017-06-29
Enabling docker version argument

jupierce:
@dbaker-rh An important change to the CD to allow us to send specific / highly controlled parameters allow to our CD operations. Please let me know your thoughts. 

+ @mwoodson

### DIFF
--- a/scripts/cicd/verify-cicd-operation.py
+++ b/scripts/cicd/verify-cicd-operation.py
@@ -7,9 +7,15 @@
 # pylint: disable=bare-except
 
 #Jenkins:
-#ssh use-tower1.ops.rhcloud.com <operation>
+#ssh use-tower1.ops.rhcloud.com <operation> [approved_arg1] [approved_arg2] ...
+# Current approved arguments:
+#   "docker-version=<rpmname-version-release>"
+#       This argument specifies a docker RPM NVR that should be used for openshift-ansible upgrade operations.
+#       An example "NVR: docker-1.12.6-30.git97ba2c0.el7"
 #
-#in ssh/.authorized_keys
+# Pull requests which add new arguments must be approved by the security team.
+#
+#Associate the cluster to operate on with an ssh key in .ssh/authorized_keys
 # command=verify-cicd-operation.py <clusterid> <really long key right here>
 
 import re
@@ -20,8 +26,6 @@ import logging
 import logging.handlers
 
 def runner(program, *args):
-    ''' string '''
-
     try:
         os.execlp(program, program, *args)
     except:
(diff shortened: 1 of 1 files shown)


scripts/cicd/verify-cicd-operation.py
```diff
@@ -34,17 +32,38 @@ def runner(program, *args):
...
+
+if not match:
+    logger.info("%s Restricted key '%s' disallowed operation: %s" % (os.path.basename(__file__), cluster_id, op))
+    print("Operation doesn't match an allowed pattern ")
+    print("REJECTED ON HOST: " + socket.gethostname())
+    sys.exit(10)
+
+operation_args = []
+for s in args:
+    if re.match("(^docker-version=[a-zA-Z0-9.\-_]+$)", s):  # Example docker version: docker-1.12.6-30.git97ba2c0.el7
```
dbaker-arch: If that's an RPM identifier, do you also need to permit a single ":"?

scripts/cicd/verify-cicd-operation.py
```diff
@@ -34,17 +32,38 @@ def runner(program, *args):
...
+
+match = re.match(r"(?P<operation>install|delete|upgrade|status|logs|perf1|perf2|perf3|build-ci-msg|smoketest)", op)
+
+if not match:
+    logger.info("%s Restricted key '%s' disallowed operation: %s" % (os.path.basename(__file__), cluster_id, op))
+    print("Operation doesn't match an allowed pattern ")
+    print("REJECTED ON HOST: " + socket.gethostname())
+    sys.exit(10)
+
+operation_args = []
```
dbaker-arch: Needs a comment to say "validating args".  Needs a comment to say what approved args are -- just package names and their versions?

jupierce: ack . Updated.

scripts/cicd/verify-cicd-operation.py
jupierce: I think we only need RPM N-V-R support right now. We can expand if it ever doesn't meet our needs.

jupierce:
@dbaker-rh Update submitted. Thanks!

scripts/cicd/verify-cicd-operation.py
```diff
@@ -7,7 +7,7 @@
 # pylint: disable=bare-except
 
 #Jenkins:
-#ssh use-tower1.ops.rhcloud.com <operation>
+#ssh use-tower1.ops.rhcloud.com <operation> [approved_arg1] [approved_arg2] ...
```
dbaker-arch, line 10: Comment would do well to address - what are approved args?  what are they used for?  how are they approved?

scripts/cicd/verify-cicd-operation.py
```diff
@@ -34,17 +32,42 @@ def runner(program, *args):
 logger.setLevel(logging.INFO)
 logger.addHandler(logging.handlers.SysLogHandler('/dev/log'))
 
-keyname = sys.argv[1] if len(sys.argv) >= 2 else "<none specified>"
+cluster_id = sys.argv[1] if len(sys.argv) >= 2 else "<none specified>"
```
dbaker-arch: Since this script is scrutinized from a security perspective, I would honestly do an input validation on cluster_id anyway, and also note here that it comes from the authorized_keys file, not from direct user input.  ANY time you pass args through a script like this they should have been validated.

dbaker-arch (changes requested):
Add a couple of comments for future eyes reviewing the doc.  If you have time, validate clusterid.  Otherwise, all looks good.

jupierce:
@dbaker-rh Comments and validation added

scripts/cicd/verify-cicd-operation.py
```diff
@@ -34,17 +38,54 @@ def runner(program, *args):
...
+# In other words, the ssh user cannot specify the cluster_id directly (they must have the correct private key
+# AND the key must be setup in the authorized_keys file appropriately).
+if len(sys.argv) != 2:
+    print("No cluster identifier specified")
+    sys.exit(12)
+
+cluster_id = sys.argv[1]
+
+# Sanity check the cluster_id from authorized_keys
+if not re.match("(^[a-zA-Z0-9.\-_]+$)", cluster_id):
```
dbaker-arch: Minor : put the "-" at the end of the char ranges to not need to escape it.
Less minor : cluster names starting with a hyphen will complicate things terribly.

Suggest:   if not re.match("(^[a-zA-Z0-9][a-zA-Z0-9._-]+$)", cluster_id):

jupierce: @dbaker-rh Tighter regex for cluster validation pushed.

scripts/cicd/verify-cicd-operation.py
```diff
@@ -34,17 +38,54 @@ def runner(program, *args):
...
+    print("Operation doesn't match an allowed pattern ")
+    print("REJECTED ON HOST: " + socket.gethostname())
+    sys.exit(10)
+
+# Validate remaining arguments against strict patterns.
+# Only matched patters may be passed on to cicd-control. See approved list
+# at the top of this file.
+operation_args = []
+for s in args:
+    if re.match("(^docker-version=[a-zA-Z0-9.-_]+$)", s):  # Example docker version: docker-1.12.6-30.git97ba2c0.el7
```
dbaker-arch: Sorry, missed this one.  That hyphen in the regex needs to be at the end, else you're taking all chars between . and _

jupierce: Fixed

dbaker-arch (approved):
Thanks for being so responsive to my changes.  +1

PR #2694, opened 2017-07-05
change the role path for auto heal docker high usage

BlueShells:
change the role path for auto heal docker high usage


zhiwliu:
:+1:

PR #2696, opened 2017-07-05
adding heal to cleanup docker storage

zhiwliu:


joelddiaz (changes requested):
This doesn't need to be handled as a remote heal.

If possible we prefer to do the heal local to the machine, and this heal action can be done from right inside the host-monitoring container. I'd rather have these cleanup tasks be performed inside the script that detects the low-docker-storage condition.

We can detect the low storage condition, then perform the mitigation actions, then re-check the low storage condition (so that the trigger/alert is quieted down as soon as possible).

zhiwliu:
ok, then I will rollback the zabbix trigger tomorrow.

PR #2700, opened 2017-07-10
Adding eu-west-2 to our supported ami copy region.

kwoodson:


kwoodson:
@mwoodson 👍  in #2699

PR #2703, opened 2017-07-10
allow users to be defined who have no specific role assigned

joelddiaz:
it will only use the default role

### DIFF
--- a/scripts/devaccess/test_devget.py
+++ b/scripts/devaccess/test_devget.py
@@ -145,5 +145,16 @@ ip-172-31-62-44.ec2.internal    Ready,SchedulingDisabled   45d
         with self.assertRaises(DevGetError):
             dg = DevGet()
 
+    @mock.patch('devaccess_wrap.WhitelistedCommands.rpm_qa')
+    def test_user_with_no_defined_roles(self, mock_rpm_qa):
+        ''' Test user with no defined roles (ie only default role membership) '''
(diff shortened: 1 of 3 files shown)


scripts/devaccess/test_devget.py
```diff
@@ -145,5 +145,16 @@ def test_invalid_username(self):
...
             dg = DevGet()
 
+    @mock.patch('devaccess_wrap.WhitelistedCommands.rpm_qa')
+    def test_user_with_no_defined_roles(self, mock_rpm_qa):
+        ''' Test user with no defined roles (ie only default role membership) '''
+        os.environ['SSH_ORIGINAL_COMMAND'] = 'rpm -qa'
+        sys.argv = ['devaccess_wrap.py', 'READ_SSH', 'noroleuser']
+        dg = DevGet()
+
+        mock_rpm_qa.side_effect = ['kernel-3.12.22']
```
dbaker-arch: What does kernel-3.12.22 have to do with a user with no roles?  This isn't expected output from "rpm -qa".

joelddiaz: right, it's just some dummy output.
the test is to make sure that the function gets called because 'rpm -qa' is in the default list of allowed commands.

i can probably change it so that there is no side_effect at all (it was just the remnants of copying an existing test)

dbaker-arch (commented):
+1

PR #2707, opened 2017-07-11
added the second time metrics check if the first time check failed

zhiwliu:


drewandersonnz (approved):
+1, I like the commandDelay between checks.

PR #2728, opened 2017-07-13
added the heal to the low docker storage

zhiwliu:


openshift_tools/monitoring/dockerutil.py
```diff
@@ -8,12 +8,13 @@
 # Adding the ignore because it does not like the naming of the script
 # to be different than the class name
 # pylint: disable=invalid-name
+# pylint: disable=line-too-long
```
joelddiaz: Please put reason we are disabling the particular pylint, and if it only for one line, you can put the disable next to the line that requires it.

See https://github.com/openshift/openshift-tools/blob/prod/openshift_tools/monitoring/zabbix_metric_processor.py#L174-L177 for example

Also, see https://github.com/openshift/openshift-tools/blob/prod/docs/best_practices_guide.adoc#PyLint-rules-MUST-NOT-be-disabled-on-a-whole-file for more details on our pylint guidelines.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -8,12 +8,13 @@
...
+# pylint: disable=wrong-import-position
```
joelddiaz: same. please document reason for pylint disable.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+
+        self.exited_con = []
+        self.useless_img = []
+
+    def get_exited_cons(self):
+        'get the exited containers list'
+        self.all_exited_con_info = self.client.containers(all='true', filters={'status': 'exited'})
+        for container in self.all_exited_con_info:
+            self.exited_con.append(container['Id'])
+        return self.exited_con
```
joelddiaz: Do you really need to return self.exited_con? The only caller of this method is CleanupDockerStorage.run() and it already has access to self.exited_con.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+
+    def get_nouse_images(self):
+        'get dangling images'
+        self.all_dangling_img_info = self.client.images(filters={'dangling': 'true'})
+        for dangling_img in self.all_dangling_img_info:
+            self.useless_img.append(dangling_img['Id'])
+        return self.useless_img
```
joelddiaz: Do you really need to return self.useless_img? The only caller of this method is CleanupDockerStorage.run() and it already has access to self.useless_img.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+
+    def remove_con(self, exited_cons_list):
+        'remove the exited containers from the exited_cons_list if not empty'
+        if len(exited_cons_list) > 0:
+            for con_id in exited_cons_list:
+                try:
+                    self.client.remove_container(container=con_id)
+                except APIError:
+                    pass
```
joelddiaz: Maybe rather than just 'pass', record that fact that an error occurred while trying to remove the exited containers, and return success/failed from this remove_con() method.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+            print "no containers in exited status to remove"
+
+    def remove_dangling_img(self, useless_img_list):
+        'remove the dangling images from the list if the list not empty'
+        if len(useless_img_list) > 0:
+            for img_id in useless_img_list:
+                try:
+                    self.client.remove_image(image=img_id)
+                except APIError:
+                    pass
```
joelddiaz: Maybe rather than just 'pass', record that fact that an error occurred while trying to remove the dangling images, and return success/failed from this remove_dangling_img() method.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+        else:
+            print "no dangling images can be removed"
+
+    def run(self):
+        'run this function to remove containers and images'
+        exited_cons_list = self.get_exited_cons()
```
joelddiaz: no need to save exited_cons_list since the data is already saved in self.exited_con.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+        self.client = docker_client
+        self.all_exited_con_info = None
+        self.all_dangling_img_info = None
+
+        self.exited_con = []
+        self.useless_img = []
+
+    def get_exited_cons(self):
+        'get the exited containers list'
+        self.all_exited_con_info = self.client.containers(all='true', filters={'status': 'exited'})
```
joelddiaz: I don't see a need for the class variable self.all_exited_con_info. Rather than self.all_exited_con_info, just use all_exited_con_info inside get_exited_cons().

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+        for container in self.all_exited_con_info:
+            self.exited_con.append(container['Id'])
+        return self.exited_con
+
+    def get_nouse_images(self):
+        'get dangling images'
+        self.all_dangling_img_info = self.client.images(filters={'dangling': 'true'})
```
joelddiaz: I don't see a need for the class variable self.all_dangling_img_info. Rather than self.all_dangling_img_info, just use all_dangling_img_info inside of get_nouse_images().

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+                    self.client.remove_image(image=img_id)
+                except APIError:
+                    pass
+        else:
+            print "no dangling images can be removed"
+
+    def run(self):
+        'run this function to remove containers and images'
+        exited_cons_list = self.get_exited_cons()
+        useless_img_list = self.get_nouse_images()
```
joelddiaz: no need to save useless_img_list since the data is already saved in self.useless_img

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+        self.remove_con(exited_cons_list)
```
joelddiaz: no need to pass exited_cons_list since remove_con() already has access to self.exited_con.

openshift_tools/monitoring/dockerutil.py
```diff
@@ -196,3 +197,57 @@ def get_ctr_stats(self, ctr, use_cgroups=False):
...
+        self.remove_dangling_img(useless_img_list)
```
joelddiaz: no need to pass useless_img_list since remove_dangling_img() already has access to self.useless_img.

scripts/monitoring/cron-send-docker-metrics.py
```diff
@@ -8,15 +8,17 @@
 # Adding the ignore because it does not like the naming of the script
 # to be different than the class name
 # pylint: disable=invalid-name,import-error
+# pylint: disable=line-too-long
```
joelddiaz: Please put reason we are disabling the particular pylint, and if it only for one line, you can put the disable next to the line that requires it.

joelddiaz (changes requested):

PR #2729, opened 2017-07-13
Provide more information when a check fails.

rubenvp8510:
@jsanda Do you think we could need more info? like `nodetools` info from Cassandra?

### DIFF
--- a/scripts/monitoring/cron-send-metrics-checks.py
+++ b/scripts/monitoring/cron-send-metrics-checks.py
@@ -275,6 +275,13 @@ class OpenshiftMetricsStatus(object):
             logger.warn("Cannot found cassandra-1 PV, the results cannot be persisted")
 
         remote_details_directory = os.path.join(remote_details_directory, 'failure_reports/')
+
+        # Make sure directory exists or create if not exists.
+        try:
+            self.oc.run_user_cmd('exec {} -- mkdir -p {}'.format(cassandra_main_pod_name, remote_details_directory))
+        except subprocess.CalledProcessError:
+            logger.error("Cannot create reports directory inside cassandra PV")
+
         # Trim files, this delete old files and make sure that only have certain number of files
         # this is to prevent fill up the PV.
 
--- a/scripts/monitoring/cron-send-metrics-checks.py
+++ b/scripts/monitoring/cron-send-metrics-checks.py
@@ -24,27 +24,30 @@
 #pylint: disable=bare-except
 #pylint: disable=wrong-import-position
 #pylint: disable=line-too-long
-import ssl
-import urllib2
 import argparse
-import time
-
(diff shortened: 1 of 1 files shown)


mwringe:
I am not a big fan of putting this into the Cassandra PV, but it looks like that is the only choice we have at the moment. At least having this data will be a big gain for us.

Do we have any concern that writing the logs may fill up the PV too much? I don't know if we even need the logs at this point TBH, if we get an alert that this check failed, we should be able to go into the cluster and investigate.

rubenvp8510:
@mwringe I was thinking the same, as you said, if something goes wrong, we can get those logs from the cluster, so I think at the end I gonna modify the PR to only store the report with which pods fail and why.

WDYT?

rubenvp8510:
Or is there any concern about losing logs if some C* cluster dies?

mwringe:
We probably do want to handle logs eventually, I think storing the logs is of secondary concern at the moment and what we need now is just the reason for the failures.

scripts/monitoring/cron-send-metrics-checks.py
```diff
@@ -158,6 +160,9 @@ def get_hawkular_creds(self):
             sys.exit(1)
 
     def check_node_metrics(self):
+
+        result_report = {'success':1, 'failed_pods':[] }
+
```
wshearn: should be after the doc string

scripts/monitoring/cron-send-metrics-checks.py
```diff
@@ -239,13 +289,16 @@ def run(self):
...
+            # persist "false" failure.
+            self.persist_details(metrics_report, pod_report)
             # sleep for 5 seconds, then run the second time node check
             logger.info("The first time metrics check failed, 5 seconds later will start a second time check")
             time.sleep(commandDelay)
             logger.info("starting the second time metrics check")
             metrics_report = self.check_node_metrics()
-
+            # persist second attempt.
+			self.persist_details(metrics_report, pod_report)
```
wshearn: tab to space

rubenvp8510:
This is ready for another review and merge if it's OK. I could reintroduce the code of storing logs in another PR.

wshearn:
[test]

mwringe:
I think this does what we want. Are we able to easily correlate the alert we get from zabbix to the file we need to check in the Cassandra PV?

rubenvp8510:
I named the file with the date and time `failure_report_{date}_{time}.yaml`. What information do we need to correlate it with Zabbix alert?

mwringe:
What else needs to be done to get this merged?

wshearn:
I found a bug, I am going to talk to rubens today when he gets on.

wshearn:
Closing this PR and opening a new one so the bot is happy,

PR #2730, opened 2017-07-17
2 check dnsmasq

zhiwliu:


drewandersonnz (approved):
@zhiwliu has demonstrated a success and failure case, both work as expected (retry on failure 5s later).

PR #2736, opened 2017-07-18
add var that can disable app create on special clusters

ivanhorvath:


ivanhorvath:
Approved by @dak1n1 in #2735

PR #2737, opened 2017-07-18
Add in-memory scanner

joelsmith:
Missing features:
* Select namespaces, pods, containers, and/or processes for scanning
* Support signature whitelists
* Cache results of long-running processes


kwoodson (approved):
This PR looks good.  I like the documentation steps to follow the logic.  Looks good to me.

Only thing I might suggest is using the logging facility to store or write messages to a log.

PR #2744, opened 2017-07-21
Provide Context as to why the check is failing

wshearn:
Moving #2729 to here as jenkins is being stubborn.


wshearn:
@mwoodson ok, I have fully tested this and it is working, can we get a review now?

stenwt (approved):

PR #2746, opened 2017-07-24
Changed openshift_user_policy to take a list of cluster role bindings instead of a single one.

twiest:


twiest:
This pr was +1d in STG PR #2745

PR #2749, opened 2017-07-25
Allow optional nodes for gather logs script

codificat:
The cluster log gathering script is being expanded to allow an optional list of nodes to get logs from, see openshift/aos-cd-jobs#461

This modifies the entry point script to add a parameter for that list of nodes. With this, you can get master logs as before with:

    ssh -i logs_access_key tower -- -u username -c clusterID

or add a list of nodes to get these extra logs:

    ssh -i logs_access_key tower -- -u username -c clusterID -n node1 node2 node3

@dbaker-rh would you have a few mins to review? Feel free to schedule a chat if needed, thanks!

### DIFF
--- a/scripts/cicd/verify-gather-logs-operations.py
+++ b/scripts/cicd/verify-gather-logs-operations.py
@@ -23,6 +23,7 @@ kerberosID is the Kerberos ID of the developer requesting logs, for tracking pur
 #  command="verify-gather-logs-operations.py" ssh-rsa ThePublicKey logs_access_key_week_22-YMD-H:M:S
 
 import argparse
+import getpass
 import os
 import re
 import sys
@@ -41,10 +42,11 @@ VALID_CLUSTER_NAMES = [
 
 # This regex matches a valid hostname: an optional set of DNS labels each
 # ending with dot, followed by a final DNS label. Each DNS label is 1 to
-# 63 digits, letters or '-', but '-' can't be the first or last character.
+# 63 digits, letters or '-', but must start with a letter. This is used
+# to match each of the provided node names
 HOSTNAME_RE = re.compile(
-    r'((?!-)[a-z\d-]{1,63}(?<!-)\.)*'
-    r'((?!-)[a-z\d-]{1,63}(?<!-))$',
+    r'([a-z][a-z\d-]{0,62}\.)*'
+    r'([a-z][a-z\d-]{0,62})$',
     re.IGNORECASE
 )
 
@@ -55,6 +57,8 @@ HOSTNAME_RE = re.compile(
 LOG_GATHER_CMD = '/home/opsmedic/aos-cd/git/aos-cd-jobs/tower-scripts/bin/gather-logs.sh'
 
 HOSTNAME = socket.gethostname()
+USERNAME = getpass.getuser()
+INVOCATION = "%s@%s" % (USERNAME, HOSTNAME)
 
 logger = logging.getLogger('verify_command_logger')
 logger.setLevel(logging.INFO)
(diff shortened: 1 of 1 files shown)


scripts/cicd/verify-gather-logs-operations.py
```diff
@@ -39,6 +39,15 @@
     'starter-us-west-2'
 ]
 
+# This regex matches a valid hostname: an optional set of DNS labels each
+# ending with dot, followed by a final DNS label. Each DNS label is 1 to
+# 63 digits, letters or '-', but '-' can't be the first or last character.
+HOSTNAME_RE = re.compile(
```
dbaker-arch: Minor -- looks like this regexp will incorrectly accept 999.999.999.999 as a hostname.

I don't understand your ?!- $<!- syntax off-hand.  Can this be simplified in the code to make it less opaque?

scripts/cicd/verify-gather-logs-operations.py
```diff
@@ -73,9 +89,15 @@ def gather_logs(command):
                         required=True, type=valid_krbid)
     parser.add_argument('-c', dest='cluster', help="Cluster name",
                         required=True, choices=VALID_CLUSTER_NAMES)
+    parser.add_argument('-n', dest='nodes', help="Nodes to gather logs from",
```
dbaker-arch, line 106: Is it more normal/expected to use: -n node1 node2 node3
Or would the "-n" be repeated:  -n node1 -n node2 -n node3

scripts/cicd/verify-gather-logs-operations.py
codificat: Right: the regex would match IPs too - it's not strictly checking for valid DNS labels... that was the initial intention but I thought that developers might prefer the ability to specify nodes in different ways depending on how they obtain the reference to the node they are interested in (e.g. ansible output might be pointing to an inventory name instead of a hostname). This way they can use node name, inventory name, internal IP or external IP... I will adjust the comment to reflect this, if that's ok.

A verification that each of the provided node names actually refers to a real existing node and that the node belongs to the specified cluster requires a check of the inventory - this is done in the gather logs script itself (see the referenced PR). The validation here is only a sanity check to verify that it matches "something that could be in the host inventory".

The (?!-) and (?<!-) check that the next and previous character are not "-" respectively, without actually "consuming" a character match, i.e. lookahead/lookbehind checks This prevents node names like -this.one or that-.other. It's probably not necessary (argparse should complain with the former, inventory lookup would fail on the latter) so I could just remove them if they confuse more than help...

scripts/cicd/verify-gather-logs-operations.py
codificat: It's using "-n node1 node2" here... I can change the argument to `action=append` for the other syntax "-n node1 -n node2" but I thought that not having to add all the "-n" would save some typing to users...

scripts/cicd/verify-gather-logs-operations.py
dbaker-arch: Got it.  So long as we filter out malicious input, it's ok if we let an IP address in when we really wanted a hostname.  No problems here.

scripts/cicd/verify-gather-logs-operations.py
dbaker-arch: From a security perspective, I'm happy with either.  I feel that "-n node1 node2" will be open to misunderstanding later if "node2" in that example could be considered an appending on the -n list, or an arg in it's own right.

codificat:
I modified the regex to simplify it and make it more clear. Now it won't allow IPs... but looking closer I believe these would not be used, so I hope this looks better now.

I also changed the way usage help is presented. This is a sample of how it looks like now:
```
$ ssh opsmedic@atomic735.example.net 
usage: ssh -i gather_logs_key opsmedic@atomic735.example.net -- -u USER -c CLUSTER [-n node1 node2...]
opsmedic@atomic735.example.net: error: argument -u is required
Connection to atomic735.example.net closed.
```

Messages coming from argument validation errors ("argument -u is required" in the example above) look quite more clean now.

scripts/cicd/verify-gather-logs-operations.py
```diff
@@ -39,13 +40,25 @@
     'starter-us-west-2'
 ]
 
+# This regex matches a valid hostname: an optional set of DNS labels each
```
dbaker-arch: I like this simpler regex.  I would suggest adding to the comment that this matches against the expected format but does not ensure validity.  As such it's only input sanitization, not input validation.  Invalid IP addresses and non-existent hostnames will both be accepted.

dbaker-arch (approved):
I would suggest the one edit to the comment I mention, but aside from this it's +1.

PR #2756, opened 2017-07-27
tiered-access: add 'oc get routes'

joelddiaz:
Right now only non-yaml and non-json output (will need to redact certificate data if we want to whitelist json/yaml route output)
Add the following as default allowed commands:
oc get routes -ndefault
oc get routes -nlogging

Add test case covering 'oc get routes'.

Add ability to override default config file location to make it easier to test by setting DEVACCESS_CONFIG env var.


joelddiaz:
👍 #2734 in stg

PR #2757, opened 2017-07-28
adding a plethora of safe commands

ivanhorvath:


scripts/devaccess/devaccess_wrap.py
```diff
@@ -270,19 +304,51 @@ def rpm_qa(cmd):
...
+            results = subprocess.check_output(cmd.split(), env=our_env)
+        except subprocess.CalledProcessError:
+            log.exception("Call to process (%s) failed", cmd)
 
         return results
 
     @staticmethod
-    def oc_get_pods(occmd):
-        ''' provide 'oc get pods' '''
+    def super_safe_command(cmd):
```
dbaker-arch: Suggest renaming as something like "sudo_safe_command", as "super safe" is too easily confused for the descriptive "very safe".

scripts/devaccess/devaccess_users.yaml
```diff
@@ -1,14 +1,179 @@
...
+        - names:
+            - clusterresourcequota
+
+    - base: rpm
+      runner: safe_command
+      switches:
+        - -qa
+
+    - base: cat
+      runner: super_safe_command
```
dbaker-arch: cat is very much NOT a "safe command" unless filenames remain tightly controlled.  If anything in this system fails in the future, I would put money on unintentional loosening of controls around a command like this - I think it's worth adding comments or other protections to help prevent.

scripts/devaccess/devaccess_users.yaml
```diff
@@ -1,14 +1,179 @@
...
+        - /etc/origin/node/node-config.yaml
+
+    - base: ethtool
+      runner: safe_command
+      switches:
+        - -i
+      leftover:
+        - '^eth[0-9]$'
+        - '^enp\ds\d$'
+    - base: uptime
```
dbaker-arch, line 90: Add whitespace between each block for readability.

scripts/devaccess/devaccess_users.yaml
```diff
@@ -1,14 +1,179 @@
...
+    - base: lsblk
+      runner: safe_command
+      switches:
+        - -at
+        - --output
+      leftover:
+        - NAME,KNAME,FSTYPE,RA,SIZE,SCHED,TYPE
+    - base: lscpu
+      runner: safe_command
+    - base: uname
```
dbaker-arch, line 101: suggest adding: uname -a

scripts/devaccess/devaccess_users.yaml
```diff
@@ -1,14 +1,179 @@
...
+      switches:
+        - -x
+        - -y
+    - base: mpstat
+      runner: cmd_with_tail
+      switches:
+        - -P
+        - ALL
+      leftover:
+        - '^\d \d+$'
```
dbaker-arch: Suggest delay (first value) should accept more than just \d.  A 10 or 30 second delay is appropriate.

scripts/devaccess/devaccess_users.yaml
```diff
@@ -1,14 +1,179 @@
...
+    - base: vmstat
+      runner: cmd_with_tail
+      switches:
+        - -w
+      leftover:
+        - '^\d \d+$'
```
dbaker-arch: Suggest delay (first value) should accept more than just \d.  A 10 or 30 second delay is appropriate.

scripts/devaccess/devaccess_users.yaml
```diff
@@ -1,14 +1,179 @@
...
+        - ALL
+    - base: systemctl
+      runner: safe_command
+      switches:
+        - status
+        - -al
+        - --no-pager
+    - base: ps
+      runner: safe_command
+      switches:
```
dbaker-arch: can the switches be regex'd?  What about auxw?

scripts/devaccess/devaccess_users.yaml
```diff
@@ -18,3 +183,5 @@ users:
   roles:
   - nobody_role
 - username: noroleuser
+- username: ihorvath
```
dbaker-arch: Remove your userid from sample config.

scripts/devaccess/devaccess_wrap.py
```diff
@@ -270,19 +304,51 @@ def rpm_qa(cmd):
...
+        args = cmd.split()
+        os.execvp(args[0], args)
 
-        n_cmd = occmd.normalized_cmd()
-        run_cmd = WhitelistedCommands.oc_cmd_builder(n_cmd)
-        results = subprocess.check_output(run_cmd)
+    @staticmethod
+    def safe_command(cmd):
+        ''' return any safe command output
+            this is for all commands that do not have any type of sensitive data in them
```
dbaker-arch: This line concerns me.  The commands may only not have sensitive data due to careful control of arguments.  It would be useful to have "will never have sensitive data", "should never have sensitive data" (e.g. "ps" listing that may contain a password if someone foolishly includes on the command line), and "we take care to avoid it having sensitive data" (e.g. "cat" where we strongly restrict valid args).

dbaker-arch: This is more aimed at future enhancements, but safe vs non-safe also applies to read or read-write type commands, based on arguments and user.  e.g. "date" can set system clock and break a system if run as root, but usual use is read-only and harmless.

dbaker-arch (changes requested):
Looks good.

I really think "super_safe_command" must be renamed before use.  Everything else is more of a comment for consideration and not a blocker.  

Would like to see test cases for the new code paths you've added to demonstrate accuracy in parsing.

joelddiaz (approved):

ivanhorvath:
[ok to test]

ivanhorvath:
ok to test

PR #2759, opened 2017-07-31
scanpod-inmem: also swallow IOError to handle disappearing processes

joelsmith:


joelsmith:
approved in stg PR #2758 by @drewandersonnz

PR #2761, opened 2017-08-01
delete a router or registry pod when both are running on the same node

stenwt:


openshift_tools/monitoring/ocutil.py
```diff
@@ -69,6 +71,7 @@ def _run_cmd(self, cmd, base_cmd='oc', ):
         if self.logger:
             self.logger.debug("ocutil._run_cmd( {} )".format(cmd))
 
+        # FIXME: pylint redefined-variable-type doesn't like this, but it's working?
```
joelddiaz, line 74: we shouldn't need to disable this pylint. appeasing pylint should be as easy as changing the variable name to 'cmd_split' and changing the two references right below the assignment:

```
cmd_split = shlex.split(cmd)

if self.verbose:
  print "Running command: {}".format(str(cmd_split))

try:
  return subprocess.check_output(cmd_split)
```

joelddiaz (changes requested):

drewandersonnz (approved):
I was going to make some suggestions but I couldn't find a suitable solution to the suggestions.

Things like:
* should a retry be triggered, after delay, when a delete happens?
* should a delete be optional: check_pods method signature change allowing passing in boolean, maybe allowPodDeletion = False .. this would support blocking a delete on a retry
* do we want to delete oldest pods first?
* how to ensure *at least 1* pods are always available and in running+ready state before deleting one?

But then, problems with the suggestions: 
* how long to wait between retries?
* what if pods naturally schedule onto the same node, due to a oversubscription (2 nodes, 3 router pods)?
* deleting oldest pods may cause a rolling restart with outages

Maybe this is over-the-top .. 

+1 to PR, assuming @joelddiaz suggestion for fixing pylint complaint

PR #2767, opened 2017-08-03
adding version check for proper storage api

ivanhorvath:


ivanhorvath:
Approved by @mwoodson in #2766

PR #2768, opened 2017-08-03
testing workaround for >3 items added to the tuple

stenwt:


scripts/monitoring/cron-send-security-updates-count.py
```diff
@@ -82,7 +82,7 @@ def get_security_updates(self):
             self.sec_updates = []
             raise subprocess.CalledProcessError(proc.returncode, subprocess.list2cmdline(cmd), output=stdout+stderr)
         lineslist = [line for line in stdout.split('\n') if line and 'Errno' not in line]
-        self.sec_updates = [UpdateItem(*re.split(r'\s+', line)) for line in lineslist]
+        self.sec_updates = [UpdateItem(*re.split(r'\s+', line)) for line in lineslist if re.split(r'\s+', line) == 3]
```
joelddiaz: This test for filtering away lines that didn't split neatly into 3 fields does mean we could be missing some important information (I know...I know...I recommended the easy fix to begin with :). Maybe it's better to break out this list comprehension into iterating through each line and catching/noticing when a line isn't formatted as expected.

@dbaker-rh thoughts?

dbaker-arch: What output did you see that didn't split into 3 fields?

I recall seeing "Could not connect" type errors.  The concern with ignoring these is that first - if an actual error message DOES split into three fields it could be incorrectly processed.  And second - while I don't fundamentally object to throwing away the details of lines that don't split into three fields, we can't throw away the fact that an unknown error was found in the output.  As such, our list of errata cannot be trusted to be correct and should be flagged as such.

tl;dr - no fundamental objections, but handle the case of "threw away data" as an error in itself to report.

joelddiaz: You get lines like this:

Update notice RHEA-2017:1949 (from rhel-7-server-rpms) is broken, or a bad duplicate, skipping.
You should report this problem to the owner of the rhel-7-server-rpms repository.

In addition to the expected output of:
RHSA-2017:1865              Moderate/Sec.  libX11-1.6.5-1.el7.x86_64

ivanhorvath: We came up with this quick fix as a testing solution to understand the problem. I think the proper way to proceed would be to change the list comprehension into a proper loop where if things fail exceptions can be caught and potentially displayed or logged, then proceed with the other security update detection.

stenwt:
ok to test

scripts/monitoring/cron-send-security-updates-count.py
```diff
@@ -69,6 +69,8 @@ def __init__(self, options):
                 print(err)
             logging.error('Unable to read %s. Continuing.\n%s', wlfile, err)
 
+    #pylint: disable=too-many-branches
+    # fixme: pylint doesn't like the nested ifs here
```
dbaker-arch: Can this be fixed before merge?

scripts/monitoring/cron-send-security-updates-count.py
```diff
@@ -81,8 +83,15 @@ def get_security_updates(self):
...
             self.sec_updates = []
             raise subprocess.CalledProcessError(proc.returncode, subprocess.list2cmdline(cmd), output=stdout+stderr)
-        lineslist = [line for line in stdout.split('\n') if line and 'Errno' not in line]
-        self.sec_updates = [UpdateItem(*re.split(r'\s+', line)) for line in lineslist]
+        for line in stdout.split('\n'):
+            if line and 'Errno' not in line:
+                if len(re.split(r'\s+', line)) == 3:
+                    self.sec_updates.append(UpdateItem(*re.split(r'\s+', line)))
+                else:
+                    logging.error('Security update %s looks like error output', line)
```
dbaker-arch: This'll make for unreadable log output with the unrecognized line stuck in the middle:

logging.error('Security output looks like an error:\n- %s', line)

scripts/monitoring/cron-send-security-updates-count.py
```diff
@@ -81,8 +83,15 @@ def get_security_updates(self):
         if proc.returncode != 0:
             self.sec_updates = []
             raise subprocess.CalledProcessError(proc.returncode, subprocess.list2cmdline(cmd), output=stdout+stderr)
-        lineslist = [line for line in stdout.split('\n') if line and 'Errno' not in line]
-        self.sec_updates = [UpdateItem(*re.split(r'\s+', line)) for line in lineslist]
+        for line in stdout.split('\n'):
+            if line and 'Errno' not in line:
+                if len(re.split(r'\s+', line)) == 3:
+                    self.sec_updates.append(UpdateItem(*re.split(r'\s+', line)))
+                else:
```
joelddiaz: Should the script exit non-zero in this case?

joelddiaz (dismissed):

scripts/monitoring/cron-send-security-updates-count.py
stenwt: we're not supposed to exit non-zero, that hands errors to ops-runner which puts errors on our monitors that often get ignored for long periods (like the one this is fixing). In this case, we're logging that there's bad data, but that's not a fatal case, we can still pass data that we get later up to the monitor.

dbaker-arch (approved):
looks good

joelddiaz (approved):

PR #2770, opened 2017-08-07
add volume status in-use into STABLE_ATTACH_STATUSES

BlueShells:
today I found that the volume status 'in-use' not not in the STABLE_ATTACH_STATUSES list , and it alerting for stucking


drewandersonnz (approved):
+1, "in-use" seems to be a safe/good state

BlueShells:
seems the right status should be "creating | available | in-use | deleting | deleted | error" but what I get from the describe-volume-status is 

"VolumeStatus": {
                "Status": "ok", 
                "Details": [
                    {
                        "Status": "passed", 
                        "Name": "io-enabled"
                    }, 
                    {
                        "Status": "not-applicable", 
                        "Name": "io-performance"
                    }
                ]
            },

PR #2773, opened 2017-08-07
Added openshift_users role.

twiest:


twiest:
This PR was +1d in stg PRs #2771 and #2772.

PR #2776, opened 2017-08-08
Update with new AWS cert

dbaker-arch:
New AWS cert content from: https://signin.aws.amazon.com/static/saml-metadata.xml


blrm (approved):
lgtm

PR #2777, opened 2017-08-08
added when: logic to os_update_latest

yocum137:


ansible/roles/os_update_latest/tasks/main.yml
```diff
@@ -7,3 +7,4 @@
     name:  "{{ oul_package_names }}"
     exclude: "{{ oul_exclude }}"
     state: latest
+    when: os_update is True
```
drewandersonnz, line 10: The `when` should be scoped in the task, it is currently scoped in the `yum` module.

Short answer: align the `when` with the words `name` and `yum` (same level). Basically, remove 2x spaces.

PR #2783, opened 2017-08-09
Set a default to false for generating a default cert for router.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #2782

PR #2790, opened 2017-08-09
adding a plethora of safe commands

ivanhorvath:


ivanhorvath:
Approved by @dbaker-rh and @joelddiaz in #2757

PR #2794, opened 2017-08-09
Exclude yum update when not updating

wshearn:


yocum137 (approved):
it's my change! no one will know!  :-D

PR #2796, opened 2017-08-09
added the vendored stuff, for real:

mwoodson:


mwoodson:
@ivanhorvath approved #2795

PR #2797, opened 2017-08-10
Handle errors from unexpected output in "yum check-update" 

stenwt:


stenwt:
@joelddiaz and @dbaker-rh reviewed #2768

PR #2798, opened 2017-08-11
create aws_service_limit module

blentz:


blentz:
@mwoodson gave :+1: in #2775

PR #2799, opened 2017-08-14
first try dealing with out of disk in zagg-web

stenwt:


stenwt:
This blows, but I haven't found a better, reliable method to kill the pod when we run out of disk space. This at least works.

ivanhorvath:
Hmm, my problem with this is that the scripts can encounter many errors, some that are just normal operation. Zabbix times out on a longer heartbeat request, bam, restart container. A butterfly flaps its wings, restart container, we are going to have 100s of containers per cluster every day. Let me think about something less drastic for a bit.

PR #2800, opened 2017-08-15
[WIP]add auto heal for useless zag-web pod

BlueShells:
add auto heal for useless zag-web pod


stenwt (changes requested):
I'd like to see delete_host_from_zabbix.py in the same PR. 

This looks like it would remove any down host from Zabbix. If, for example, customerfoo-master-dbeef went down, we'd get alerted, but then it would be removed from Zabbix and not alert any further. I don't think that's what we want. Unless there's logic somewhere to only delete zagg-web hosts? I don't see that in this pr.

PR #2801, opened 2017-08-15
obg_kmskeyid is omitted which means it is defined

joelddiaz:
check whether the variable has been omitted in addition to whether it is defined


kwoodson (approved):
I swear we did this.

PR #2802, opened 2017-08-16
obg_kmskeyid is omitted which means it is defined

joelddiaz:
check whether the variable has been omitted


joelddiaz:
👍 #2801 in stg

PR #2804, opened 2017-08-16
add retries to openshift_aws_group to avoid failing due to aws api rate limiting.

blentz:


blentz:
@stenwt gave :+1: in #2803

PR #2806, opened 2017-08-21
Increasting the sensitivy of the docker storage check

mmahut:
Increasing the paging sensitivity to 15%, as we have seen customers seeing warning before hitting the 10% mark.


gburges:
👍

PR #2762, opened 2017-08-02
Adding include_role to best practices.

kwoodson:
Adding best practices doc for include_role.  Here is the easier to read version on my branch:

https://github.com/kwoodson/openshift-tools/blob/doc_update_include_role/docs/best_practices_guide.adoc

PR #2807, opened 2017-08-21
Adding include_role to best practices.

kwoodson:


kwoodson:
@twiest :+1: in #2762

PR #2810, opened 2017-08-22
Add playbook and roles for RHEL 7.4 upgrade

dak1n1:


drewandersonnz (approved):
+1 for testing in int/stg

PR #2815, opened 2017-08-22
renaming script to not cause problems

ivanhorvath:


ivanhorvath:
Approved by @mwoodson in #2814

PR #2819, opened 2017-08-23
adding fixes for older vendored lib_openshifts, so ruamel is not a pr…

ivanhorvath:
…oblem anymore


ivanhorvath:
Approved by @joelddiaz in #2818

PR #2820, opened 2017-08-23
added option to specify package name

dak1n1:
The user can specify any package name to determine if the update was successful. Useful in cases where the system already has the latest kernel, but needs additional updates.


drewandersonnz (approved):
+1

PR #2822, opened 2017-08-29
[DO NOT MERGE] Out of sync commits

wshearn:


wshearn:
@twiest any ideas on what to do? I kinda don't want to recreate the branch as it seems this branch has information missing from the stg branch.

twiest:
@wshearn I honestly have no idea how to fix this other than to start over with a newly created topic branch.

I wouldn't throw this away yet, though. I'd wait to see if it's still a problem with a brand new topic branch. Because if it is, then we have bigger problems.

If you would, please make sure this is known on the "protected branches" e-mail thread as this kind of thing is exactly why we need to protect our upstream branches.

wshearn:
Newly created topic branch with my commit cherry picked, #2833 only 1 change. Going to rename this PR and leave it open until we figure out what to do with it.

PR #2823, opened 2017-08-29
Add starter-us-west-1 to allowed clusters for log gathering

codificat:
@dbaker-rh as per your suggestion in openshift/aos-cd-jobs#555 this is to add starter-us-west-1 to the list of valid clusters to gather logs from.

### DIFF
--- a/scripts/cicd/verify-gather-logs-operations.py
+++ b/scripts/cicd/verify-gather-logs-operations.py
@@ -35,8 +35,10 @@ import logging.handlers
 VALID_CLUSTER_NAMES = [
     'free-int',
     'free-stg',
+    'starter-ca-central-1',
     'starter-us-east-1',
     'starter-us-east-2',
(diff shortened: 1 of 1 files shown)


scripts/cicd/verify-gather-logs-operations.py
```diff
@@ -37,6 +37,7 @@
     'free-stg',
     'starter-us-east-1',
     'starter-us-east-2',
+    'starter-us-west-1',
```
dbaker-arch: We have now also added starter-ca-central-1 to the list of starter clusters.

codificat: ok let me add that one too while we're at it...

codificat:
@dbaker-rh: updated to add starter-ca-central-1 as well

dbaker-arch (approved):
+1

PR #2827, opened 2017-08-30
[DO NOT MERGE] Update for AWS saml certs

dbaker-arch:
Rather than have the AWS saml cert embedded in the j2 file, I've split it out to a local yml file so it's just pulled in when needed).

There's a new command that will download and refresh the yml file that can be run at container creation time; also now run at startup and when secrets change.


docker/oso-saml-sso/centos7/run.sh
```diff
@@ -10,5 +10,5 @@
 
 
 echo -n "Running oso-centos7-saml-sso... "
-sudo docker run -ti --net=host --rm=true --name saml-sso oso-centos7-saml-sso $@
+docker run -ti --net=host --rm=true --name saml-sso oso-centos7-saml-sso $@
```
rhdedgar: This will prevent most people's configurations from running this image locally. Is there a reason it's different from the others or is this leftover config from testing?

docker/oso-saml-sso/src/root/aws_saml_cert.yml
```diff
@@ -0,0 +1,25 @@
+# autogenerated on Wed Aug 30 09:10:29 EDT 2017
+# from https://signin.aws.amazon.com/static/saml-metadata.xml
+---
+aws_saml_subject: /CN=urn:amazon:webservices/O=Amazon Web Services, Inc./ST=Washington/C=US
+aws_saml_expire: 1535155200
+aws_saml_x509: |
```
rhdedgar: Has this already been tested, and still formats the cert data in the same way?

rhdedgar (commented):
The code looks fine, just a couple questions.

docker/oso-saml-sso/centos7/run.sh
dbaker-arch: Thanks, Doug.  Yes, leftover from my testing.  I'll fix.

docker/oso-saml-sso/src/root/aws_saml_cert.yml
dbaker-arch: Yes.

rhdedgar (approved):
👍

dbaker-arch:
PR broke - accidentally pulled in unrelated changes with a minor update.  Will close and re-open.

PR #2834, opened 2017-09-01
Template-ify aws cert

dbaker-arch:
Rather than have the AWS saml cert embedded in the j2 file, I've split it out to a local yml file so it's just pulled in when needed).

There's a new command that will download and refresh the yml file that can be run at container creation time; also now run at startup and when secrets change.

(This is a duplicate PR since the last one was abandoned)


rhdedgar (approved):
👍

PR #2836, opened 2017-09-01
lower severity of EBS volumes stuck in attaching to average for the weekend

blrm:
@robotmaxtron 👍 this change to stg in #2835


blrm:
[test]

PR #2837, opened 2017-09-01
added cron-send-router-reload-time check

stenwt:


scripts/openshift-tools-scripts.spec
```diff
@@ -83,6 +83,7 @@ cp -p monitoring/cron-send-elb-status.py %{buildroot}/usr/bin/cron-send-elb-stat
 cp -p monitoring/ops-ec2-check-tags.py %{buildroot}/usr/bin/ops-ec2-check-tags
 cp -p monitoring/ops-gcp-check-tags.py %{buildroot}/usr/bin/ops-gcp-check-tags
 cp -p monitoring/cron-send-zabbix-too-old.py %{buildroot}/usr/bin/cron-send-zabbix-too-old
+cp -p monitoring/cron-send-router-reload-time.py %{buildroot}/usr/bin/cron-send-router-reload-time
```
joelddiaz, line 86: where is the accompanying piece of the spec file where one of the sub-packages claims /usr/bin/cron-send-router-reload-time?

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -411,6 +411,11 @@ host_monitoring_cron:
   minute: "30"
   job: "ops-runner -f -s 30 -n cosp.openshift.logging.check cron-send-logging-checks"
 
+- name: "Report time to reload haproxy config"
```
joelddiaz: adding the cron job as part of this PR means there is a window of time between merging this PR and building/deploying the new RPMs where this cron job will be in the failing state

scripts/monitoring/cron-send-router-reload-time.py
```diff
@@ -0,0 +1,83 @@
...
+    ms.send_metrics()
+    logger.info("Data sent to Zagg in %s seconds", str(time.time() - ms_time))
+
+def get_router_reload_time():
+    """
+        Return time in seconds it takes to reload HAProxy config
+    """
+
+    start_reload_time = time.time()
+    reload_proxy = ('rsh -n default dc/router "../reload-haproxy"')
```
joelddiaz: this is totally nit-picky (and feel free to disregard), but building out the reload_proxy variable probably should be up right before the start_reload_time so that it isn't part of the time measurement.

joelddiaz (changes requested):

joelddiaz (approved):

PR #2839, opened 2017-09-05
vendor fix again for new vendored 3.6

ivanhorvath:

PR #2841, opened 2017-09-05
vendor fix again for new vendored 3.6

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2839

PR #2845, opened 2017-09-06
[WIP]adding clam update container

rhdedgar:


docker/oso-clam-update/centos7/start.sh
```diff
@@ -0,0 +1,39 @@
...
+if [ "$OO_PAUSE_ON_START" = "true" ] ; then
+  echo
+  echo "This container's startup has been paused indefinitely because OO_PAUSE_ON_START has been set."
+  echo
+  while true; do
+    sleep 10    
+  done
+fi
+
+#echo user:x:$(id -u):0:USER:/root:/bin/bash >> /etc/passwd
```
dbaker-arch: Why is this commented out?

ansible/roles/openshift_clam_update/files/openshift_clam_update.yml
```diff
@@ -0,0 +1,181 @@
...
+      to:
+        kind: ImageStreamTag
+        name: "oso-${PLAT}-clam-update:latest"
+    postCommit: {}
+    resources: {}
+    source:
+      contextDir: "docker/oso-clam-update/${PLAT}"
+      git:
+        ref: clam_update_container
+        uri: https://github.com/rhdedgar/openshift-tools
```
dbaker-arch: Fix this.  Should come from the /openshift/ repo.

ansible/roles/openshift_clam_update/files/openshift_clam_update.yml
```diff
@@ -0,0 +1,181 @@
...
+      to:
+        kind: ImageStreamTag
+        name: "oso-${PLAT}-ops-base:latest"
+    postCommit: {}
+    resources: {}
+    source:
+      contextDir: "docker/oso-ops-base/${PLAT}"
+      git:
+        ref: clam_update_container
+        uri: https://github.com/rhdedgar/openshift-tools
```
dbaker-arch: Should come from /openshift/

ansible/roles/os_zabbix/vars/template_clam_update.yml
```diff
@@ -0,0 +1,14 @@
...
+  zitems:
+  - key: clam.update.signatures.not.updating
+    applications:
+    - Ops Clam Update
+    value_type: int
+
+  ztriggers:
+  - name: "Clam signatures older than 2 weeks"
+    expression: "{Template Ops Clam Update:clam.update.signatures.not.updating.min(#3)}>0"
+    url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/check_clam_update.asciidoc"
```
dbaker-arch, line 13: This link is not valid.

If it was, it should not be pushed into a file present in openshift-tools.

docker/oso-clam-update/centos7/start.sh
```diff
@@ -0,0 +1,42 @@
...
+  while true; do
+    sleep 10    
+  done
+fi
+
+#echo "Running config playbook"
+cd /var/lib/clamav/playbooks
+ansible-playbook /var/lib/clamav/playbooks/config.yml
+
+# Keep running freshclam and clamav-unofficial-sigs every 12 hours
```
dbaker-arch, line 19: This looks like the three steps could get out of step if any fetch step takes longer than expected.

Better, perhaps, is one run-in-loop loop that runs each of (freshclam; unofficial; push to bucket) in sequence.

docker/oso-clam-update/centos7/start.sh
```diff
@@ -0,0 +1,42 @@
...
+# Pause for another 5 minutes and sync new files to the bucket
+sleep 300
+echo
+echo 'Pushing signatures to bucket every 12 hours'
+echo '----------------'
+/usr/local/bin/ops-run-in-loop 43200 /usr/local/bin/push_clam_signatures &>/dev/null
+# Pause for another 5 minutes and send timestamp info to zabbix
+sleep 300
+echo
+echo 'Pushing signatures to bucket every 12 hours'
```
dbaker-arch: Comment doesn't match command.

docker/oso-clam-update/generate-containers.yml
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/ansible-playbook
+---
+- hosts: localhost
+  gather_facts: true
+
+  pre_tasks:
+  - name: Checking build environment
```
dbaker-arch, line 7: checking build references rhel/fedora; but generate containers references rhel/centos.  Is fedora supported? tested?  Is centos supported? tested?

rhdedgar: We only use RHEL/Centos7, both are testing and working. generate-containers.yml is tools-managed and outside the scope of this PR.

docker/oso-clam-update/centos7/start.sh
rhdedgar: Doesn't impact functionality on the current timeline, but this has been updated.

docker/oso-clam-update/centos7/start.sh
rhdedgar: ops-run-in-loop is what we use in place of cron for non-root containers. We'll switch to cron once that's supported.

ansible/roles/os_zabbix/vars/template_clam_update.yml
rhdedgar: Placeholder link. I wasn't going to write the SOP before it was known if the deadline for monitoring could be reached (or if we were even going forward with monitoring).

ansible/roles/openshift_clam_update/files/openshift_clam_update.yml
rhdedgar: Updated.

ansible/roles/openshift_clam_update/files/openshift_clam_update.yml
rhdedgar: Updated.

docker/oso-clam-update/centos7/start.sh
dbaker-arch: I'm not concerned about using ops-run-in-loop itself.  I'm concerned at having multiple copies of it running in parallel in the background when there's a distinct requirement for steps 1 and 2 to finish before step 3 starts.

docker/oso-clam-update/generate-containers.yml
dbaker-arch: Why are there references to fedora if we don't use it?  Can it be removed?

docker/oso-clam-update/centos7/playbooks/playbooks/default_vars.yml
```diff
@@ -0,0 +1 @@
+---
```
blentz, line 1: This can probably be deleted.

docker/oso-clam-update/rhel7/playbooks/playbooks/default_vars.yml
```diff
@@ -0,0 +1 @@
+---
```
blentz, line 1: This can be deleted.

docker/oso-clam-update/src/playbooks/playbooks/default_vars.yml
```diff
@@ -0,0 +1 @@
+---
```
blentz, line 1: This can be deleted.

blentz:
:+1: Seems fine to me.

PR #2850, opened 2017-09-07
updated the verify-cicd-operation script

mwoodson:


mwoodson:
@jupierce approved #2849

PR #2852, opened 2017-09-08
fixed the verify cicd operation

mwoodson:


mwoodson:
@jupierce approved #2851

PR #2855, opened 2017-09-08
added a get-cluster-var to the ohi utility. defaulted to v3

mwoodson:

PR #2857, opened 2017-09-11
added a get-cluster-var to the ohi utility. defaulted to v3

mwoodson:


mwoodson:
@kwoodson approved #2855

PR #2856, opened 2017-09-11
Multi-inventory changes for oo_name

kwoodson:

PR #2858, opened 2017-09-11
Multi-inventory changes for oo_name

kwoodson:


kwoodson:
@joelddiaz :+1:  in #2856

PR #2862, opened 2017-09-11
fixed a v3 vs v2 in ohi

mwoodson:


mwoodson:
@kwoodson approved #2861

PR #2875, opened 2017-09-13
added a fix to get clusters in verify-cicd

mwoodson:


mwoodson:
@jupierce approved #2874

PR #2883, opened 2017-09-14
Template-ify aws cert

dbaker-arch:
Rather than have the AWS saml cert embedded in the j2 file, I've split it out to a local yml file so it's just pulled in when needed.

This was approved in #2834 by @rhdedgar


blrm:
[test]

PR #2887, opened 2017-09-14
changing base image to use 7.3 because 7.4 is broken

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2886

StevenBarre:
This doesn't seem to be enough to fix the issue. Once the build runs `yum update` and gets the latest `yum-uitils` it disables `rhel-7-server-extras-rpms` again and thus can't install Ansible.

I think whatever got changed in https://bugzilla.redhat.com/show_bug.cgi?id=1403015 caused `yum-config-manager --enable` to go wonky.

ivanhorvath:
Ahh, we shouldn't ever run into that scenario because we are strongly against running yum update inside an image. Always better to rebuild the base and then rebuild the other parts on top. In any case, if you provide a PR that solves the issue that you are running into, we will gladly include it so it's in the repo until the fix for yum configs is done in 7.4. Unfortunately we won't be able to devote resources beyond that.

StevenBarre:
> we are strongly against running yum update inside an image

https://github.com/openshift/openshift-tools/blob/prod/docker/oso-ops-base/src/Dockerfile.j2#L46-L47

I think I've got a working fix and will make a PR with it. It seems we just need to run `yum repolist` before each `yum-config-manager` to get it work properly. I'm also adding `rhel-7-server-optional-rpms` repo since that's where `pcp-collector` is hiding.

ivanhorvath:
Well done sir, well done :) I forgot we are doing that at the image creation. But wait, are you saying you're experiencing the same repo loss at that step?

StevenBarre:
Yes, after the yum update, yum-utils is updated and the enabled repos is reset to just the core server one. Then ansible fails to install because its in extras.

PR #2888, opened 2017-09-15
Make `oso-host-monitoring` container compatible with CentOS

mhutter:
**Why is this needed**
CentOS does not have RHSM and as such some directories are missing. Without this change the `config.yml` playbook in the oso-host-monitoring container will fail.

**What does it do**
In order to make `src/root/config.yml` a template, I had to change the `generate_containers` role to support Templates in subdirectories.


ivanhorvath:
You are right, this was done by mistake. We'll check on your PR, we'll have to open it against stg anyway and fix things. Just FYI, we are planning on moving away from this monitoring container, so sooner or later this will be deprecated and there won't be any updates in this code.

docker/roles/generate_containers/tasks/main.yml
```diff
@@ -16,46 +16,32 @@
...
-    path: "{{ item }}"
-  with_fileglob: "{{ base_os }}/*.j2"
-  register: stat_templates
-  when: ansible_version.major > 1
+- name: find templates
+  find:
+    paths: '{{ base_os }}/'
+    patterns: '*.j2'
+    recurse: yes
+  register: find_templates
```
joelddiaz: Why not just register this task as 'template_list' and remove the 'set_fact' that follows?

Other than this one thing LGTM.

joelddiaz (changes requested):

docker/roles/generate_containers/tasks/main.yml
mhutter: Good question... I'll update the PR

mhutter: @joelddiaz fixed

joelddiaz:
@mhutter can you open the PR against the 'stg' branch? (leave this one open)

mhutter:
@joelddiaz #2971

joelddiaz (approved):

joelddiaz:
reviewed #2971 in stg

PR #2891, opened 2017-09-18
stop checking customer pods storage utilisation

drewandersonnz:
master: check all
infra: check all
node: don't check pod pv


gburges:
👍

PR #2896, opened 2017-09-19
added the online-deployer as a valid operation

mwoodson:


mwoodson:
@jupierce approved #2893

PR #2900, opened 2017-09-19
fixing namespace selection

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2899

PR #2901, opened 2017-09-20
Split out resourcequota

yocum137:
split out cluster resource quotas from storageclass role, created a new role
This goes with PR https://github.com/openshift/openshift-ansible-ops/pull/3207


ansible/roles/openshift_master_resource_quota/README.md
```diff
@@ -0,0 +1,45 @@
+openshift_master_storage_class
```
mwoodson: name is wrong here...

ansible/roles/openshift_master_resource_quota/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Requirements
+------------
+
+None
+
+Role Variables
+--------------
+
+osmsc_enable_quotas: BOOL.  Should quotas be enabled.
```
mwoodson: we try to name the role variables based on an abbreviation of the role name.

for openshift_master_storage_class = osmsc.

For this new role, probably want to name it: osmrq?

ansible/roles/openshift_master_resource_quota/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
+---
+- name: Include lib_openshift
+  include_role:
+    name: tools_roles/lib_openshift
+
```
mwoodson: This isn't needed because the dependency is in meta/main.yml.

mwoodson (changes requested):

mwoodson (approved):

PR #2903, opened 2017-09-21
add cloudhealth scripts

blentz:


scripts/cloudhealth/add_billing_role.py
```diff
@@ -0,0 +1,223 @@
...
+import sys
+
+import boto3
+import botocore.exceptions
+import requests.exceptions
+
+from cloudhealth import CloudHealthAwsAccount
+
+LOG_FORMAT = '%(asctime)-15s [%(levelname)s] (%(name)s.%(funcName)s) %(message)s'
+AWS_CREDENTIALS = os.environ['HOME'] + '/.aws/credentials'
```
dbaker-arch: Is this needed?  This is the default AWS_CREDENTIALS value and if someone has altered it, it may be for good reason.

scripts/cloudhealth/README.adoc
```diff
@@ -0,0 +1,11 @@
+This script is used to configure cross-service permissions required to
+allow CloudHealth to collect cost-related data.
+
+It creates AWS IAM Role, Profile, and Role->Profile link. Then, it adds the AWS
+Account info to CloudHealth.
+
+Usage:
+
+        $ export CLOUDHEALTH_API_KEY='some-api-key'
+        $ add-billing-role.py -p AWS_PROFILE -k random-string -n role-name
```
dbaker-arch: Can you define what "random-string" means.  What's it used for?  Why is it random?  Would you re-use one?  If it's truly random, then the script should generate it's own random data (e.g. openssl rand -hex 20) to protect against the user not providing "random enough" input.

dbaker-arch: -n role-name does not match the code below.  There's hardcoded role/oso-billing and "-n" maps to aws_name, not role-name.

scripts/cloudhealth/cloudhealth.py
```diff
@@ -0,0 +1,207 @@
...
+        else:
+            raise Exception('Missing required parameter "name".')
+
+        # pylint: disable=line-too-long
+        if 'authentication' in kwargs.keys() and \
+                'protocol' in kwargs['authentication'].keys():
+            if kwargs['authentication']['protocol'] not in ['access_key', 'assume_role']:
+                raise Exception('Protocol must be one of "access_key" or "assume_role"')
+
+            if kwargs['authentication']['protocol'] == 'access_key':
```
dbaker-arch, line 148: Code above indicates this should only use AWS_PROFILE, not access/secret keys.  Drop support for access/secret on new tools and force AWS_PROFILE usage instead wherever it's possible.

scripts/cloudhealth/add_billing_role.py
```diff
@@ -0,0 +1,223 @@
...
+    ''' parse CLI args '''
+    parser = argparse.ArgumentParser(description='MY SCRIPT DESCRIPTION')
+    parser.add_argument('-e', '--external-id', type=str, dest='aws_external_id',
+                        help="External ID String for AWS cross-account trust")
+    parser.add_argument('-f', '--file', type=str, dest='credsfile',
+                        default='%s/.aws/credentials' % os.environ['HOME'],
+                        help='aws credentials file')
+    parser.add_argument('-k', '--cloudhealth-key', type=str, dest='cht_api_key',
+                        help="The API Key for CloudHealth")
+    parser.add_argument('-n', '--name', type=str, dest='aws_name',
```
dbaker-arch: aws_name is unused in the code

scripts/cloudhealth/add_billing_role.py
```diff
@@ -0,0 +1,223 @@
...
+def setup_cloudhealth_account(api_key, account_number, external_id=None):
+    ''' create an account listing in cloudhealth '''
+    cht = CloudHealthAwsAccount(api_key=api_key,
+                                owner_id=u'%s' % str(account_number))
+    LOG.info('CloudHealth Account: %s', cht)
+    # pylint: disable=line-too-long
+    if not getattr(cht, 'id', None):
+        cht.authentication = {'assume_role_external_id': external_id,
+                              'protocol': 'assume_role',
+                              'assume_role_arn': 'arn:aws:iam::%s:role/oso-billing' % \
```
dbaker-arch: Looks like the oso-billing part is a parameter ("must match the filename") but I don't see that present here.  Should this be "aws_name"?

scripts/cloudhealth/cloudhealth.py
blentz: This is part of the Cloudhealth API, not part of the AWS CLI invocation.

scripts/cloudhealth/add_billing_role.py
blentz: It was in an earlier version of the script. Looks like it is no longer needed.

scripts/cloudhealth/README.adoc
blentz: I think I meant to reference the "-e" flag, not "-k", but either way it can be removed.

blentz: Fixed.

scripts/cloudhealth/add_billing_role.py
blentz: Yes, it should be. Fixing.

dbaker-arch (approved):
Looks like you addressed my concerns.

PR #2915, opened 2017-09-26
fixed some dependencies for lib_openshift roles

mwoodson:

PR #2916, opened 2017-09-26
fixed some dependencies for lib_openshift roles

mwoodson:


mwoodson:
@wshearn approved #2915

PR #2918, opened 2017-09-26
Fix for inventory on v2.

kwoodson:


kwoodson:
@mwoodson :+1:  in #2917

PR #2920, opened 2017-09-26
fixing logic to enable oc logs to work right

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2919

PR #2924, opened 2017-09-27
change name: to gp2 to match openshift-ansible upstream name

yocum137:


mwoodson (approved):

yocum137:
Any cluster that has a storageclass named ebs will need to have that storageclass removed, then run openshift-ansible-ops/playbooks/release/post_byo/storage_class_setup.yml to re-create it with the new name: gp2.

PR #2927, opened 2017-09-28
cicd: added some extra functions

mwoodson:


mwoodson:
@jupierce approved #2926

PR #2936, opened 2017-10-03
changing user role handling around to accomodate overloading commands

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #2930

PR #2937, opened 2017-10-03
Add redacted script that filters yaml config files for developers

wshearn:


wshearn:
[test]

ivanhorvath (approved):

wshearn:
[test]

wshearn:
@dbaker-rh when ever you get a chance, just look this over.

scripts/devaccess/helper_scripts/redacted.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
+"""
+This script will dump config files in a clean format for developer access
+"""
+import os
+import pwd
+import sys
+import yaml
+
+CONFIG_FILE = "/usr/local/bin/redacted.yaml"
```
dbaker-arch, line 10: config files shouldn't go in /usr/local/bin/ - why do we not use /usr/local/etc?

scripts/devaccess/helper_scripts/redacted.py
```diff
@@ -0,0 +1,99 @@
...
+
+    return config_data
+
+
+def main():
+    '''
+    The main entry point
+    '''
+    # Checks to see if we are root, I use pwd instead of just checking 0
+    # Easier to read this way I think
```
dbaker-arch, line 89: I /think/ checking for zero is better for uid.  

Why do we need to run as root?  This will prohibit re-use of this component inside most containers.  Better, I think, to let the script error gracefully if it can't read config files or data for whatever reason rather than an arbitrary uid restriction.

scripts/devaccess/helper_scripts/redacted.py
```diff
@@ -0,0 +1,99 @@
...
+    if os.getuid() != pwd.getpwnam('root').pw_uid:
+        print "You must run this script with sudo"
```
dbaker-arch, line 91: The error text doesn't match the condition being met.  Running with sudo != logging in as root and running it.  Also != running as setuid binary.  For what it's worth, running with sudo is also not always equivalent to running as root.

scripts/devaccess/helper_scripts/redacted.py
wshearn: Because of how this directory works, it just uploads everything to /usr/local/bin

scripts/devaccess/helper_scripts/redacted.py
wshearn: This script does not run in a container, it is so that devs can read system configs on the host. If they have access to exec into the container then they would have full permissions to read the files anyways

scripts/devaccess/helper_scripts/redacted.py
wshearn: It does but also doesn't, if you run with sudo then os.getuid() will return the uid of the user to sudo'd as(and developers will not have access to log in as root). We also do not do setuid on scripts that the developers run so that case is also null and void.

scripts/devaccess/helper_scripts/redacted.py
dbaker-arch: My point is your code isn't able to be re-used.  I won't block the review based on this, but I feel that it warrants being flagged for attention.

scripts/devaccess/helper_scripts/redacted.py
dbaker-arch: Ugly, but if that's what everything else does then at least we're consistent even if we're not correct.

dbaker-arch (approved):

PR #2949, opened 2017-10-04
Remove duplicate snapshot trim cron.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #2947

PR #2957, opened 2017-10-05
prod: Add adhoc playbook docker_storage_to_overlay2.yml

mbarnes:
Wraps the [similarly named role in `openshift-ansible`](https://github.com/openshift/openshift-ansible/pull/5216).

Heavily based on the `os_update.yml` adhoc playbook.

@dak1n1 : I'm not sure if I'm calling the vendored in roles correctly.  Can't seem to find any other examples.  Please check.

### DIFF
--- a/ansible/playbooks/adhoc/docker_storage_to_overlay2/docker_storage_to_overlay2.yml
+++ b/ansible/playbooks/adhoc/docker_storage_to_overlay2/docker_storage_to_overlay2.yml
@@ -1,0 +1,125 @@
+# This playbook wraps the similarly named role in openshift-ansible.
+#
+# It's idempotent and safe to run repeatedly.  If the node's storage driver
+# is already overlay2, the ephemeral storage is left untouched.
+#
+# The role itself wipes all ephemeral storage with "atomic storage reset", so
+# this playbook drains nodes before invoking the role.
+#
+# Usage example:
+#
+#  ansible-playbook docker_storage_to_overlay2.yml -e cli_clusterid=testcluster
+#
+
+- hosts: localhost
+  gather_facts: no
+  become: no
+  user: root
+
+  tasks:
+
+  - name: Check for required variables
+    fail:
+      msg: "Please define {{ item }}"
+    when: "{{ item }} is undefined"
+    with_items:
+    - cli_clusterid
+    run_once: True
+
+##############################################
+# Master upgrade
+##############################################
+
+- hosts: "oo_clusterid_{{ cli_clusterid }}:&oo_hosttype_master"
+  gather_facts: yes  # docker_storage_to_overlay2 needs ansible_distribution
+  become: no
+  user: root
+  serial: 1
+
+  roles:
+  - role: ../../../../openshift/installer/atomic-openshift-3.7/roles/docker_storage_driver
+
+  - role: ../../../roles/openshift_aws_elb_instance_manager
+    osaeim_elb_name: "{{ cli_clusterid }}-master"
+    osaeim_state: absent
+    osaeim_instance_ids: "{{ ec2_id }}"
+    osaeim_region: "{{ oo_sublocation }}"
+    osaeim_aws_access_key: "{{ lookup('env', 'ACCESS_KEY_ID') }}"
+    osaeim_aws_secret_key: "{{ lookup('env', 'SECRET_ACCESS_KEY') }}"
+    when: docker_storage_driver != "overlay2"
+
+  - role: ../../../../openshift/installer/atomic-openshift-3.7/roles/docker_storage_to_overlay2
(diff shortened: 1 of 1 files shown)


mbarnes:
Validation test are going to fail until https://github.com/openshift/openshift-ansible/pull/5216 gets merged and the new roles are vendored in.

ansible/playbooks/adhoc/docker_storage_to_overlay2/docker_storage_to_overlay2.yml
```diff
@@ -0,0 +1,125 @@
...
+    osaeim_aws_access_key: "{{ lookup('env', 'ACCESS_KEY_ID') }}"
+    osaeim_aws_secret_key: "{{ lookup('env', 'SECRET_ACCESS_KEY') }}"
+    when: docker_storage_driver != "overlay2"
+
+  - role: ../../../../openshift/installer/atomic-openshift-3.7/roles/docker_storage_to_overlay2
+    when: docker_storage_driver != "overlay2"
+
+  - role: ../../../roles/os_reboot_server
+    when: docker_storage_driver != "overlay2"
+
```
dak1n1, line 56: Is a reboot needed after updating the storage driver? Or can this role be removed?

ansible/playbooks/adhoc/docker_storage_to_overlay2/docker_storage_to_overlay2.yml
```diff
@@ -0,0 +1,125 @@
...
+- hosts: "oo_clusterid_{{ cli_clusterid }}:&oo_hosttype_master"
+  gather_facts: yes  # docker_storage_to_overlay2 needs ansible_distribution
+  become: no
+  user: root
+  serial: 1
+
+  roles:
+  - role: ../../../../openshift/installer/atomic-openshift-3.7/roles/docker_storage_driver
+
```
dak1n1, line 41: We're going to run this playbook on hosts that are running OCP 3.6 and 3.5 too. Is this role identical in `atomic-openshift-3.7`, `atomic-openshift-3.6` and `atomic-openshift-3.5`?

ansible/playbooks/adhoc/docker_storage_to_overlay2/docker_storage_to_overlay2.yml
mbarnes: Strictly speaking, no.  But `docker-storage-setup.service` mounts the newly-configured volume on each reboot, so I thought it best to leave the reboot there.  I thought if reboot's gonna fail because of these changes, I'd rather we see it immediately.

ansible/playbooks/adhoc/docker_storage_to_overlay2/docker_storage_to_overlay2.yml
mbarnes: Well, it's not in *any* of them yet.  ATM, I have `atomic-openshift-3.7` symlinked to my local `openshift-ansible` repo with my topic branch checked out.

This is the part I guess I'm still confused about how it's supposed to work.

* To begin with, how does one invoke a product-provided role from an `openshift-tools` playbook?

* ~~And if we're invoking a new role that's supposed to come from the product, how do we deal with older versions of the product which lack these roles?  A local copy in `openshift-tools` is the only thing I can think of.~~  Hmm, I think I may have misunderstood what you were asking.  If you meant "does the role behave the same in those environments?" then I believe so, as long as the updated packages we need are available (`docker`, `atomic`, `container-storage-setup`).

dak1n1: Usually we request the role to be backported into all the versions we need. And then we call the role using the version of the installer that matches the version of OCP on a running cluster. So for most of Dedicated, we would call `../../../../openshift/installer/atomic-openshift-3.5/roles/docker_storage_driver` since Dedicated is currently running 3.5. Though since this is a one-time script, we might be able to leave the version hard-coded to 3.7. But we can only do that if we're sure there wont be different versions of this role for each version of OCP.

Once the role is present upstream (in the openshift-ansible repo), we'll vendor the appropriate version(s) into openshift-tools.

ansible/playbooks/adhoc/docker_storage_to_overlay2/docker_storage_to_overlay2.yml
dak1n1: Ok, sounds good.

dak1n1 (approved):
:+1:

mbarnes:
[test]

mbarnes:
Scrapping this PR since I didn't follow the SOP correctly.

PR #2962, opened 2017-10-05
stg: Add adhoc playbook docker_storage_to_overlay2.yml

mbarnes:
Wraps the similarly named role in `openshift-ansible`.

See https://github.com/openshift/openshift-tools/pull/2957 for `prod`.


dak1n1 (approved):
:+1:

PR #2963, opened 2017-10-05
Docker storage to overlay2

mbarnes:
This adds an adhoc playbook to change the Docker storage driver to overlay2.

This time I branched off of `prod`.


dak1n1 (approved):
:+1: We're having some bot issues that caused the extra commits to come along. But this is fine in stg.

PR #2968, opened 2017-10-09
allow world read for the multi_inventory.cache file

joelddiaz:
after review, it has been determined there's no sensitive data in there


joelddiaz:
👍 #2967 in stg

PR #2969, opened 2017-10-10
[WIP]Heartbeat v2 20171010

BlueShells:
remove the times control for how many instance could reboot at 1 hour
remove the get cred part 
remove reboot instance part

### DIFF
--- a/scripts/remote-heal/heal_for_heartbeat.yml
+++ b/scripts/remote-heal/heal_for_heartbeat.yml
@@ -1,0 +1,108 @@
+---
+#prybook will try heal for the hear.beat.ping alert
+# * oso-moniotoring pod dead (restarted)
+# * atomic-node service dead（restarted)
+# * host lost connection (try to restart the instance)
+# * if this is a zag pod , delete it from the zabbix
+
+
+- name: check if the this is zag pod
+  hosts: localhost
+
+  pre_tasks:
+    - name: Ensure needed variables are set
+      fail:
+        msg: "Value {{ item }} is not set!"
+      when: "{{ item }} is undefined"
+      with_items:
+      - cli_nodename
+
+    - name: check if the alert is a zag pod
+      set_fact :
+        zag_pod_or_not : true
+      when: cli_nodename | search("oso-rhel7-zagg-web")
+#     when: "'oso-rhel7-zagg-web' in cli_nodename"
+
+    - name: pull the openshift-ansible-private repo for the cred
+      git:
+        repo: /srv/git/openshift-ansible-private
+        dest: /tmp/auto-heal-repo/openshift-ansible-private
+        update: yes
+
+    - name: pull the private repo for the cred
+      git:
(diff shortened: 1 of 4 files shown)


scripts/remote-heal/heal_for_heartbeat.yml
```diff
@@ -0,0 +1,108 @@
...
+      with_items:
+      - cli_nodename
+
+    - name: check if the alert is a zag pod
+      set_fact :
+        zag_pod_or_not : true
+      when: cli_nodename | search("oso-rhel7-zagg-web")
+#     when: "'oso-rhel7-zagg-web' in cli_nodename"
+
+    - name: pull the openshift-ansible-private repo for the cred
```
joelddiaz, line 26: rather than having to clone all of private for only the zabbix creds, can we just place the zabbix creds in the opsmedic user's home dir somewhere (with appropriate file permissions)?

scripts/remote-heal/heal_for_heartbeat.yml
```diff
@@ -0,0 +1,108 @@
...
+      git:
+        repo: /srv/git/openshift-ansible-private
+        dest: /tmp/auto-heal-repo/openshift-ansible-private
+        update: yes
+
+    - name: pull the private repo for the cred
```
joelddiaz, line 32: why are we cloning the openshift-ansible-ops repo?

scripts/remote-heal/heal_for_heartbeat.yml
```diff
@@ -0,0 +1,108 @@
...
+      git:
+        repo: git@github.com:openshift/openshift-ansible-ops.git
+        dest: /tmp/auto-heal-repo/openshift-ansible-ops
+        update: yes
+
+    - name: pull the openshift-tools to get the lib-zabbix module
```
joelddiaz, line 38: can't we just make sure that the openshift-tools-ansible-zabbix RPM is installed on the system with the opsmedic user?

scripts/remote-heal/heal_for_heartbeat.yml
```diff
@@ -0,0 +1,108 @@
...
+      git:
+        repo: git@github.com:openshift/openshift-tools.git
+        dest: /tmp/auto-heal-repo/openshift-tools
+        update: yes
+
+    - name: set the right hostname to run command on
```
joelddiaz, line 44: i don't see where the target_host var is ever used. is this necessary?

scripts/remote-heal/heal_for_heartbeat.yml
```diff
@@ -0,0 +1,108 @@
...
+      set_fact:
+        target_host : "{{ cli_nodename }}"
+
+    - name: set the right hostname to run command on
```
joelddiaz, line 48: i don't see where the target_host var is ever used. is this necessary?

joelddiaz (changes requested):

PR #2970, opened 2017-10-10
add cloudhealth scripts

blentz:


blentz:
@dbaker-rh gave +1 in #2903

PR #2971, opened 2017-10-10
Make oso-host-monitoring container CentOS compatible

mhutter:
see #2888


joelddiaz:
[ok to test]

joelddiaz:
[test]

joelddiaz:
[test]

joelddiaz:
[test]

joelddiaz:
[test]

blentz:
:+1:  not sure why jenkins isn't showing the right status, but the console output doesn't show any errors. I'm okay merging this.

kwoodson:
[test]

ivanhorvath:
it's not showing because user is not whitelisted and somewhere along the line that is getting checked

kwoodson:
@ivanhorvath,
```
[0mWARN: User mhutter not in admin or org whitelist.
```

Wonder why that would cause it to not report its status though?  Seems like we need to revisit that.

PR #2974, opened 2017-10-11
adding orphaned snapshot delete support

ivanhorvath:

PR #2976, opened 2017-10-12
adding orphaned snapshot delete support

ivanhorvath:


ivanhorvath:
Approved by @twiest in #2974

PR #2977, opened 2017-10-13
Fix ohi -o json slow performance, add fields

joelsmith:
Adds fields hosttype, subhosttype and cluster


ivanhorvath:
We added some changes to the underlying library to streamline accessing the inventory cache some. Also incorporated your changes into the PR, so I'm going to close this.

PR #2980, opened 2017-10-13
delete a router or registry pod when both are running on the same node

stenwt:


stenwt:
Closing this. The logic to determine when it's safe to delete a pod is not fully derived, and about to change dramatically as we provision extra infra nodes. Implementing daemonsets and/or affinity would reduce the volume of these alerts to something a human could deal with, but that's also tabled pending third infra node.

PR #2986, opened 2017-10-16
Removing all groups except all_hosts and oo_

kwoodson:


kwoodson:
@joelddiaz :+1: in #2985

PR #2990, opened 2017-10-17
roles/docker_storage_setup: Support overlay2

mbarnes:
A new variable `dss_docker_storage_driver` determines whether to use the `devicemapper` or `overlay2` storage driver configuration.

For backward-compatibility, the variable defaults to `devicemapper`, but `overlay2` is the new recommendation.


ansible/roles/docker_storage_setup/README.md
```diff
@@ -1,23 +1,30 @@
...
 --------------
 
-dss_docker_device: defaults to /dev/xvdb
+* `dss_docker_device`: defaults to /dev/xvdb
+* `dss_docker_storage_driver`: `devicemapper` (default) or `overlay2`
+* `dss_docker_storage_dm_basesize`: defaults to 3G
+
+For `devicemapper`, `dss_docker_storage_dm_basesize` applies to
+the `dm.basesize` storage option. For `overlay2`, it applies to
+the `overlay2.size` storage option.
```
twiest: Hmm. I know you're re-using `dss_docker_storage_dm_basesize` for backwards compatibility here, but I'm leaning towards having separate vars, one for dm and one for overlay2.

So basically, if you pass in `overlay2` for `dss_docker_storage_driver` then you should also pass in a separate variable to say the size restriction for overlay as well.

Thoughts?

ansible/roles/docker_storage_setup/tasks/main.yml
```diff
@@ -1,17 +1,19 @@
...
   template:
-    src: docker-storage-setup.j2
+    src: "{{ item }}"
     dest: /etc/sysconfig/docker-storage-setup
     owner: root
     group: root
     mode: 0664
+  with_first_found:
+    - "{{ dss_docker_storage_driver }}_dss.j2"
+    - devicemapper_dss.j2
```
twiest: Is there a valid case where the first item in this list isn't the correct one?

In other words, if they pass in `foo` for `dss_docker_storage_driver`, I think I would prefer that it to error here rather than take `devicemapper_dss.j2`.

Thoughts?

twiest (changes requested):

ansible/roles/docker_storage_setup/tasks/main.yml
mbarnes: Agreed, will fix.

ansible/roles/docker_storage_setup/README.md
mbarnes: I was working off the assumption we want the same container size limit for either driver, so I wanted a single default value to cover both.

But I take your point.  A separate variable would be more readable when invoking the role for overlay2, and would better future-proof the role.

mbarnes:
@twiest :arrow_up: Addressed your review comments.

ansible/roles/docker_storage_setup/tasks/main.yml
```diff
@@ -1,17 +1,18 @@
...
+- name: Copy the docker-storage-setup config file
   template:
-    src: docker-storage-setup.j2
+    src: "{{ item }}"
     dest: /etc/sysconfig/docker-storage-setup
     owner: root
     group: root
     mode: 0664
+  with_first_found:
+    - "{{ dss_docker_storage_driver }}_dss.j2"
```
twiest: Is there a reason to use `with_first_found` here?

If we only have 1 in the list, why not just put this as the `src` to the module?

The `template` module will error out if it doesn't find this file as the `src`.

twiest (changes requested):

ansible/roles/docker_storage_setup/tasks/main.yml
mbarnes: Just a vestige of having a fallback file there.  I'll fix.

Good to merge after that?

twiest (approved):
Looks great!

mbarnes:
I merged this manually since I had a couple fixup commits to squash.

PR #2992, opened 2017-10-17
forgot to change these calls after altering the library

ivanhorvath:


ivanhorvath:
Approved by @kwoodson  in #2991

PR #2995, opened 2017-10-18
adding another exception catch so the execution continues

ivanhorvath:


ivanhorvath:
Approved by @blrm  in #2994

PR #3005, opened 2017-10-19
send only internal or external expirations

stenwt:
We're working on https://trello.com/c/RJAPd2Nv/ and as part of that, we're going to create SNow tickets out of upcoming SSL expiration events. I don't want a ticket for every cert, just internal or external. I've refactored some of the check to do that. I'm thinking about ditching the discovery part, since internal or external would be static, any other thoughts?


gburges:
Looks good, what comes to mind is 2nd router. We get an internal SNOW ticket, and another let's say 2 weeks later. We should change our policy to renew all certs at the end of the first contract (we're getting back the money for unused time, right?) in case of a secondary router or push the costs on the customer. That would solve it.

Otherwise looks good 👍

PR #3007, opened 2017-10-23
Auto-heal for heartbeat.ping

BlueShells:
this heal will try to 
1:remove the useless zag-pod from zabbix 
2:try to restart the monitor container service
3:try to restart the node service 

as @joelddiaz suggested , change the method that get the cred for zabbix

### DIFF
--- a/scripts/remote-heal/heal_for_heartbeat.yml
+++ b/scripts/remote-heal/heal_for_heartbeat.yml
@@ -1,0 +1,79 @@
+---
+#prybook will try heal for the hear.beat.ping alert
+# * oso-moniotoring pod dead (restarted)
+# * atomic-node service dead（restarted)
+# * host lost connection (try to restart the instance)
+# * if this is a zag pod , delete it from the zabbix
+
+
+- name: check if the this is zag pod
+  hosts: localhost
+
+  pre_tasks:
+    - name: Ensure needed variables are set
+      fail:
+        msg: "Value {{ item }} is not set!"
+      when: "{{ item }} is undefined"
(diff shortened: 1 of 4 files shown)


scripts/remote-heal/heal_for_heartbeat.yml
```diff
@@ -0,0 +1,84 @@
...
+      with_items:
+      - cli_nodename
+
+    - name: check if the alert is a zag pod
+      set_fact :
+        is_zagg_pod : true
+      when: cli_nodename | search("oso-rhel7-zagg-web")
+
+
+    - name: make sure openshift-tools-ansible-zabbix is installed
```
joelddiaz: doesn't the fact that this file is on the filesystem already mean that openshift-tools-ansible-zabbix is installed (b/c of the RPM dependencies)?

scripts/remote-heal/heal_for_heartbeat.yml
```diff
@@ -0,0 +1,84 @@
...
+
+  tasks:
+
+    - set_fact:
+        oo_instance_id: "{{ hostvars[''+cli_nodename].ec2_id }}"
+      when: is_zagg_pod is not defined
+
+
+    - name: include zabbix info
+      include_vars: /home/opsmedic/zabbix_server.yml
```
joelddiaz, line 38: where is the playbook/role that installs /home/opsmedic/zabbix_server.yml?

joelddiaz (changes requested):

joelddiaz (approved):

PR #3014, opened 2017-10-25
fix the saml check ,change from service to route and add ssl

BlueShells:
fix the saml check ,change from service to route and add ssl


ivanhorvath:
[test]

PR #3027, opened 2017-10-26
gcp lb takes a while to come alive - let's wait 2 minutes and keep trying every 6s

yocum137:


blentz:
:+1:  seems fine, despite misspelling "route"  ;-)

PR #3031, opened 2017-10-27
Making group whitelisting easier.

kwoodson:


kwoodson:
@mwoodson :+1: in #3030

PR #3040, opened 2017-10-31
add project count delta trigger

stenwt:
ref: https://trello.com/c/ERaB2aQk

### DIFF
--- a/ansible/roles/os_zabbix/vars/template_openshift_master.yml
+++ b/ansible/roles/os_zabbix/vars/template_openshift_master.yml
@@ -690,6 +690,15 @@ g_template_openshift_master:
     applications:
     - Openshift Master
 
+  - key: openshift.master.projects.delta
+    description: "Percent change in projects count" 
+    value_type: float
+    zabbix_type: calculated
+    interval: 7200
+    params: "abschange(\"openshift.master.projects.count\") / prev(\"openshift.master.projects.count\") * 100"
+    applications:
+    - Openshift Master
+
   - key: openshift.master.api.process.cpu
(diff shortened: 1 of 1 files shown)


ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -966,6 +975,16 @@ g_template_openshift_master:
     url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/openshift_master.asciidoc"
     priority: high
 
+  - name: ">20% change in project count over 2 hours on {HOST.NAME}"
+    expression: "{Template Openshift Master:openshift.master.projects.delta.last()}>20"
+    url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/delta_trigger.asciidoc"
+    priority: info
+
+  - name: ">60% change in project count over 2 hours on {HOST.NAME}"
```
joelddiaz, line 983: is it 60% or 50%? i see two different values in this trigger

also, should this 50 (or 60) trigger be of priority 'info'?

joelddiaz (changes requested):

ansible/roles/os_zabbix/vars/template_openshift_master.yml
stenwt: fixed, should be 60 in both places. I want "info" for now in case e.g. small dedicated clusters trip this alarm, it's late in the week and I don't want to have to tune it over the weekend. will change to high on Monday.

joelddiaz (approved):

PR #3049, opened 2017-11-04
limit ebs volume reporting to only cluster volumes

joelddiaz:
current version fetches and reports on all volumes in the AWS account/region, so two clusters in the same account/region wold report on each other's volumes.

filter out the non-cluster volumes so we only report on a cluster's volumes

some additional pylint fixes: import ordering and variable re-definition


joelddiaz:
[test]

robotmaxtron (approved):

blrm (approved):

PR #3051, opened 2017-11-06
include cluster name with alert

joelddiaz:
EBS alert would only tell you the volume name previously. Add the host reporting the issue to give the cluster name along with the alert.


joelddiaz:
👍 #3050 in stg

PR #3055, opened 2017-11-06
stop storage alert flapping

joelddiaz:
we have some storage volumes that can hover around 90% storage utilization by design (their high watermark). this means the 90% storage alert can trigger multiple times from 91->89->91 type transitions.

add some hysteresis so that we only clear the storage alert after storage utilization has dropped below 80%


joelddiaz:
👍 #3054 in stg

PR #3057, opened 2017-11-06
changing host monitoring source registry

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #3056

PR #3059, opened 2017-11-07
change the smal pod check method(without disable ssl)

BlueShells:
the new method will check the pod status and pod count to decide whether saml pod is in good status 

this is based on the readinessProbe of this pod , which will do this 

                "readinessProbe": {
                    "exec": {
                        "command": [
                            "/usr/bin/curl",
                            "-k",
                            "https://127.0.0.1:8443/logged_out.html"
                        ]
                    },
                    "failureThreshold": 3,
                    "initialDelaySeconds": 10,
                    "periodSeconds": 10,
                    "successThreshold": 1,
                    "timeoutSeconds": 1

### DIFF
--- a/scripts/monitoring/cron-send-saml-status.py
+++ b/scripts/monitoring/cron-send-saml-status.py
@@ -7,6 +7,7 @@
 
 # pylint: disable=wrong-import-position
 # pylint: disable=broad-except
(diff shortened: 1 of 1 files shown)


drewandersonnz (approved):
+1

PR #3076, opened 2017-11-08
reorder discovery items

joelddiaz:
this allows a clean zabbix container to be configured without erroring out (tested using local development steps)


joelddiaz:
👍 #3074 in stg

PR #3077, opened 2017-11-08
add cgroup memory usage metrics

joelddiaz:
cron-send-cgroup-slice-metrics will send the max memory usage and current memory usage for the system.slice and kubepods.slice cgroup
    
gathering this data over time will allow setting the non-OpenShift and OpenShift memory allocation for the rest of the system


scripts/monitoring/cron-send-cgroup-slice-metrics.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
twiest: consider using `bash -e`

twiest (changes requested):

twiest (approved):
Looks great!

PR #3061, opened 2017-11-07
adding region option to snap creation and trimming

ivanhorvath:

PR #3079, opened 2017-11-08
adding region option to snap creation and trimming

ivanhorvath:


ivanhorvath:
Approved by @twiest in #3061

PR #3081, opened 2017-11-08
Fixed cron-send-docker-metrics and dockerutil to work with overlayfs.

twiest:


twiest:
This PR was +1d in STG PR #3080.

PR #3084, opened 2017-11-08
adding the region to the snapshot related commands in crontab

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #3062

PR #3088, opened 2017-11-08
Fix reporting ES pod disk free

stenwt:
… stats


stenwt:
The current script iterates through each word (from split()) that came back from "df"; if it finds "/elasticsearch/persistent" then it backs up 2 words and returns the value there. I don't like this pattern and am rewriting it.

PR #3091, opened 2017-11-09
add docker and atomic-openshift-node memory reporting

joelddiaz:
change system and kubepods slice reporting to read RSS (to avoid counting page cache)

remove max reporting (we can just view the graph over time)


joelddiaz:
👍 #3090 in stg

PR #3093, opened 2017-11-09
tag centos images so fedora know to push them out to docker hub

joelddiaz:
newer fedora releases don't assume docker hub as the default repo. tag images to make things more explicit.


joelddiaz:
👍 #3092 in stg

PR #3103, opened 2017-11-15
cicd: added repo operations and new vars to be passed

mwoodson:


mwoodson:
@jupierce  I think the regex for these is right.  We need http://.....   Let me know if you think of anything else that needs to be in the regex

jupierce (approved):

PR #3106, opened 2017-11-15
cicd: added repo operations and new vars to be passed

mwoodson:


mwoodson:
@jupierce approved #3104

PR #3107, opened 2017-11-15
Add early log/drop for udp 80 traffic

dbaker-arch:
Adding early log and drop rules within oo-udp to help identify and drop traffic prior to rate limited exceptions for expediency.


rhdedgar (approved):
👍

PR #3108, opened 2017-11-15
Add early log/drop for udp 80 traffic

dbaker-arch:
Was approved by @rhdedgar in #3107 - pending further testing before pushing to prod.


dbaker-arch:
Testing in stg is complete.

PR #3109, opened 2017-11-16
Add reporting to trello script

aweiteka:
Signed-off-by: Aaron Weitekamp <aweiteka@redhat.com>

### DIFF
--- a/openshift_tools/ircbot/trello/trello.py
+++ b/openshift_tools/ircbot/trello/trello.py
@@ -174,8 +174,8 @@ class Trello(object):
             print("Report emailed to {}".format(email))
         if move:
             list_name = "Week #{}".format(week_number)
-            self.move_cards(to_list=list_name)
-            print("Cards moved to list '{}'".format(list_name))
+            if self.move_cards(to_list=list_name):
+                print("Cards moved to list '{}'".format(list_name))
 
     def report_payload(self):
         """Return report payload
@@ -193,12 +193,15 @@ class Trello(object):
             data += "{} {}\n".format(card['shortUrl'], card['name'])
         return data
 
-    def move_cards(self, from_list=None, to_list=None):
+    def move_cards(self, to_list, from_list=None):
         """Move cards from one list to another
-        :param from_list: name of list to move card from
-        :param to_list: name of list to move cards to
+        :param to_list (required): name of list to move cards to
+        :param from_list (optional, use default): name of list to move card from
         :return: None"""
         params = {}
+        if not to_list:
+            print "Cannot move: no destination list provided"
+            return False
         if not from_list:
             from_list = DEFAULT_RESOLVED_LIST
         to_list_id = self.create_list(to_list)
--- a/openshift_tools/ircbot/trello/trello.py
+++ b/openshift_tools/ircbot/trello/trello.py
@@ -3,9 +3,12 @@
 ''' Interact with Trello API as a CLI or Sopel IRC bot module '''
 
 import argparse
+from datetime import date
+from email.mime.text import MIMEText
 import json
 import os
 import re
+import smtplib
 import sys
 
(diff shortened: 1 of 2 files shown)


aweiteka:
[test]

openshift_tools/ircbot/trello/trello.py
```diff
@@ -128,6 +150,97 @@ def get(self):
                                                  members)
         return results
 
+    def report(self, email=None, move=False):
+        """Generate reports"""
+        payload = self.report_payload()
+        print(payload)
+        _env_email = os.environ.get("trello_report_email_addresses", None)
```
wgordon17, line 157: Why duplicate this when it should already be available in `self. email_addresses`?

openshift_tools/ircbot/trello/trello.py
```diff
@@ -128,6 +150,97 @@ def get(self):
...
+        if email:
+            subj = "OpenShift SRE Report for Week #{} (ending {})".format(
+                week_number, date.today().strftime("%d-%m-%Y"))
+            _board_url = "https://trello.com/b/{}".format(
+                os.environ.get("trello_board_id", None))
+            payload = "For more information visit the SRE 24x7 board {}\n\n{}".format(
+                _board_url, payload)
+            self.email_report(email, subj, payload)
+            print("Report emailed to {}".format(email))
+        if move:
```
wgordon17, line 175: You have a lot of `if` blocks without `else` blocks. If you're not printing a reason for the `if` block, it would be hard to troubleshoot later why a specific command isn't working

openshift_tools/ircbot/trello/trello.py
```diff
@@ -128,6 +150,97 @@ def get(self):
...
+            data += "{} {}\n".format(card['shortUrl'], card['name'])
+        return data
+
+    def move_cards(self, from_list=None, to_list=None):
+        """Move cards from one list to another
+        :param from_list: name of list to move card from
+        :param to_list: name of list to move cards to
+        :return: None"""
+        params = {}
+        if not from_list:
```
wgordon17: No check if `to_list` exists. What happens if it tries to `self.create_list(None)`?

wgordon17: Ahh, if I had kept reading...it looks like it just creates a list called "None"?

openshift_tools/ircbot/trello/trello.py
```diff
@@ -128,6 +150,97 @@ def get(self):
                                                  members)
         return results
 
+    def report(self, email=None, move=False):
```
wgordon17, line 153: I wasn't able to find a sopel command that can run a report. Is that intended? It might be useful to run a report on demand? Although I can't say that I know the intended use case =)

aweiteka: This is CLI only right now. Depending on how the use case settles out, I suspect we'll run a timed sopel cmd. Not sure yet.

openshift_tools/ircbot/trello/trello.py
aweiteka: self.email_address is passed in as CLI, which overrides environment variable. CLI args aren't available in sopel use case so we fallback to env var.

openshift_tools/ircbot/trello/trello.py
aweiteka: The non-print if blocks are really just to handle optional CLI args. Without a debug path I don't have a suggestion on improving this.

openshift_tools/ircbot/trello/trello.py
aweiteka: Update pushed to handle this case. Thanks.

wgordon17: :+1: Awesome!

wgordon17 (approved):
LGTM

PR #3112, opened 2017-11-16
mostly for reference, fixed the error that keeps us from reporting df…

stenwt:
… stats

@portante reviewed #3088


stenwt:
Merging this so we can get it in before the freeze, I won't have time to refactor the check before then.

PR #3114, opened 2017-11-16
added docker registry rate limiting to new cluster installs

mwoodson:


mwoodson:
@jupierce approved #3113

PR #3124, opened 2017-11-17
ircbot: move .all-list reply to privmsg

wgordon17:
Signed-off-by: Will Gordon <wgordon@redhat.com>


aweiteka:
@aweiteka +1 PR #3125

PR #3132, opened 2017-11-27
install ntpd too (needed for gcp

mwoodson:


ansible/roles/chrony/tasks/main.yml
```diff
@@ -1,13 +1,17 @@
 ---
-- name: remove ntp package
+- name: ensure chrony and ntp package is installed
```
dbaker-arch: *are* installed.

yocum137 (approved):

ansible/roles/chrony/tasks/main.yml
mwoodson: oh you and your proper English!

PR #3133, opened 2017-11-27
install ntpd too (needed for gcp

mwoodson:


mwoodson:
@yocum137 approved #3132

PR #3139, opened 2017-11-29
update tbyo generator template

mwoodson:


mwoodson:
@kwoodson approved #3138

PR #3144, opened 2017-12-01
added more regions to the copy script

mwoodson:


mwoodson:
@dbaker-rh @blentz approved #3143

PR #3154, opened 2017-12-06
extrack rpms from openshift ansible -- leave the rpms around for now

mwoodson:


mwoodson:
@jupierce approved #3153

PR #3157, opened 2017-12-07
change the app create to the primary master

BlueShells:
for now , it alert on 3 masters , we only need it run on one primary master

@vizakua @gburges


gburges:
👍

PR #3160, opened 2017-12-07
Fixed cron-send-filesystem-metrics so that it sends filesystem usage information for overlayfs volumes.

twiest:


twiest:
Approved in STG PR #3155.

PR #3161, opened 2017-12-07
add python2-cryptography rpm for ansible 2.4

joelddiaz:


joelddiaz:
@kwoodson over-the-shoulder +1

PR #3167, opened 2017-12-11
Adapt ops-docker-storage-reinitialize.yml for overlay

mbarnes:
I got a little ambitious and tried to utilize the platform tooling more.

`"atomic storage reset"` covers several of the manual steps that were being done here.  I tried to comment exactly what the command is doing.

I have some questions, though:

 * I assume the filesystem is being wiped for security reasons? The old code used `dd if=/dev/zero ...`, but a recent email from @stenwt showed him using `wipefs --all ...` so I went with that for both `devicemapper` and `overlay2`.  Is that okay?

* I don't understand why the docker service was being restarted twice (once without the immutable attribute hook and once with).  I simplified that part down to what I think is necessary and it seems to work, but perhaps I've missed some undocumented rationale?

 * `docker-storage-setup` was being started as a service, even though it's just a one-shot that runs the `docker-storage-setup` command.  I changed it to just run the command directly.  I assume this is fine unless there was some specific reason it was started as a service.

Tested this on both `devicemapper` and `overlay2` backends on my test cluster, but not on a real production node yet.

### DIFF
--- a/ansible/playbooks/adhoc/docker_storage_reinitialize/ops-docker-storage-reinitialize.yml
+++ b/ansible/playbooks/adhoc/docker_storage_reinitialize/ops-docker-storage-reinitialize.yml
@@ -19,7 +19,7 @@
 #     ops-docker-storage-reinitialize.yml -e cli_name=opstest-node-compute-s3g5s
 
 - name: Reinitialize Docker Storage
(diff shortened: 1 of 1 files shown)


stenwt (approved):
lgtm, thanks!

PR #3169, opened 2017-12-11
fixing some new openshift-3.8 vendoring issues

mwoodson:


mwoodson:
@jupierce approved #3169

PR #3170, opened 2017-12-12
Fix to permit closing of incidents.

dbaker-arch:
Handle presumed data structure change of incidents; permit proper matching of past incidents in order to update or close them.


dbaker-arch:
Automated tests error is on this line:
E:526,24: Non-iterable value self.incidents is used in an iterating context (not-an-iterable)

    def find_incident(self):
...
        for incident in self.incidents:

dbaker-arch:
[test]

kwoodson (approved):

PR #3172, opened 2017-12-12
renamed filter plugins to have different names. needed for 2.4

mwoodson:


mwoodson:
@jupierce approved #3171

PR #3174, opened 2017-12-12
Fix to permit closing of incidents.

dbaker-arch:
Approved by @kwoodson in #3170


dbaker-arch:
Tested in stg successfully.

PR #3176, opened 2017-12-12
vendored 3.8 openshift-ansible

mwoodson:


mwoodson:
@jupierce approved #3176

PR #3185, opened 2017-12-13
workflow updates

rhdedgar:
Bringing openshift-tools code up to date with the current state of the openshift-image-inspector image.


docker/oso-image-inspector/centos7/root/config.yml
```diff
@@ -17,10 +17,10 @@
...
 #      job: "/usr/local/bin/cron-in-memory-scan.sh &>/dev/null"
 
-    - name: upload_scanlogs
-      minute: "15"
-      hour: "14"
-      job: "/usr/local/bin/upload_scanlogs &>/dev/null"
+#    - name: upload_scanlogs
+#      minute: "15"
+#      hour: "14"
+#      job: "/usr/local/bin/upload_scanlogs &>/dev/null"
```
dbaker-arch: Why commented out?  Would prefer upload_scanlogs.sh if it's a shell script to make that more clear.

docker/oso-image-inspector/centos7/root/config.yml
```diff
@@ -53,14 +53,11 @@
...
 
-  - name: print nodename
-    debug:
-      msg: "{{ temp_name[4].split('=')[1] }}"
-
   - name: populate templated config file
     copy:
       content: |
         node_hostname: {{ temp_name[4].split('=')[1] }}
+        cluster_name: {{ temp_name[4].split('=')[1].split('-node')[0] }}
```
dbaker-arch, line 55: This is fragile and liable to break in dedicated with random cluser names.  Can you use {{ oo_clusterid }} instead?

docker/oso-image-inspector/centos7/scripts/cron-scan-pods.sh
```diff
@@ -2,4 +2,6 @@
 
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
-for line in $(chroot /host /usr/bin/docker ps -q); do echo "$line" && image-inspector -scan-type=clamav -clam-socket=/host/run/clamd.scan/clamd.sock -container="$line" -post-results-url http://localhost:8080; done
+for line in $(chroot /host /usr/bin/docker ps --format 'table {{.ID}} {{.Image}} {{.Command}} {{.CreatedAt}} {{.Status}} {{.Ports}} {{.Label "io.kubernetes.pod.namespace"}} {{.Names}}' | awk '{if (NR>1 && $NF !~ /^k8s_POD/ && $(NF-1) !~ /^openshift-/ && $(NF-1) !~ "logging" && $NF !~ /^oso-rhel7-host-monitoring/) print $1;}'); do image-inspector -scan-type=clamav -clam-socket=/host/run/clamd.scan/clamd.sock -container="$line" -post-results-url http://localhost:8080; done
```
dbaker-arch, line 5: I think this is much too confusing for a one-liner.  Too hard to understand and debug.  Can you split it up into a multi-line block instead?

docker/oso-image-inspector/centos7/scripts/cron-scan-pods.sh
```diff
@@ -2,4 +2,6 @@
...
+
+/usr/local/bin/upload_scanlogs
```
dbaker-arch, line 7: This corresponds to the commented out cron above, I think.  Suggest you delete, not comment out, above if so.  Same comment about naming ".sh".

dbaker-arch: Oh ... I mean same comment about naming ".py" now I see the script.

dbaker-arch (approved):
Approved, optionally go ahead and address the couple of comments inline if you wish.

PR #3186, opened 2017-12-13
ircbot: localize shift change times instead of using UTC in topic

wgordon17:
Signed-off-by: Will Gordon <wgordon@redhat.com>


aweiteka:
@aweiteka +1 PR#3187

PR #3199, opened 2018-01-05
updated the vendor openshift ansible role to support 3.8 and also support previous versions

mwoodson:


mwoodson:
@jupierce approved #3198

PR #3203, opened 2018-01-11
Fix logging dynamic issue

BlueShells:
changing the script from using #OSO_ELASTIC to #OSO_METRICS
and add trigger when es cluster status is not green


drewandersonnz (approved):
+1

PR #3214, opened 2018-01-11
vendored o-a-3.7.14

mwoodson:


mwoodson:
@yocum137 approved #3213

PR #3216, opened 2018-01-12
update to support cloudhealth's external id generator

blentz:


blentz:
@stenwt gave :+1: in #3217

PR #3230, opened 2018-01-16
lftp to nodes, because it's useful

gburges:


gburges:
@mmahut 👍 https://github.com/openshift/openshift-tools/pull/3229

PR #3246, opened 2018-01-18
use echo instead of tee in Docker.j2

yocum137:


drewandersonnz (approved):
+1, maybe use COPY instead of echo on line 46 .. not too important though.

PR #3247, opened 2018-01-18
srebot: Only warn once per nick when using .all/all: that .msg is a less noisy option

wgordon17:


aweiteka:
+1 @aweiteka PR #3248

PR #3251, opened 2018-01-18
Stop managing fullName for openshift users. The IDP fills this in, an d we're not asking for it when we grant tiered access.

twiest:


twiest:
This was +1d in STG PR #3250.

PR #3262, opened 2018-01-23
add monitors for pending tasks and active primary shards. more can be…

stenwt:
… added easily from the cluster_health dict


drewandersonnz (approved):
+1

PR #3265, opened 2018-01-24
Logging tune triggers

stenwt:


stenwt:
@robotmaxtron reviewed #3264

PR #3270, opened 2018-01-25
added the docker options to the byo gen template

mwoodson:


mwoodson:
@ivanhorvath approved #3269

PR #3286, opened 2018-01-30
check df with -c so we get info from ocp 3.7+, avoid divide by 0

stenwt:


portante (dismissed):
Can you add the `-c elasticsearch` to all the other `oc exec` calls that don't have it as well?

portante (approved):

PR #3291, opened 2018-02-01
Adding package to tower package list.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #3290

PR #3292, opened 2018-02-05
Adding scale group monitoring to the monitoring role.

kwoodson:

PR #3295, opened 2018-02-06
Adding scale group monitoring to the monitoring role.

kwoodson:


kwoodson:
@mwoodson  :+1:  in #3292

PR #3301, opened 2018-02-06
Updating variable to point at correct roles.

kwoodson:


kwoodson:
@mwoodson :+1:  in #3300

PR #3302, opened 2018-02-07
rkhunter is stopping our monitoring from starting

drewandersonnz:
* separate rkhunter to rkhunter.yml
* execute by ops-runner to track result in zabbix
* add timeout to ops-runner to ensure script doesn't take too long
* execute configuration in background, allowing container to start crond

Note:  `oc adm` snuck into this due to someone not running `generate-containers.yml`, but I'm ok with the change.

### DIFF
--- a/docker/oso-host-monitoring/src/root/rkhunter.yml.j2
+++ b/docker/oso-host-monitoring/src/root/rkhunter.yml.j2
@@ -1,0 +1,97 @@
+{{ generated_header }}
+---
+- hosts: localhost
+  gather_facts: no
+  vars:
+{% raw %}
+    rkhunter_cron:
+    - name: check_rkhunter
+      minute: "30"
+      hour: 21
+      job: "/usr/bin/cron-send-rkhunter-checks"
+
+    - name: rkhunter_scan
+      minute: "0"
+      hour: "21"
+      job: "chroot /var/local/rkhunter_chroot /tmp/rkhunter/bin/rkhunter --sk --check --configfile /tmp/rkhunter/etc/rkhunter.conf --dbdir /tmp/rkhunter/db --tmpdir /tmp/rkhunter >/dev/null 2>&1"
+
+  pre_tasks:
+
+  - name: Setup Cron
+    cron:
+      name: "{{ item.name }}"
+      job: "{{ item.job }}"
+      minute: "{{ item.minute | default('*', True) }}"
+      hour: "{{ item.hour | default('*', True) }}"
+      day: "{{ item.day | default('*', True) }}"
+      month: "{{ item.month | default('*', True) }}"
+      weekday: "{{ item.weekday | default('*', True) }}"
+    with_items:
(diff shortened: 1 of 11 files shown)


dbaker-arch:
Looks reasonable to me.  What sort of testing have you done on the change?

docker/oso-host-monitoring/src/start.sh
```diff
@@ -17,6 +17,10 @@ source /root/.bashrc
 # Configure the container
 time ansible-playbook /root/config.yml
 
+# Configure rkhunter
+echo '/usr/bin/ops-runner -f -t 60 -n monitoring.root.rkhunter.yml ansible-playbook /root/rkhunter.yml >> /var/log/rkhunter.yml.log &'
+/usr/bin/ops-runner -f -t 60 -n monitoring.root.rkhunter.yml ansible-playbook /root/rkhunter.yml >> /var/log/rkhunter.yml.log &
```
rhdedgar: We might not see much value from alerting based on this. I think the timeout on this should be increased or removed because just the ["Set initial values for rkhunter to use"](https://github.com/openshift/openshift-tools/pull/3302/files/#diff-251e4637cf03704fc835c6c60a7df056R72) and ["Run a no-log scan to get copies of passwd and group"](https://github.com/openshift/openshift-tools/pull/3302/files/#diff-251e4637cf03704fc835c6c60a7df056R74) tasks take around 2.5 minutes to complete on a fairly quiet node.

docker/oso-host-monitoring/src/root/rkhunter.yml.j2
```diff
@@ -0,0 +1,97 @@
+{{ generated_header }}
```
rhdedgar, line 1: If we wanted to be real consistent, we could call this file rkhunter_config.yml to match the usual config.yml playbook, but it's fine as-is.

rhdedgar (changes requested):
+1 if we can set the ops-runner timeout higher or skip alerting altogether.

docker/oso-host-monitoring/src/root/rkhunter.yml.j2
dbaker-arch: Very much agree to this.  Name this rkhunter_config.yml, and rename the other one to FOO_config.yml (for an appropriate value of foo).

dbaker-arch:
Tangential request -- one git commit for the actual change, a second git commit for the output of generate-containers in order to better isolate the change from the two copies for rhel/centos.

docker/oso-host-monitoring/src/start.sh
drewandersonnz: The timeout is intended to bring visibility on a failed start. At the moment we have failed starts with no visibility for very long periods of time (one problem has been ongoing for the last 28 days).

I will update from 60s to 600s. If it takes longer than this then we can see which nodes are affected and extend accordingly.

It is more important that we have some monitoring quickly, than none like we have today.

PR #3303, opened 2018-02-07
rkhunter is stopping our monitoring from starting

drewandersonnz:
stg PR here: https://github.com/openshift/openshift-tools/pull/3302

=====

* separate rkhunter to rkhunter.yml
* execute by ops-runner to track result in zabbix
* add timeout to ops-runner to ensure script doesn't take too long
* execute configuration in background, allowing container to start crond

Note:  `oc adm` snuck into this due to someone not running `generate-containers.yml`, but I'm ok with the change.


drewandersonnz:
[test]

PR #3311, opened 2018-02-09
rename registry.ops.openshift.com to registry.reg-aws.openshift.com

yocum137:


dak1n1 (changes requested):
`registry.reg-aws.openshift.com` needs to be `registry.reg-aws.openshift.com:443` instead, since that's the hostname we use when we do a docker login. It's a remnant of an old bug. (The alternative would be to migrate all of production to use `registry.reg-aws.openshift.com`).

yocum137:
Port 443 added.

dak1n1 (approved):
:+1:

PR #3312, opened 2018-02-12
do the pull before the kill

drewandersonnz:
keep the container alive until the very last second before switching to
newer version


blentz:
:+1:  lgtm

PR #3314, opened 2018-02-12
alert on >10% hosts not ready as high, 1 as average

stenwt:


robotmaxtron (commented):
Would this replace our normal 'Host is not ready' high alert?

robotmaxtron (approved):

PR #3315, opened 2018-02-12
do the pull before the kill

drewandersonnz:
+1 from @blentz on https://github.com/openshift/openshift-tools/pull/3312

=====

keep the container alive until the very last second before switching to
newer version


drewandersonnz:
tested in stage, config loop makes this change and daemon-reloads systemd, everythings looking good

PR #3317, opened 2018-02-13
Fixing format to json

kwoodson:


kwoodson:
@mwoodson :+1: in #3316

PR #3324, opened 2018-02-15
changing the redis data folder so server can start in zagg container

ivanhorvath:


ivanhorvath:
Approved by @drewandersonnz in #3323

PR #3331, opened 2018-02-22
Add udp/31685 and tidy up other rules

dbaker-arch:


rhdedgar (approved):
We can always add the deleted comments to an SOP if needed. :+1: to the new port.

PR #3338, opened 2018-02-27
Adding autokeys and yum repositories.

kwoodson:


kwoodson:
@joelddiaz :+1:  in #3337

PR #3340, opened 2018-02-27
added scalegroup operation to verify cicd; did some pylint cleanup in ohi

mwoodson:


mwoodson:
@jupierce approved #3339

PR #3343, opened 2018-02-28
Adding back files which were moved.

kwoodson:


kwoodson:
@mwoodson :+1:  in #3341

PR #3345, opened 2018-03-01
verify cicd: updated variable name

mwoodson:


mwoodson:
@jupierce approved #3344

PR #3358, opened 2018-03-05
ansible api changed, so we must change too

ivanhorvath:


ivanhorvath:
Approved by @kwoodson  in #3357

PR #3361, opened 2018-03-06
start pmcd in container without checking for other running instances

drewandersonnz:
original script source
https://github.com/redhat-developer/osd-monitor-poc/blob/master/pcp-node-collector/run-pmcd.sh

Testing method:

```
# pick a node where it is failing

docker exec -it oso-rhel7-host-monitoring bash

... in docker exec ...

git clone https://github.com/drewandersonnz/openshift-tools.git;
cd openshift-tools;
git checkout pmcd-start;
docker/oso-host-monitoring/centos7/root/start-pmcd.bash;

... <another docker exec instance> ...

run scripts which rely on PCP data and confirm validity

```
Note: ignoring @blrm Merge to stg.


BlueShells:
👍

zhiwliu:
:+1:

PR #3382, opened 2018-03-12
adding variable checks to the start so it fails early

ivanhorvath:


ivanhorvath:
Approved by @mwoodson in #3381

PR #3383, opened 2018-03-15
[wip]Add dynamic item for unresolved banner

BlueShells:
Add dynamic item for unresolved banner

### DIFF
--- a/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
+++ b/ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
@@ -94,6 +94,7 @@ metric_sender_config:
     - Template Config Loop      # So we can send config loop related metrics
     - Template Zabbix Config    # So we can send Zabbix Config metrics
     - Template Multi-Inventory  # So we can send multi-inventory failures
+
 {% endif %}
 
     hostgroups:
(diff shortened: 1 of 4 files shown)


scripts/monitoring/cron-send-status-banner-check.py
```diff
@@ -0,0 +1,126 @@
...
+
+        try:
+
+            result = urllib2.urlopen(request)
+
+            result_d = result.read()
+
+            data = json.loads(result_d)
+
+        except:
```
drewandersonnz: Don't try to hide all failures, we may never know if the script is working or not if it always exits with 0

Allow an exception to be raised

```
Except Exception as e:
  logger.exception("something wrong when trying get the banner info") 
  raise e
```

scripts/monitoring/cron-send-status-banner-check.py
```diff
@@ -0,0 +1,126 @@
...
+
+            data = json.loads(result_d)
+
+            # do the loop to check the component name
+            for component in data:
+                if component['id'] == groupid:
+                    group_name = component['name']
+
+
+        except:
```
drewandersonnz: again, allow exceptions to be raised

ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -193,6 +194,12 @@ host_monitoring_cron:
 - name: run docker dns test
   minute: "*/10"
   job: source $HOME/.bashrc; ops-runner -f -s 60 -n csddr.docker.dns.resolution /usr/bin/cron-send-docker-dns-resolution
+
+
+- name: check status for unresovled banner
+  minute: "*/10"
+  job: pageid={{ g_status_banner_pageid }};authid={{ authid=g_status_banner_authid  }} ops-runner -f -s 60 -n cssbc.status.unresolved.banner  /usr/bin/cron-send-status-banner-check
```
drewandersonnz, line 201:
```
 +  job: 
pageid={{ g_status_banner_pageid }};
authid={{ authid=g_status_banner_authid  }} 
ops-runner -f -s 60 -n cssbc.status.unresolved.banner  /usr/bin/cron-send-status-banner-check 
```

Is `authid=g_status_banner_authid` the correct instantiation?
Also is: `pageid={}; authid={}` using the semicolon the right invocation? I haven't tested this part myself yet.

drewandersonnz (changes requested):
I think this script should allow exceptions to be raised, that way the failure will be tracked in ops-runner.

Also, please confirm the environment variables are correctly set up.

PR #3399, opened 2018-03-20
docker storage set: allow package to be passed in

mwoodson:


mbarnes:
:+1: That looks nice and simple.

PR #3402, opened 2018-03-22
removing cruft from zabbix-web container, so it builds

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #3401

PR #3408, opened 2018-03-26
vendored first 3.10 o-a

mwoodson:


mwoodson:
@jupierce approved #3407

PR #3411, opened 2018-03-28
Add adhoc/docker_storage_to_overlay2.yml playbook

mbarnes:
The playbook itself is mostly a wrapper to drain nodes.

There's a couple new roles here too:

- `docker_storage_driver` is trivial, just queries `docker info`.
- `docker_storage_to_overlay2` does the actual storage conversion + reboot.

@jupierce indicated this is all still working as of today.

I actually merged the playbook to **stg** last October, but never to **prod**.  At the time it was still referencing the roles in my `openshift-ansible` branch.  I was waiting for that to get accepted before merging to **prod**, but that never happened.

### DIFF
--- a/ansible/roles/docker_storage_to_overlay2/meta/main.yml
+++ b/ansible/roles/docker_storage_to_overlay2/meta/main.yml
@@ -1,0 +1,3 @@
+---
+dependencies:
+- tools_roles/lib_repoquery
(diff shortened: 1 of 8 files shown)


mwoodson (approved):

ansible/roles/docker_storage_to_overlay2/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+dependencies:
+- tools_roles/lib_repoquery
+- ../../../openshift/installer/vendored/openshift-ansible-3.7.0/roles/openshift_facts
```
mwoodson: this should reference atomic-openshift-3.7, not 3.7.0.

PR #3415, opened 2018-03-29
generic url checker script

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #3414

PR #3422, opened 2018-04-02
fixing zabbix templates from stg-to-prod mess

ivanhorvath:


ivanhorvath:
Approved by @drewandersonnz in #3421

PR #3430, opened 2018-04-04
final step for web checks

ivanhorvath:


ivanhorvath:
Approved by @blrm in #3429

PR #3437, opened 2018-04-05
adding a check so if file not present it doesn't fail

ivanhorvath:


ivanhorvath:
Approved by @blrm in #3436

PR #3439, opened 2018-04-06
forgot the host name from alert, hard to know where it came from with…

ivanhorvath:
…out it


ivanhorvath:
Approved by @joelddiaz in #3438

PR #3440, opened 2018-04-06
inventory utils and python libs for python version of cicd-control

mwoodson:


mwoodson:
@jupierce approved #3404

PR #3449, opened 2018-04-06
added sleep values for the snapshotter

mwoodson:

PR #3450, opened 2018-04-06
added sleep values for the snapshotter

mwoodson:


mwoodson:
@twiest and @BigN approved #3449

PR #3455, opened 2018-04-09
made the playbook executor more verbose

mwoodson:


mwoodson:
@jupierce approved #3453

PR #3458, opened 2018-04-11
the old way was right

ivanhorvath:

PR #3459, opened 2018-04-11
the old way was right

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #3458

PR #3461, opened 2018-04-11
made some unbuffered fixes for the cicd python stuff; added some better printing

mwoodson:


mwoodson:
@jupierce approved #3460

PR #3467, opened 2018-04-12
ossh: if verbose, print out the command

mwoodson:


mwoodson:
@ivanhorvath approved #3466

PR #3468, opened 2018-04-12
fixing bug with all namespaces so any namespace is valid

ivanhorvath:


ivanhorvath:
Approved by @blrm in #3465

PR #3470, opened 2018-04-12
ossh: added verbosity count

mwoodson:


mwoodson:
@kwoodson approved #3469

PR #3474, opened 2018-04-17
inventory utils, added vfull version

mwoodson:


mwoodson:
@jupierce approved #3473

PR #3475, opened 2018-04-17
Allow vendored versions of openshift-ansible to include playbooks/aws

blrm:
Currently we blacklist some files and directories when we vendor openshift-ansible. This removes "playbooks/aws" from that blacklist so that we can use those provisioning playbooks.


yocum137:
vendor-github.yml could almost be deprecated completely.
:-1:

PR #3480, opened 2018-04-17
changing the webconsole check to not alert on flapping

ivanhorvath:


ivanhorvath:
Approved by @mwoodson in #3479

PR #3485, opened 2018-04-19
adding the podpsec to the monitoring container and adding it to cron

ivanhorvath:


ivanhorvath:
Approved by @joelddiaz in #3484

PR #3487, opened 2018-04-19
oops, there was already a script with the same name, renaming to oc-c…

ivanhorvath:
…luster-capacity


ivanhorvath:
Approved by @joelddiaz in #3486

PR #3490, opened 2018-04-20
Fix host monitoring

portante:
When a server is nearing overload, the cron jobs might take a long time
to start up just to be able to check the flock file given the overhead
of python.  If they can't start up quick enough, then they'll just keep
piling up as the cron deamon does not wait for the previous one to
finish.

By using the external flock command, and using /host/run/lock/ for the
location of the lock file, we avoid the python overhead of startup and
place the lock file in local memory.

Note that /host/run is the host's /run tmpfs mounted into the container.
f034be4

### DIFF
--- a/docker/oso-host-monitoring/centos7/container_setup/monitoring-config.yml
+++ b/docker/oso-host-monitoring/centos7/container_setup/monitoring-config.yml
@@ -136,74 +136,74 @@ metric_sender_config:
 host_monitoring_cron:
 - name: send pcp ping every 5 minutes
   minute: "*/5"
-  job: ops-runner -f -s 15 -n cspp.pcp.ping /usr/bin/cron-send-pcp-ping
+  job: flock /host/run/lock/cspp.pcp.ping ops-runner -s 15 -n cspp.pcp.ping /usr/bin/cron-send-pcp-ping
 
 - name: run pcp checks every 5 minutes
   minute: "*/5"
-  job: ops-runner -f -s 15 -n ozpc.send.pcp /usr/bin/ops-metric-pcp-client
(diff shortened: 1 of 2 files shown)


dbaker-arch (approved):
Usually we'd approve changes before they went to stg.  Since this is the merge to prod I'll +1 proviso it was tested there.

joelddiaz:
what happens when/if the host-monitoring container is terminated ungracefully? will some cron jobs refuse to start now that there are previous lock files permanently on the host filesystem?

also, aren't we reserving a minimum amount of cpu resources for openshift/host workloads through kubeletArguments.system-reserved?

portante:
> what happens when/if the host-monitoring container is terminated ungracefully? will some cron jobs refuse to start now that there are previous lock files permanently on the host filesystem?

Yes, good catch.  We can detect that situation in the container start.sh I believe.  I'll propose something.

> also, aren't we reserving a minimum amount of cpu resources for openshift/host workloads through kubeletArguments.system-reserved?

We are, but we've blown way past that due to other factors.

I don't believe in monitoring and logging collectors running as a docker container or under kubernetes, because they are monitoring the very subsystems in which they are running.  When the environment fails, monitoring fails.

Running this as a simple systemd service with the systemd service cron mechanism on the host would be the safest way to make sure monitoring does not affect the environment.

twiest (changes requested):
The noc team, @portante  and I discussed this today and the conclusion is that this won't fix our overall problem and so we're going to punt on evaluating this for now.

I believe what we want is for ops-runner to continue to do the flock, but change where the flock is kept to a ramdisk location.

I would prefer the ramdisk be in the container, if possible. If not, then let's discuss what the next best would be.

portante:
@twiest, performing a flock on ramdisk in a container seems okay.

Closing this since its not considered a concern right now.

PR #3493, opened 2018-04-23
wrong path for cron command

ivanhorvath:


ivanhorvath:
Approved by @blrm in #3492

PR #3503, opened 2018-04-27
we need to not delete certain roles from openshift-ansible when we vendor

mwoodson:


mwoodson:
@jupierce approved #3502

PR #3505, opened 2018-05-03
fixes to the playbook exececutor

mwoodson:


mwoodson:
@jupierce approved #3504

PR #3509, opened 2018-05-08
added verbosity option to ansible playbook executor; added some tools to the Cluster class

mwoodson:


mwoodson:
@jupierce approved #3508

PR #3511, opened 2018-05-09
ansible playbook executor: do verbosity

mwoodson:


mwoodson:
@jupierce approved #3510

PR #3527, opened 2018-05-23
Changed use-tower[12].ops.rhcloud.com and tower.ops.rhcloud.com DNS entries to the new bastion centric names.

twiest:

PR #3528, opened 2018-05-23
Changed use-tower[12].ops.rhcloud.com and tower.ops.rhcloud.com DNS entries to the new bastion centric names.

twiest:


twiest:
This PR was +1d in STG PR #3527. 

Note the comment in that PR on why the bot is failing.

PR #3533, opened 2018-05-24
inventory utils: added a oc_get_md; cleaned up master config

mwoodson:


mwoodson:
@wshearn approved #3532

PR #3544, opened 2018-05-30
change the check from every 5 minutes to every 1 minutes

BlueShells:
change the check from every 5 minutes to every 1 minutes because of the bug of ELB


drewandersonnz (approved):
runs on one master per cluster, effect on zabbix should be minimal.

PR #3546, opened 2018-05-30
Changed metadata for the bastion host rename.

twiest:


twiest:
This PR was +1d in STG PR #3543.

PR #3551, opened 2018-05-30
Fixed a bug with active media strings.

kwoodson:


kwoodson:
@ivanhorvath :+1:  in #3550

PR #3552, opened 2018-05-31
changing logging in snapshot related parts

ivanhorvath:


ivanhorvath:
Approved by @blrm in #3549

PR #3553, opened 2018-05-31
Creating a sosreport streamer script

sferich888:
This should allow/enable sosreports to be run and pulled off of a node directly.

### DIFF
--- a/scripts/devaccess/helper_scripts/sos_streamer
+++ b/scripts/devaccess/helper_scripts/sos_streamer
@@ -1,0 +1,7 @@
+#!/bin/bash
+# run sosreport with supplied case id
+sosreport --case-id=${1} --log-size=100 --batch --quiet --no-report
+
+# stream back the sosreport archive made by sosreport to the user
+cat /var/tmp/*${1}*.tar.xz
+rm ${file}


loleary:
Why aren't we just modifying the existing ceesos script? I do not think creating a new script (something extra to maintain separately) is what we want.

The ceesos script should accept a command-line switch that streams its output rather then attaching it to a case. And then, we should protect against output to stdout/stderr and perhaps capture any unexpected errors in a log file. For example:

```
#/bin/bash
#

_case_id="$1"
if [ -z "$_case_id" ]; then
    _case_id="$(date -u +'%Y%m%dT%H%M%SZ' 2>/dev/null)"
fi
_tmp_dir="$(mktemp -d -p /var/tmp ohdiag-collection-"${_case_id}"-XXXX)"
_log_file="${_tmp_dir}/sosreport-${_case_id}.log"

trap "{ [ -n "${_tmp_dir}" ] && [ -e "${_tmp_dir}" ] && rm -rf "${_tmp_dir}">/dev/null 2>&1; }" EXIT

if [ -n "${_tmp_dir}" ] && [ -e "${_tmp_dir}" ]; then
    {
        echo "sosreport generation executing for case_id=${_case_id} on $(date)..."
        sosreport --case-id="${_case_id}" --log-size=100 --batch --quiet --no-report --tmp-dir="${_tmp_dir}"
        result=$?
        if [ $result -eq 0 ]; then
            echo "sosreport was successful"
        else
            echo "sosreport report failed with exit status ${result}"
        fi
    } >>"${_log_file}" 2>&1
    (cd "${_tmp_dir}" && tar cf - .) | cat
fi
```

We would of course want to modify the `(cd "${_tmp_dir}"...) | cat` line to check the switch to see if it should stream to stdout or attach to a case.

sferich888:
@loleary can you file a PR for your suggestions against https://github.com/openshift/openshift-tools/blob/prod/scripts/devaccess/helper_scripts/ceesos I want to make sure that we start the discussion about enabling this access for CEE sooner rather than later, having the PR is the first step in that process. Once we have that PR we can close this one.

sferich888:
@loleary I am closing this as I have taken your suggestion and provided an update to the current script.

PR #3555, opened 2018-05-31
guess not every cron syntax is usable

ivanhorvath:


ivanhorvath:
Approved by @blrm in #3554

PR #3557, opened 2018-05-31
changed the verify cicd to send starter clusters to python version

mwoodson:


mwoodson:
@jupierce approved #3556

PR #3567, opened 2018-06-08
monitoring: added the hyperkube as a valid option for node process count

mwoodson:


mwoodson:
@jupierce approved #3566

PR #3571, opened 2018-06-13
multi_inventory.py: Ensure boolean groups exist

mbarnes:
For each boolean variable name in `clone_groups`, convert values to boolean and ensure both an `oo_NAME_True` and `oo_NAME_False` group exist, ~~inserting a placeholder if necessary to prevent Ansible from discarding empty groups in a dynamic inventory~~.

This way playbooks can safely reference `oo_NAME_True` or `oo_NAME_False` groups without having to supply a default empty list.

So for example:
```
   intersect(groups['oo_provision_True'])
```
is now safe, whereas before one would need:
```
   intersect(groups['oo_provision_True'] | default([]))
```


kwoodson (approved):

mbarnes:
I think maybe we don't need the placeholder for empty groups after all, now that I'm retesting.  I can't reproduce the original failure with an empty group in the inventory.

I removed the placeholder bit and left the rest.  I suggest merging this first and if it turns out Ansible _does_ skip empty inventory groups, I'll do another PR.

PR #3572, opened 2018-06-13
multi_inventory.py: Ensure boolean groups exist

mbarnes:
For each boolean variable name in `clone_groups`, convert values to boolean and ensure both an `oo_NAME_True` and `oo_NAME_False` group exist.

This way playbooks can reference `oo_NAME_True` or `oo_NAME_False` groups without having to supply a default empty list.

So for example:
```
   intersect(groups['oo_provision_True'])
```
is now safe, whereas before one would need:
```
   intersect(groups['oo_provision_True'] | default([]))
```


mbarnes:
@kwoodson :+1: #3571

PR #3575, opened 2018-06-18
multi_inventory.py: Fix handling of boolean values

mbarnes:
Values from EC2 are of type `unicode`, not `str`.


ansible/inventory/multi_inventory.py
```diff
@@ -298,7 +298,7 @@ def apply_clone_groups(self):
 
                 # Convert boolean variables to 'True' or 'False'.
                 if to_name in BOOLEAN_CLONE_GROUPS:
-                    if isinstance(val, str):
+                    if isinstance(val, unicode):
```
mwoodson: you should look at basestring or six.text_type to get both str and unicode

mwoodson (approved):

PR #3583, opened 2018-06-22
added docker registry symlinker in the daemoneset config loop pod thingy

mwoodson:


mwoodson:
@kwoodson approved #3582

PR #3586, opened 2018-06-22
docker registry try, take 2

mwoodson:


mwoodson:
@kwoodson approved #3585

PR #3593, opened 2018-06-25
updated SREBOT to run every 10 minutes instead of every 15 seconds

rafael-azevedo:


rafael-azevedo:
was approved by @aweiteka on STG branch

PR #3599, opened 2018-06-28
added a rsyslog restarter to scale group config pods to overcome rsyslog bug

mwoodson:


mwoodson:
@jupierce approved #3598

PR #3605, opened 2018-07-02
track evictions from the event stream

yithian:
once we start keeping a tally of evictions we can set up a zabbix
trigger for mass evictions


yithian:
@blrm :+1: #3600

PR #3617, opened 2018-07-05
fixed an ses var naming bug

mwoodson:


mwoodson:
@jupierce approved #3616

PR #3622, opened 2018-07-05
 Bind to loopback interface only

dbaker-arch:


ivanhorvath (approved):

dbaker-arch:
Merged to stg in #3618 
Merged to int in #3619

PR #3635, opened 2018-07-09
fixed generate byo to include node group var on each node type

mwoodson:


mwoodson:
@jupierce approved #3634

PR #3646, opened 2018-07-10
added kubeconfig change that is coming in 3.10

mwoodson:


mwoodson:
@jupierce approved #3645

PR #3653, opened 2018-07-11
added gather_facts to no on scalegroup pod

mwoodson:


mwoodson:
@jupierce approved #3652

PR #3658, opened 2018-07-13
fixed quota resource role to be idempotent for 3.9; ensured 3.7 works as well

mwoodson:


mwoodson:
@robotmaxtron approved #3657

PR #3659, opened 2018-07-13
updated SREbot since there was a change on the pagerduty escalation p…

rafael-azevedo:
…olicy


rafael-azevedo:
approved by @aweiteka on stg

PR #3662, opened 2018-07-25
exclude nvme volumes that were being incorrectly included

yithian:
volumes such as nvme12 didn't match the old regex but need to


yithian:
The automated testing image build still appears to be broken, hence the test failure.

mwoodson (approved):

PR #3663, opened 2018-07-25
update the openshift-tools-scripts rpm to include customer PV filter

yithian:


yithian:
updating the rpm for  #3662

PR #3677, opened 2018-08-02
3.11 openshift-ansible vendored

mwoodson:


mwoodson:
@jupierce approved #3676

PR #3686, opened 2018-08-13
add checks for nodes label

BlueShells:
sometimes a node lost some label causing us problem

### DIFF
--- a/scripts/monitoring/cron-send-node-labels-status.py
+++ b/scripts/monitoring/cron-send-node-labels-status.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python
+""" Node label check for OpenShift V3 """
+
+# Adding the ignore because it does not like the naming of the script
+# to be different than the class name
+# pylint: disable=invalid-name
+
+# pylint: disable=wrong-import-position
+# pylint: disable=broad-except
+# pylint: disable=line-too-long
+
+import argparse
+import time
+
+import logging
+logging.basicConfig(
+    format='%(asctime)s - %(relativeCreated)6d - %(levelname)-8s - %(message)s',
+)
+logger = logging.getLogger()
+logger.setLevel(logging.WARN)
+
+# Our jenkins server does not include these rpms.
+# In the future we might move this to a container where these
+# libs might exist
+#pylint: disable=import-error
+
+from openshift_tools.monitoring.ocutil import OCUtil
+from openshift_tools.monitoring.metric_sender import MetricSender
+
+ocutil = OCUtil()
+
+def runOCcmd(cmd, base_cmd='oc'):
+    """ log commands through ocutil """
+    logger.info(base_cmd + " " + cmd)
+    oc_time = time.time()
+    oc_result = ocutil.run_user_cmd(cmd, base_cmd=base_cmd, )
+    logger.info("oc command took %s seconds", str(time.time() - oc_time))
+    return oc_result
+
+def runOCcmd_yaml(cmd, base_cmd='oc'):
+    """ log commands through ocutil """
+    logger.info(base_cmd + " " + cmd)
+    ocy_time = time.time()
+    ocy_result = ocutil.run_user_cmd_yaml(cmd, base_cmd=base_cmd, )
+    logger.info("oc command took %s seconds", str(time.time() - ocy_time))
+    return ocy_result
+
+def get_type(hostname):
+    """get the host type , is the host is a master or node or a infra"""
+    host_type = 'compute'
+    if 'master' in hostname:
+        host_type = 'master'
+    elif 'infa' in hostname:
+        host_type = 'infra'
+    else:
+        host_type = 'compute'
+
+    return host_type
+
+def check_label_on_host(host_labels):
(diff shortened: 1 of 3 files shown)


scripts/monitoring/cron-send-node-labels-status.py
```diff
@@ -0,0 +1,100 @@
...
+    args = parser.parse_args()
+
+    if args.verbose > 0:
+        logger.setLevel(logging.INFO)
+    if args.verbose > 1:
+        logger.setLevel(logging.DEBUG)
+
+    return args
+
+def check_lable_status():
```
drewandersonnz: label*

scripts/monitoring/cron-send-node-labels-status.py
```diff
@@ -0,0 +1,100 @@
...
+        if len(ip_status) < 5:
+            logger.info('this node lack of ip: %s', item['metadata']['name'])
+            result_ip = result_ip + 1
+        # if the label on the host less than 11 , means lack of label, for 3.9
+        if len(labels) < 11:
+            logger.info('this node lack of labels: %s', item['metadata']['name'])
+            result_status = result_status  + 1
+
+    #if everything fine , result_ip and result_status both should 0 , if not 0 , means something wrong
+    return result_ip + result_status
```
drewandersonnz: should we seperate the number of "address" faults vs "label" faults?

scripts/monitoring/cron-send-node-labels-status.py
```diff
@@ -0,0 +1,100 @@
...
+    label_info = runOCcmd_yaml("get node ")
+    for item in label_info['items']:
+        labels = item['metadata']['labels']
+        ip_status = item['status']['addresses']
+        # if the host have less than 5 address , means node miss ip
+        if len(ip_status) < 5:
+            logger.info('this node lack of ip: %s', item['metadata']['name'])
+            result_ip = result_ip + 1
+        # if the label on the host less than 11 , means lack of label, for 3.9
+        if len(labels) < 11:
```
drewandersonnz: would it be better to check if specific labels are available? I'm not 100% sure on which ones, but relying on a count may not work. If the team adds new labels, but important ones are still missing, then this check will still pass.

scripts/monitoring/cron-send-node-labels-status.py
```diff
@@ -0,0 +1,100 @@
...
+def check_lable_status():
+    """get all the info of all node """
+    result_ip = 0
+    result_status = 0
+    label_info = runOCcmd_yaml("get node ")
+    for item in label_info['items']:
+        labels = item['metadata']['labels']
+        ip_status = item['status']['addresses']
+        # if the host have less than 5 address , means node miss ip
+        if len(ip_status) < 5:
```
drewandersonnz: should we check specific keys for this? I don't hink a count is suitable.

scripts/monitoring/cron-send-node-labels-status.py
```diff
@@ -0,0 +1,100 @@
...
+            logger.info('this node lack of ip: %s', item['metadata']['name'])
+            result_ip = result_ip + 1
+        # if the label on the host less than 11 , means lack of label, for 3.9
+        if len(labels) < 11:
+            logger.info('this node lack of labels: %s', item['metadata']['name'])
```
drewandersonnz: we should say which labels are missing.

scripts/monitoring/cron-send-node-labels-status.py
wanghaoran1988: I'd like add checks like this:
1. If it has "type=master", we should check whether the node has "node-role.kubernetes.io/master=true"
2. If it has "type=infra", we should check whether the node has "node-role.kubernetes.io/infra=true"
3. If it has "type=compute", we should check whether the node has "node-role.kubernetes.io/compute=true"

scripts/monitoring/cron-send-node-labels-status.py
```diff
@@ -0,0 +1,142 @@
...
+    result = True
+    hostname = labels['hostname']
+    host_type = labels['type']
+    # if the node miss the hostname and type label
+    if hostname and host_type:
+        pass
+    else:
+        return False
+
+    master_labels = {'beta.kubernetes.io/arch':'amd64', 'beta.kubernetes.io/instance-type': 'm4.xlarge', 'beta.kubernetes.io/os': 'linux', 'failure-domain.beta.kubernetes.io/region': 'us-east-1', 'failure-domain.beta.kubernetes.io/zone': 'us-east-1a', 'hostname': 'tke-master-9f206', 'kubernetes.io/hostname': 'ip-10-147-203-48.ec2.internal', 'network.openshift.io/not-enforcing-egress-network-policy':'true', 'node-role.kubernetes.io/master': 'true', 'region': 'us-east-1', 'type': 'master'}
```
wanghaoran1988: Do we need check all the labels? I don't think every node will have label 'region': 'us-east-1' .

drewandersonnz:
+1

drewandersonnz (approved):

PR #3688, opened 2018-08-14
Add k8s node labels to BYO generator template

blrm:
This adds node labels to the BYO generator so they will be set in node-config.yaml. Currently a node can lose these labels on 3.9 if it leaves the cluster due to a restart. This makes sure the label persists in that situation.


blrm:
I generated a BYO inventory with this and the labels look fine. I'm going to provision a stg cluster once this is in so I can confirm they land correctly with that as well before it goes to prod.

mwoodson (approved):

PR #3711, opened 2018-08-17
Changed oo_rsync_private_key to be the key, not a path to the key

jewzaam:


jewzaam:
+1 by tparikh in https://github.com/openshift/openshift-tools/pull/3710

PR #3718, opened 2018-08-20
Vendor in openshift-ansible-3.9.40

mbarnes:
Need the patch for:

[Bug 1579513 - 3.9 Upgrade Fails When Masters Are On Older Version Than Available](https://bugzilla.redhat.com/show_bug.cgi?id=1579513)

which is blocking us from scaling up nodes on 3.9.31/33 clusters in preparation for upcoming 3.9.41 / EC2 M5 upgrades.


dbaker-arch:
I didn't check all 3000+ files, but +1 presumed this is all vendored code, and the one line code diff.

dbaker-arch (approved):

PR #3721, opened 2018-08-21
de-prio node label alert, it's storming

gburges:


gburges:
@blrm and @ivanhorvath :+1:  in https://github.com/openshift/openshift-tools/pull/3720

ivanhorvath (approved):

PR #3724, opened 2018-08-21
Add role to optionally disable hyper-threading at runtime.

jewzaam:


ansible/roles/disable_hyperthreading/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Requirements
+------------
+
+None
+
+Role Variables
+--------------
+
+dh_active - if true, hyper-threading is disabled
```
dmsimard: It might just be me but I would find the reversed logic to be more straightforward ?
By that, I mean the role would be "hyperthreading" which would default to be enabled (true) and could be disabled (by setting false)

ansible/roles/disable_hyperthreading/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+- name: disable hyper-threading
+  block:
+  - name: install disableHT.sh
+    copy:
+      src: files/disableHT.sh
+      dest: /usr/local/bin/disableHT.sh
+      owner: root
+      group: root
+      mode: 0755
+    register: dh_script_out
```
dmsimard: Nit: can you add line breaks between tasks so it's easier to parse ?

ansible/roles/disable_hyperthreading/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+  - name: enable disableHT.service
+    systemd:
+      name: disableHT
+      enabled: yes
+  - name: start disableHT.service (if something changed)
+    systemd:
+      name: disableHT
+      state: started
+      daemon_reload: yes
+    when: dh_service_out.changed or dh_script_out.changed
```
dmsimard: Ansible is changing a little bit how conditions like "changed", "skipped" or "succeeded" are written: https://docs.ansible.com/ansible/2.6/user_guide/playbooks_conditionals.html#the-when-statement

I think moving forward, the way to write this would be ``dh_service_out is changed or dh_script_out is changed``.
You could also do away with this entirely by using a handler instead which is automatically fired when the task is changed so you don't need to look at the conditions for idempotency.
So, example:
```
  - name: install disableHT.service
    copy:
      src: files/disableHT.service
      dest: /etc/systemd/system/disableHT.service
      owner: root
      group: root
      mode: 0644
    notify:
      - reload disable hyperthread
```

and then a task in ``handlers/main.yaml`` with that name that does what you want.

ansible/roles/disable_hyperthreading/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: disable hyper-threading
+  block:
+  - name: install disableHT.sh
```
dmsimard: "hyperthread" (or "hyperthreading") isn't that long to type and everything is automated. Can we spell it out in file names and service names so that it's obvious what it does ?
i.e, ``disable-hyperthreading.sh`` or ``disable-hyperthreading.service``

ansible/roles/disable_hyperthreading/files/disableHT.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Disable Hyper-Threading at Runtime
+Documentation=https://access.redhat.com/security/vulnerabilities/L1TF https://access.redhat.com/solutions/352663
+Requires=atomic-openshift-node.service
+Before=atomic-openshift-node.service
+
+[Service]
+User=root
+Type=oneshot
```
dmsimard: Reading the docs, it sounds like we might want to also add ``nosmt`` to the kernel boot parameters so that the change is not only applied at runtime but persisted through reboots. Should we do both ?

ansible/roles/disable_hyperthreading/files/disableHT.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Disable HT at Runtime
+
+# see https://access.redhat.com/solutions/352663
+# preferred: /sys/devices/system/cpu/smt/control
+# fallback: /sys/devices/system/cpu/cpu
+
```
dmsimard: Out of curiosity, in which case would we not have ``smt/control`` available ? Is it a matter of cloud provider, virtualization platform or OS version ?

dmsimard: Replying to self, it seems like the only factor is the kernel version ? Looking at a limited sample of some nodes we have, I don't see smt available anywhere .. not even on my Fedora laptop ?

ansible/roles/disable_hyperthreading/README.md
jewzaam: That is the default unless disabled by something else like nosmt boot param.  Service and script will not be deployed unless you want to disable HT.  Alternately, they could be deployed, just not used.  Though I don't know what the value of that would be other than consistency across all node types, other than which (compute) have it enabled.

ansible/roles/disable_hyperthreading/files/disableHT.service
jewzaam: No, discussed in email thread.  We have a single AMI for master, infra, and compute for all styles of nodes.  We want to keep it this way.  Only a subset will have hyper-threading disabled.

ansible/roles/disable_hyperthreading/files/disableHT.sh
jewzaam: Yes, kernel version.  I went ahead and handled older version just in case, and because my test cluster is on an old version and I didn't update.

ansible/roles/disable_hyperthreading/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Dependencies
+------------
+
+None
+
+Example Playbook
```
dmsimard: Let's remove this part if we're not going to put something relevant in it

dmsimard:
Also somewhat related -- what's the workflow for re-enabling hyperthreading if/when security issues are resolved ?

jewzaam:
Guess having a way to back this out would be good.  Without any changes to the role as it stands, you'd disable or no longer use the role, disable the service, and reboot.

jewzaam:
fwiw, to enable at runtime:
```
echo on > /sys/devices/system/cpu/smt/control || for CPU in `ls /sys/devices/system/cpu/cpu[0-9]* -d`; do echo 1 > $CPU/online; done
```

jewzaam:
@dmsimard refactored to be "enabled" based in the naming and to support toggling between enabled and disabled.

I'm not a huge fan of the handler that turns the cpu's back on, but it didn't make sense to land a script that would be run just that one time.  Looking for feedback on if it needs to be changed.

ansible/roles/hyperthreading/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
+---
+- name: disable hyper-threading
+  block:
+
```
dmsimard: I didn't even realize you could name blocks until now -- makes me wonder if it actually does anything. I'll have to look.
Anyway, something that we can do to improve readability when writing conditional blocks is to put the conditions at the top so that it's obvious when reading top to bottom:
```
# Disable hyperthreading
- when: not hyperthreading_enabled
  block:
    - name: install disable-hyperthreading.sh
      copy:
        src: files/disable-hyperthreading.sh
        dest: /usr/local/bin/disable-hyperthreading.sh
        owner: root
        group: root
        mode: 0755
      notify:
        - reload and start disable-hyperthreading

[...]

# Enable hyperthreading
- when: hyperthreading_enabled
  block:
    - name: check service exists
      stat:
        path: /etc/systemd/system/disable-hyperthreading.service
      register: service_out

[...]
```

ansible/roles/hyperthreading/handlers/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: reload and start disable-hyperthreading
+  systemd:
+    name: disable-hyperthreading
+    state: started
+    daemon_reload: yes
+
+- name: enable CPUs that were disabled
+  shell: "if [ -f /sys/devices/system/cpu/smt/control ]; then echo on > /sys/devices/system/cpu/smt/control; else for CPU in `ls /sys/devices/system/cpu/cpu[0-9]* -d`; do echo 1 > $CPU/online; done fi"
```
dmsimard: I'd split this out to multiple lines, otherwise it looks fine.

```
- name: enable CPUs that were disabled
  shell: >
    if [ -f /sys/devices/system/cpu/smt/control ]; then
        echo on > /sys/devices/system/cpu/smt/control
    else
        for CPU in $(ls /sys/devices/system/cpu/cpu[0-9]* -d); do
            echo 1 > $CPU/online
        done
    fi
```

dmsimard:
@jewzaam looks good to me, just some minor style comments. Handlers make sense if you need to run *something* only once (i.e, 10 notify for the same handler will only fire it once) -- turning the hyperthreading back on doesn't need to be a handler if we don't feel it's a good fit.

jewzaam:
@dmsimard updates made, thanks for the comments!  Regarding the handler that enables HT, I started down the non-handler path but it only is run if we disable the service.  Handler was the easiest way to do it so I do think it makes sense.

ansible/roles/hyperthreading/files/disable-hyperthreading.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Disable Hyper-Threading at Runtime
+
+# see https://access.redhat.com/solutions/352663
+# preferred: /sys/devices/system/cpu/smt/control
+# fallback: /sys/devices/system/cpu/cpu
+
+if [ -f /sys/devices/system/cpu/smt/control ];
+then
```
dmsimard: Nitpick: let's be consistent with our bash formatting, here we have:
```
if [ -f /sys/devices/system/cpu/smt/control ]; 
then
```

and in the handler we have:
```
if [ -f /sys/devices/system/cpu/smt/control ]; then
```

I personally prefer the latter than the former but can live with either, so long as we have consensus on which to use.
We can probably implement CI jobs for things like [shellcheck](https://www.shellcheck.net/) in the future to help us keep things consistent.

dmsimard (approved):

ansible/roles/hyperthreading/files/disable-hyperthreading.sh
jewzaam: fwiw, the majority of scripts in this repo use the latter so went with that.

dmsimard (approved):

PR #3731, opened 2018-08-28
fix the timeout error in script

BlueShells:
fix the timeout error in script


BlueShells:
@drewandersonnz had +1 #3731

PR #3749, opened 2018-09-19
Updates to grow var script for m5

jewzaam:
- add reserved sectors option (default is 0)
- repartition script sets explicit end sector
- trigger filesystem metrics at end (also added for m4 script)
- keep repartition script until execution is successful (to help debug)

### DIFF
--- a/ansible/playbooks/adhoc/grow_root_vg/grow_root_vg-m5.yml
+++ b/ansible/playbooks/adhoc/grow_root_vg/grow_root_vg-m5.yml
@@ -257,5 +257,11 @@
       path: /root/grow_root_vg-m5-recreate-partition.sh
       state: absent
 
-  - name: Trigger filesystem metrics
+  - name: Trigger filesystem metrics (master or infra)
     shell: "docker exec -ti oso-rhel7-host-monitoring /usr/bin/cron-send-filesystem-metrics -v"
+    when: "('oo_hosttype_master' in group_names or 'oo_subhosttype_infra' in group_names)"
+
+  - name: Trigger filesystem metrics (compute)
+    shell: "docker exec -ti oso-rhel7-host-monitoring /usr/bin/cron-send-filesystem-metrics --filter-pod-pv -v"
+    when: "'oo_subhosttype_compute' in group_names"
+
--- a/ansible/playbooks/adhoc/grow_root_vg/grow_root_vg-m5.yml
+++ b/ansible/playbooks/adhoc/grow_root_vg/grow_root_vg-m5.yml
@@ -29,8 +29,9 @@
   vars:
     cli_volume_size: 90
     cli_var_size: 64
-    cli_reserved_sector_count: 5000
+    cli_reserved_sector_count: 0
     os_size_min: 20
+    cli_var_partition_number: 2
 
   tasks:
   - name: "Check for required variables" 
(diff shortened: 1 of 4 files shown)


ansible/playbooks/adhoc/grow_root_vg/grow_root_vg.yml
```diff
@@ -163,3 +163,6 @@
 
   - name: Resize /var
     command: "xfs_growfs /var "
+
+  - name: Trigger filesystem metrics
+    shell: "docker exec -ti oso-rhel7-host-monitoring /usr/bin/cron-send-filesystem-metrics -v"
```
blrm: This command runs with an additional flag on compute nodes so that we don't alert on customer PVs that are full:
`docker exec -i oso-rhel7-host-monitoring cron-send-filesystem-metrics --filter-pod-pv` . This should probably check which group the node its running on is a member of and either run one or the other.

You could probably use a when clause for this like:
`when: "'oo_subhosttype_compute' in group_names"` for compute and `when: "('oo_subhosttype_infra' in group_names) or ('oo_hosttype_master' in group_names)"` for masters and infras.

ansible/playbooks/adhoc/grow_root_vg/grow_root_vg-m5.yml
```diff
@@ -225,7 +247,15 @@
...
-    when: new_var_size.stdout|int != cli_var_size|int
+    when: new_var_size.stdout|int < cli_var_size|int
+
+  - name: Remove recreate script when successful
+    file:
+      path: /root/grow_root_vg-m5-recreate-partition.sh
+      state: absent
+
+  - name: Trigger filesystem metrics
+    shell: "docker exec -ti oso-rhel7-host-monitoring /usr/bin/cron-send-filesystem-metrics -v"
```
blrm: See my other comment. This should check which node type its running on.

ansible/playbooks/adhoc/grow_root_vg/grow_root_vg.yml
jewzaam: Can the `--filter-pod-pv` always be set or does it specifically only get set for compute?

blrm: It shouldn't be set for infra nodes because this alerts us to when infra PVs are full (like elasticsearch and cassandra volumes). I think it matters a little less for the masters nodes (since they don't have any PVs anyway),

ansible/playbooks/adhoc/grow_root_vg/grow_root_vg-m5.yml
```diff
@@ -179,6 +181,14 @@
   - name: Run partprobe to see if the size changed
     command: partprobe
 
+  - name: Get partition info (updated)
+    parted: 
+      device: "{{ vol_instance.fullname }}"
+      number: "{{ cli_var_partition_number }}"
+      state: info
+      unit: s
+    register: parted_info
```
blrm, line 190: We talked a little bit about failing early if the playbook detects that new partitioning scheme is present to prevent clobbering those nodes. I don't think that necessarily needs to be present for this PR to be approved but wanted to see if this point in the playbook would be where to do that?

jewzaam: It would be, yes.  You can check length of the partitions array.  I don't have the exact json handy.

For `/var/log` being split out, I did look at that and there is still only two partitions.  `/var/log` is just a logical volume.  So I don't think we need to spend time to add a check here.  If we need to use this playbook again there's a cli arg to indicate the partition you wish to grow, defaulting to 2.

jewzaam: `parted_info.partitions | length`

blrm (approved):

PR #3757, opened 2018-09-27
os_zabbix: fixed up the zabbix server db checks

mwoodson:


mwoodson:
@yithian approved #3756

PR #3759, opened 2018-09-27
Do not update sublocation for synthetic_hosts.

kwoodson:


kwoodson:
@mwoodson :+1:  in #3758

PR #3766, opened 2018-10-02
Fix m units

fahlmant:
After reading upstream documentation (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/resources.md#resource-quantities) only available in a design proposal, I see that lowercase m represents millibytes. This script change is to fix the conversion of m to bytes.


fahlmant:
There are extra commits here since I based this branch off prod, and prod and stg have diverged

yithian (approved):

PR #3768, opened 2018-10-02
added lib_openshift_ansible_utils to roles

mwoodson:


wshearn (approved):
jenkins is broken but this shouldn't affect anything

PR #3769, opened 2018-10-02
added lib_openshift_ansible_utils to roles

mwoodson:


mwoodson:
@wshearn approved #3768

PR #3770, opened 2018-10-05
reduce config loop severity failure from high to average

aweiteka:
Reason: a config loop failure does not impact a customer so it's not worth waking someone up. It's important to fix but should be addressed by better CI. Without CI it isn't fair to wake an engineer.

Signed-off-by: Aaron Weitekamp <aweiteka@redhat.com>


aweiteka:
cc @jharrington22 we already have (broken) tests. below

jewzaam:
If severity is reduced will we ever see it?  Can we trigger a higher severity alert if config loop continues to fail after some period of time?  I fear with this change we'll not know config loop is broken until debugging some symptom of it being broken.

robotmaxtron:
@jewzaam +1 to adding a secondary alert, maybe somrthing like 'Config loop has been disabled for longer than 72h?'

jewzaam:
Something like that.  Need to figure out how long being down is
"acceptable".  And by "disabled" you mean "failing"?  With the assumption
that config loop is running.

PR #3732, opened 2018-08-28
add inner check for kibana if customer disable public access

BlueShells:
add inner check for kibana if customer disable public access

PR #3774, opened 2018-10-10
Enable service ip check for kibana

wanghaoran1988:
Rework of https://github.com/openshift/openshift-tools/pull/3732


BlueShells:
:+1

PR #3775, opened 2018-10-10
enable service ip check for kibana

wanghaoran1988:


wanghaoran1988:
+1 from https://github.com/openshift/openshift-tools/pull/3774

PR #3778, opened 2018-10-11
Allow same certs to be uploaded next time playbook runs

tparikh:
Right now running the cert update playbook multiple times fails if the certificate itself has not changed. This change allows us to upload same certificate again.

https://docs.ansible.com/ansible/2.5/modules/iam_cert_module.html

stg pr: https://github.com/openshift/openshift-tools/pull/3777


tparikh:
+1 in #3777

PR #3779, opened 2018-10-12
new role for the htpasswd

BlueShells:
new role for the htpasswd


ansible/roles/openshift_htpasswd/templates/htpasswd.j2
```diff
@@ -0,0 +1,3 @@
+{% for user in oshtpasswd_users %}
```
drewandersonnz, line 1: can you add a defaults file for this variable? I think an empty list would be fine.

drewandersonnz: After chatting, it may be better to just have the role fail with "undefined variable" if the var is not defined correctly.

drewandersonnz (approved):

PR #3782, opened 2018-10-15
Configloop htpasswd

BlueShells:
Configloop htpasswd


BlueShells:
@drewandersonnz had +1 on https://github.com/openshift/openshift-tools/pull/3779

PR #3787, opened 2018-10-16
a ansible/python module to calcuate subnets needed for OSD 3.11 install

tparikh:
STG PR: https://github.com/openshift/openshift-tools/pull/3786

An ansible module to calculate subnets based on VPC CIDR block, number of availability zones in the deployment AWS region and desired number of subnets (1 or 3). Maximum hosts supported per subnet is 2048. Minimum hosts subnet must be able to support is 14. If the CIDR block provided is too small, module will throw error.

Scenario 1:
Deployment region has at least 3 availability zones, CIDR block is 172.16.0.0/16 and we want 3 subnets. Calculator will return subnets "172.16.0.0/21", "172.16.8.0/21", "172.16.16.0/21".

Scenario 2:
Deployment region has 2 availability zones, CIDR block is 172.16.0.0/16 and we want 3 subnets. Calculator will return only 1 subnet with CIDR "172.16.0.0/21".

Scenario 3:
Deployment region has 3 availability zones, CIDR block is 172.16.0.0/24 and we want 3 subnets. Calculator will return subnets "172.16.0.0/26", "172.16.0.64/26", "172.16.0.128/26".

Scenario 4:
Deployment region has 3 availability zones, CIDR block is 172.16.0.0/29 and we want 3 subnets. Calculator will return error as the CIDR block is too small. We require CIDR for each subnet to support at least 14 hosts.

### DIFF
--- a/ansible/roles/lib_utils/library/oo_vpc_subnets_calc.py
+++ b/ansible/roles/lib_utils/library/oo_vpc_subnets_calc.py
@@ -1,0 +1,109 @@
+#!/usr/bin/python
+# vim: expandtab:tabstop=2:shiftwidth=2
+
+import boto3
+import netaddr
+import math
+
+''' vpc_subnets_calc module
+
+    The purpose of this module is to cat the contents of a file into an Ansible fact for use later.
+    
+    
+Test Playbook:
+---
+- hosts: localhost
+  gather_facts: no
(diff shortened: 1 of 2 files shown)


tparikh:
@mwoodson approved #3786

PR #3795, opened 2018-10-22
need apiversion for this to work now, it might stop working in 4.X at…

ivanhorvath:
… some point, was told that there will be deprecation errors sooner or later, so will need different fix then


ivanhorvath:
Plust by @mwoodson in #3794

PR #3797, opened 2018-10-22
map each subnet to an az in the region

tparikh:
STG PR: https://github.com/openshift/openshift-tools/pull/3796


tparikh:
+1 in #3796 from @mwoodson

PR #3821, opened 2018-11-01
added olm logic, updated the sre provisioning container to have a default ansible.cfg

mwoodson:


mwoodson:
@blrm approved #3820

PR #3842, opened 2018-11-06
Adding scripts for insights and hyperthreading

blrm:


ansible/roles/openshift_firstboot_scripts/files/disable-hyperthreading.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo off > /sys/devices/system/cpu/smt/control
```
mwoodson, line 4: I think we should add the grubby command here in case the machine reboots.

ansible/roles/openshift_firstboot_scripts/files/insights-register.sh
```diff
@@ -0,0 +1,14 @@
...
+case "${option}"
+in
+c) CLUSTERNAME=${OPTARG};;
+t) NODETYPE=${OPTARG};;
+esac
+done
+
+INTERNAL_HOSTNAME=$(curl 169.254.169.254/latest/meta-data/local-hostname)
+
+/usr/bin/redhat-access-insights --register --display-name=$CLUSTERNAME-$NODETYPE-$INTERNAL_HOSTNAME --group=$CLUSTERNAME
```
mwoodson: should we add this to the current subscription manager script?  I don't have a feeling one way or another, just a thought...

2nd issue, should we background this?

ansible/roles/openshift_firstboot_scripts/files/disable-hyperthreading.sh
blrm: I added a grubby command to add this.

ansible/roles/openshift_firstboot_scripts/files/insights-register.sh
blrm: Good call. The timeout for this command is much lower than it used to be, but still isn't time we should spend waiting if we don't need to. I updated the command to background it.

mwoodson (approved):

PR #3846, opened 2018-11-08
removed handlers from the quota role

mwoodson:


mwoodson:
@mbarnes approved #3845

PR #3848, opened 2018-11-08
Port oc groups bug fix from upstream openshift-ansible

fahlmant:
Porting bugfix in upstream `oc_groups` to our vendored versions

https://github.com/openshift/openshift-ansible/pull/10075


fahlmant:
+1 from @mwoodson in #3849

PR #3852, opened 2018-11-08
added restart handlers to look at the master-restart instead of the rpm version

mwoodson:


mwoodson:
@mbarnes approved #3851

PR #3860, opened 2018-11-15
Revert label check

BlueShells:
remove the label check temp before the labels confirmed


BlueShells:
@grantomation has +1 on #3859

PR #3871, opened 2018-11-20
Determine the node's sub host type on start-up

blrm:


blrm:
@wshearn :+1: in slack.

PR #3882, opened 2018-11-28
changed the version rpm check to look at atomic-openshift-clients

mwoodson:


mwoodson:
@wshearn approved #3881

PR #3893, opened 2018-12-06
add configmap manage role in configloop

BlueShells:


ansible/roles/openshift_configmap_fluentd/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: createthe tmp file
+  copy:
+   src: private_vars/global/{{ ipv_location }}/{{ ipv_account }}/{{ ipv_environment }}/{{ ipv_clusterid }}/fluentd.conf
+   dest: /tmp/tmp-fluentd.conf
```
mwoodson, line 5: we can't do this like this.  This references private, and that can't happen in openshift-tools.

PR #3896, opened 2018-12-07
Update openshift_udp_limits to not delegate to master

blrm:
This updates the udp_limits role to make its oc call from the node it is running from instead of delegating to a master. This will allow this role to run inside a scalegroup compute node without having to have any masters in its ansible inventory.


dbaker-arch:
Looks good, provided the compute nodes always have valid kube creds in that file.

dbaker-arch (approved):

PR #3901, opened 2018-12-14
Fix template reference

jharrington22:
This PR removes a space that is causing a template error during zabbix ansible runs


dak1n1 (approved):
:+1:

jharrington22:
+1 by @jgouldin in IRC too

PR #3905, opened 2018-12-20
Update the Zabbix local development environment tooling

yithian:
This updates the local development scripts to use the current oc syntax
and secrets format. It also updates the lib_zabbix Ansible module and
the zbxapi Python module for Python 3.


yithian:
prod still uses python 2 so

PR #3911, opened 2019-01-10
Update for new finer grained permissions per account

dbaker-arch:
Add new get_and_verify_saml_token script for finer grained filtering on assumed roles.


dbaker-arch:
This is pending final testing from CICD team before merging.

rhdedgar (approved):
:+1:

PR #3919, opened 2019-01-15
[spre-447] only select az if it can support required instance types

tparikh:
Not all EC2 instance types are available for all AWS availability zone. We should only select AZs that can support instance types needed for master, compute and infra nodes.


tparikh:
+1 from @mwoodson in https://github.com/openshift/openshift-tools/pull/3918

PR #3921, opened 2019-01-15
Add cloud_storage_udev_rules role

mbarnes:
Installs udev rules to assist with cloud storage configuration.

* Make predictable NVMe device names on AWS.
  Based on: coreos/init#268


mbarnes:
:+1: @mwoodson #3920

PR #3909, opened 2019-01-09
Update lib_openshift

gburges:

PR #3923, opened 2019-01-17
Update lib_openshift

gburges:


gburges:
@tparikh and @mwoodson :+1:  in https://github.com/openshift/openshift-tools/pull/3909

PR #3928, opened 2019-01-17
updated the playbook executor code

mwoodson:


mwoodson:
@wshearn approved #3927

PR #3936, opened 2019-01-28
Add tolerations to openshift-config DaemonSet

wanghaoran1988:


mwoodson:
I dont' know much about tolerations.  Where did these values come from?

I'm probably +1, just curious how these were arrived at?

wanghaoran1988:
This is k8s docs about "taints" and "tolerations" https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
the parts I added means it will tolerate the node that have any "taint" with "NoExecute" effect. so when a node running into problems, the node will have a "taint" (you can run oc describe node <node> to see the taints it have) if our config pod does not tolerate the taint, the daemonset controller will delete the pod. this will make the node have no monitor container running on that.
we should tolerate everything making sure it's under monitor no matter what the node status is.

wanghaoran1988:
Oh, another thing is that the apiserver, controller-manager, and etcd status pods running under kube-system project have the same "tolerations", you can check with: oc get pods -n kube-system -o yaml |grep -A 2 tolerations

mwoodson (approved):

mwoodson:
+1, thanks for doing this.

PR #3944, opened 2019-01-31
Srep 785

rafael-azevedo:


ansible/roles/lib_utils/library/oo_ec2_elb_set_cert.py
```diff
@@ -0,0 +1,104 @@
...
+        self.elb_client = boto3.client('elb')
+
+        if not external_elb_name or not SSLCertificateId:
+            self.module.fail_json(failed=True,
+                                  msg='external_elb_name or aws_iam_openshift_cert_arn not provided')
+
+        # if cert is already applied exit without errors
+        current_cert_id = self.get_ssl_cert_arn(external_elb_name)
+        if current_cert_id == SSLCertificateId:
+            self.module.exit_json(failed=True, changed=False, msg="SSLCertificateId %s already assigned to load balancer %s" % (SSLCertificateId, external_elb_name))
```
mwoodson: should we fail here?

This seems to be an idempotent operation (no action needed).  So i would think failed=Fase, changed=false

rafael-azevedo: changing

PR #3947, opened 2019-02-01
cloud_storage_udev_rules: Rename "sd*" devices to "xvd*"

mbarnes:
For consistency with infra (R4) instances.


mbarnes:
:+1: @mwoodson #3946

PR #3950, opened 2019-02-05
updated the sre first boot script insights to register with the cluster name and node type

mwoodson:


jewzaam (approved):
Not a huge fan of using `meta-data/iam/info` and parsing it but we don't have any other options that don't require additional credentials.  Can't get tags from the `meta-data` endpoint from what I can see.

mwoodson:
I tried looking at a lot of things:

1) the openshift node-config.  problem is, it's not configured when it comes up. you have to wait.
2) query aws for tags, but again, that's more credentials than I want to give
3) security groups are listed, and it has the type and cluster, but, harder to parse
4) i settled on metda data iam.  It works better than

PR #3951, opened 2019-02-05
updated the sre first boot script insights to register with the cluster name and node type

mwoodson:


mwoodson:
@jewzaam approved #3950

mwoodson:
@blrm approved #3958

PR #3954, opened 2019-02-15
Add adhoc playbook: add_etcd_volume.yml

mbarnes:
Provision and attach a volume for etcd storage to each master node, format them, and migrate etcd data.


ansible/playbooks/adhoc/add_etcd_volume/add_etcd_volume.yml
```diff
@@ -0,0 +1,352 @@
...
+
+    - name: Remove stopped member from etcd cluster
+      shell: ". /etc/profile.d/etcdctl.sh && etcdctl3 member remove {{ '%x' | format(etcd_self.ID) | quote }}"
+      delegate_to: "{{ etcd_delegate_to }}"
+
+    - name: Add stopped member to etcd cluster
+      shell: ". /etc/profile.d/etcdctl.sh && etcdctl3 member add {{ etcd_self.name | quote }} --peer-urls={{ etcd_self.peerURLs | join(',') | quote }}"
+      register: etcd_member_add
+      delegate_to: "{{ etcd_delegate_to }}"
+
```
dak1n1, line 279: Do you have any example output from `etcdctl3 member add`? It might be helpful to put it here in comments, to give people an idea of what exactly is being parsed in etcd_conf_updates. Just in case we have to troubleshoot it later.

dak1n1 (approved):
:+1: (one optional comment to add, but the logic looks good).

ansible/playbooks/adhoc/add_etcd_volume/add_etcd_volume.yml
mbarnes: Good idea, will add some sample output.

Maybe also for `etcdctl3 member list` so the JSON structure is clear.

PR #3964, opened 2019-03-01
add_etcd_volume: Support M4 instances

mbarnes:
On M5 instances, the expected device name for the etcd volume is actually a symbolic link created by udev at boot time that points to whatever NVMe device name the volume was assigned to. The NVMe
device assignment is arbitrary and can change on reboot.

On M4 instances, however, the expected device name is actually a block device, and the device assignment is stable across reboots.

The playbook must support both instance types.


mbarnes:
:+1: @yithian #3963

PR #3975, opened 2019-03-13
add cron job for the ssl check by http header

BlueShells:
check the code


BlueShells:
@drewandersonnz  have +1 on #3976

PR #3978, opened 2019-03-13
add zabbix item for ssl check

BlueShells:
add zabbix item for ssl check


BlueShells:
@drewandersonnz has +1 on #3977

PR #3982, opened 2019-03-14
add avg trigger for ssl check

BlueShells:
add avg trigger for ssl check


BlueShells:
@drewandersonnz has +1 on #3981

PR #3995, opened 2019-04-01
add method to allow ssl check on no 443

BlueShells:
add method to allow ssl check on no 443


BlueShells:
@drewandersonnz has +1 on #3994

PR #3996, opened 2019-04-01
add cron job for inter ssl check

BlueShells:
add zabbix item for inter ssl check 
add avg trigger for inter ssl check 
add cron job for inter ssl check

PR #3997, opened 2019-04-01
add zabbix item for inter ssl check

BlueShells:
add zabbix item for inter ssl check
add avg trigger for inter ssl check
add cron job for inter ssl check


BlueShells:
@drewandersonnz has +1 on #3996

PR #3999, opened 2019-04-02
change ssl check from avg to high

BlueShells:


BlueShells:
@drewandersonnz has +1 on #3998

PR #4000, opened 2019-04-08
add checks for dedicated-admin service status

BlueShells:
add checks for dedicated-admin service status


drewandersonnz:
Can this task be completed by using `cron-send-process-count.sh` here https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-process-count.sh instead of creating the new script `cron-send-dedicated-admin-status.py`?

PR #4002, opened 2019-04-08
add monitoring for dedicated-admin service

BlueShells:
add monitoring for dedicated-admin service


BlueShells:
@drewandersonnz has +1 on #4001

PR #4004, opened 2019-04-08
add trigger for the openshift-dedicated-admin service

BlueShells:
add trigger for the openshift-dedicated-admin service


BlueShells:
@drewandersonnz has +1 on #4003

PR #4005, opened 2019-04-10
change level from avg to high for dedicated-admin service trigger

BlueShells:
change level from avg to high for dedicated-admin service trigger


wanghaoran1988:
+1

PR #4006, opened 2019-04-10
change level from avg to high for dedicated-admin service trigger

BlueShells:


BlueShells:
@wanghaoran1988 has +1 on #4005

PR #4007, opened 2019-04-15
bugfix for the ssl check

BlueShells:
bugfix for the ssl check

### DIFF
--- a/scripts/monitoring/cron-send-ssl-check.py
+++ b/scripts/monitoring/cron-send-ssl-check.py
@@ -72,7 +72,10 @@ def get_ssl_certificate_expiry_days(domain_name):
     if ":" in domain_name:
         ssl_port = domain_name.split(":")[1]
         domain_name = domain_name.split(":")[0]
-    cert = ssl.get_server_certificate((domain_name, ssl_port))
+    conn = ssl.create_connection((domain_name, ssl_port))
+    context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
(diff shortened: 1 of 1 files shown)


scripts/monitoring/cron-send-ssl-check.py
```diff
@@ -72,15 +72,18 @@ def get_ssl_certificate_expiry_days(domain_name):
...
+    cert = ssl.DER_cert_to_PEM_cert(sock.getpeercert(True))
     x509 = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, cert)
 
     notafter = x509.get_notAfter()
     notafter = notafter[0:8]
     now = datetime.datetime.now()
     notafter_t = datetime.datetime.strptime(notafter, '%Y%m%d')
     delta = notafter_t - now
-    logger.info("public url [" + domain_name + "] will expire in: " + str(delta.days) + " days")
+    logger.info("public url [" + str(x509.get_subject()) + "] will expire in: " + str(delta.days) + " days")
```
drewandersonnz:
```
logger.info("public url [" + domain_name + "] certificate subject: " + str(x509.get_subject()))
logger.info("public url [" + domain_name + "] will expire in: " + str(delta.days) + " days")	
```

drewandersonnz: the purpose of showing both is so we can identify where exactly the fault is if there is one.

drewandersonnz (approved):

PR #4008, opened 2019-04-15
fix the bug of ssl check

BlueShells:
fix the bug of ssl check


BlueShells:
@drewandersonnz has +1 on #4007

PR #4012, opened 2019-05-10
bug fix for the hostname verify for ssl check

BlueShells:
bug fix for the hostname verify for ssl check


drewandersonnz (approved):
approved, and live tested. this change validates hostname and will error if hostname does not match.

PR #4013, opened 2019-05-10
bug fix for the hostname verify for ssl check

BlueShells:


BlueShells:
@drewandersonnz already +1 on #4012

PR #4014, opened 2019-05-10
add_ca_file and add_ca_path

drewandersonnz:
https://docs.python.org/2.7/library/ssl.html#ssl.SSLContext.load_verify_locations

Adding CA files and paths to params list.


BlueShells:
👍

PR #4016, opened 2019-05-10
add_ca_file

drewandersonnz:
Requires https://github.com/openshift/openshift-tools/pull/4015 otherwise the check will fail due to unknown argument.


BlueShells:
👍

PR #4019, opened 2019-05-14
tmp remove internel check of ssl

BlueShells:


BlueShells:
@drewandersonnz has +1 on #4018

PR #4025, opened 2019-05-30
Only alert after ES does automated disk balancing

dak1n1:
In order to prevent humans from being paged for an issue that
automation can solve, set the alert thresholds lower than those of Elastic
Search's watermark thresholds. This will cause the disks to automatically
become more balanced, rather than alerting humans when no action is
needed.

This also changes the expression used in Zabbix to determine whether
there is a disk problem. This is meant to prevent alert flapping, by
setting a specific recovery threshold higher than that of the alert
threshold.

Related documentation on watermark defaults:
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/disk-allocator.html#disk-allocator

Related documentation on Zabbix expression hysteresis:
https://www.zabbix.com/documentation/3.0/manual/config/triggers/expression#hysteresis

This is part of ticket SREP-1095.


dak1n1:
Approved in PR #4026

PR #4030, opened 2019-06-13
Add monitor scripts to check pods status on compute node

wanghaoran1988:


scripts/monitoring/cron-send-node-pods-status.py
```diff
@@ -0,0 +1,88 @@
...
+        logger.setLevel(logging.DEBUG)
+
+    return args
+
+def check_node_pods_status(nsList):
+    """get all the info of all node """
+    result_status = 0
+    nsFilter = ""
+    for ns in nsList:
+        nsFilter += ",metadata.namespace!="+ns
```
bmeng, line 62: Will this work for the first item?

wanghaoran1988: What do you mean the first item?

bmeng: Looks like you generate a namespace list to be excluded and pass them to the oc api.

But in the beginning of the list there will be a `,`? I am not sure if this would work well.

wanghaoran1988: it works as expected

drewandersonnz (approved):
+1 for new script. lets see it in action and what it discovers.

PR #4031, opened 2019-06-17
Add monitor scripts to check pods status on compute node

wanghaoran1988:


wanghaoran1988:
+1 from https://github.com/openshift/openshift-tools/pull/4030

PR #4037, opened 2019-06-18
Add dns resolution timeout monitor item

wanghaoran1988:


drewandersonnz (approved):
+1, disables an old data point and tracks both "can we resolve dns" and "can we resolve dns within 0.2s". I think there is benefit to having both data points.

PR #4043, opened 2019-06-20
add zabbix monitor status check script

BlueShells:


BlueShells:
@drewandersonnz has +1 on #4040

PR #4048, opened 2019-06-20
add cronjob for zabbix check

BlueShells:
add cronjob for zabbix check


BlueShells:
@wanghaoran1988 has +1 on #4047

PR #4052, opened 2019-06-26
add trigger for zabbix check

BlueShells:


BlueShells:
@wanghaoran1988 has +1 #4051

PR #4053, opened 2019-06-27
[stg] monitoring: add the monitoring script for console pods

bmeng:
Add the monitoring script for both web console and admin console.

JIRA: SREP-1524


BlueShells:
👍

PR #4067, opened 2019-07-11
Adjust Zabbix alert thresholds for ES disks

dak1n1:
With the settings in ElasticSearch now configured to auto-balance the
disk usage across the 3 disks, Zabbix will only alert *after* that
balancing has been done. This corresponds with the ES setting:

`cluster.routing.allocation.disk.watermark.low: 70%`

At 29% disk free, ES will stop placing indexes on the node and instead
place them on nodes with more space. Zabbix will alert at 28% disk free.

https://jira.coreos.com/browse/SREP-1536


lisa (approved):
/lgtm

PR #4074, opened 2019-07-17
Added zabbix check for dedicated-admin-operator

dak1n1:
* Updated the cron job that checks for dedicated-admin.
* Added new script to detect the status of dedicated-admin-operator.

https://jira.coreos.com/browse/SREP-1743


dak1n1:
New PR here https://github.com/openshift/openshift-tools/pull/4076

PR #4081, opened 2019-07-18
oso-zagg-web: Clean up leftover Ansible temporary directories

mbarnes:
Use of `simplezabbix.py` causes temporary directories to accumulate in `~/.ansible/tmp`, possibly due to an [Ansible bug](https://github.com/ansible/ansible/issues/56144).  This is a hack to periodically clean them up.


mbarnes:
:+1: @mwoodson #4080

PR #4095, opened 2019-07-29
cron-send-zabbix-inventory-check.py: Fix hostnames for masters

mbarnes:
Older master nodes may still have a `hostname` metadata label.  Otherwise construct the hostname from the provided cluster ID and the `type` and `kubernetes.io/hostname` metadata labels.


mbarnes:
:+1: @robotmaxtron #4094

PR #4097, opened 2019-07-30
fix trigger for dedicated admin check

BlueShells:


BlueShells:
@drewandersonnz has +1 on #4096

PR #4098, opened 2019-07-30
[stg]add monitoring script for miner program

bmeng:
Check the miner programs running on the pro clusters.

It filters the pod by name, since it is the most efficient way that we do not need to go through the pods/projects one by one.
And it supports two pods name by regex for now.

And we can modify the cron job parameter to change the match strings if the pattern changed.


bmeng:
Updated to use a list of pod names.

BlueShells:
👍

PR #4091, opened 2019-07-28
Added ARN ignore functionality that reads from

dofinn:
/etc/openshift_tools/v4_arn_ignore.yaml

Due to V4 instances becoming available for access, the current aws_api_key_manager script falls over when it can not handled failed auth to these ARNs. Until  V4 access has a SOP, I propose we pass a config file that holds V4 ARNS that aws_api_key_manager will parse and ignore.

PR #4099, opened 2019-07-31
Added ARN ignore functionality that reads from

dofinn:
/etc/openshift_tools/v4_arn_ignore.yaml


dofinn:
@rhdedgar :+1: #4091

RiRa12621:
seems like a duplicate of https://github.com/openshift/openshift-tools/pull/4091
Should be closed if that's the case.

dofinn:
> seems like a duplicate of #4091
> Should be closed if that's the case.

This is a PR into PROD, the #4091 was into STG

PR #4106, opened 2019-08-02
add daemonset check

BlueShells:


scripts/monitoring/cron-send-daemonset-status.py
```diff
@@ -0,0 +1,107 @@
...
+# In the future we might move this to a container where these
+# libs might exist
+#pylint: disable=import-error
+
+from openshift_tools.monitoring.ocutil import OCUtil
+from openshift_tools.monitoring.metric_sender import MetricSender
+
+ocutil = OCUtil()
+
+def runOCcmd(cmd, base_cmd='oc'):
```
bmeng: Looks like this is not been used, we can remove it.

scripts/monitoring/cron-send-daemonset-status.py
```diff
@@ -0,0 +1,107 @@
...
+
+    if args.verbose > 0:
+        logger.setLevel(logging.INFO)
+    if args.verbose > 1:
+        logger.setLevel(logging.DEBUG)
+
+    return args
+
+
+def test_configpod_pod(args=None, ):
```
bmeng: Cannot we just rely on the dameonset mechanism?
```
# oc get ds -n openshift-config 
NAME             DESIRED   CURRENT   READY     UP-TO-DATE   AVAILABLE   NODE SELECTOR     AGE
compute-config   7         7         7         7            7           config_pod=true   1d
```

How do we determine if the `--master-included` should be present?

BlueShells: when you add the cronjob , you already know which ds , then you already know if master included or not

scripts/monitoring/cron-send-daemonset-status.py
```diff
@@ -0,0 +1,107 @@
...
+
+def parse_args():
+    """ parse the args from the cli """
+    parser = argparse.ArgumentParser(description='configpod status check ')
+    parser.add_argument('-v', '--verbose', action='count', default=0,
+                        help='verbosity level, specify multiple')
+    parser.add_argument('--namespace', default="openshift-config", help='service namespace')
+    parser.add_argument('--ds', default="compute-config", help='service namespace')
+    parser.add_argument('-ma', '--master_included', action='store_true', default=None, help='if master included')
+    parser.add_argument('-key', default="openshift.master.public.ssl.left", help='zabbix key')
```
drewandersonnz: this `openshift.master.public.ssl.left` key should not be here.
maybe remove the default, and make the `-key` required.

scripts/monitoring/cron-send-daemonset-status.py
```diff
@@ -0,0 +1,107 @@
...
+    """ log commands through ocutil """
+    logger.info(base_cmd + " " + cmd)
+    ocy_time = time.time()
+    ocy_result = ocutil.run_user_cmd_yaml(cmd, base_cmd=base_cmd, )
+    logger.info("oc command took %s seconds", str(time.time() - ocy_time))
+    return ocy_result
+
+def parse_args():
+    """ parse the args from the cli """
+    parser = argparse.ArgumentParser(description='configpod status check ')
```
drewandersonnz: is this a "configpod" check?

drewandersonnz (approved):

PR #4109, opened 2019-08-05
add daemonset check script into rpm

BlueShells:


BlueShells:
@drewandersonnz has +1 #4108

PR #4110, opened 2019-08-05
add cronjob for fluentd and configpod

BlueShells:


BlueShells:
@wanghaoran1988 has +1 #4111

PR #4114, opened 2019-08-06
add try and catch for script

BlueShells:


scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -344,11 +344,18 @@ def run(self):
...
-        logging_status['kibana'] = self.check_kibana()
+        try:
+            logging_status['elasticsearch'] = self.check_elasticsearch()
+            logging_status['fluentd'] = self.check_fluentd()
+            logging_status['kibana'] = self.check_kibana()
+            self.report_to_zabbix(logging_status)
+        except:
+            result = 0 #something wrong with the checking script
+            mts = MetricSender(verbose=self.args.verbose)
+            mts.add_metric({'openshift.master.logging.elasticsearch.script.status': result})
```
wanghaoran1988: new metrics item? 0 stands for bad ? if there is no exception, should we send the metrics ? and what's the value?

BlueShells: its dynamic ones , too complex , its far more easy to create a new one

scripts/monitoring/cron-send-logging-checks.py
```diff
@@ -344,11 +344,18 @@ def run(self):
...
-        logging_status['elasticsearch'] = self.check_elasticsearch()
-        logging_status['fluentd'] = self.check_fluentd()
-        logging_status['kibana'] = self.check_kibana()
+        try:
+            logging_status['elasticsearch'] = self.check_elasticsearch()
+            logging_status['fluentd'] = self.check_fluentd()
+            logging_status['kibana'] = self.check_kibana()
+            self.report_to_zabbix(logging_status)
+        except:
+            result = 0 #something wrong with the checking script
```
drewandersonnz: how does this get set to something which `!= 0` ?? If this script errors, there is no "cleared" state.

drewandersonnz (changes requested):
-1, there is no clear state in the script. It will error on a fault, but will not remove the error when the fault clears.

drewandersonnz (approved):

PR #4116, opened 2019-08-06
oso-zabbix-server: Support RHEL7 only

mbarnes:
Discontinue generating unused CentOS7 images.

Make the generated RHEL7 code canonical and drop the templates.

This allows the `oso-rhel7-zabbix-server` BuildConfig to point to a repo directory without keeping generated code under source control.


blrm:
:+1: this looks good to me.

blrm (approved):

PR #4121, opened 2019-08-06
oso-zabbix-web: Support RHEL7 only

mbarnes:
Discontinue generating unused CentOS7 images.

Make the generated RHEL7 code canonical and drop the templates.

This allows the `oso-rhel7-zabbix-web` BuildConfig to point to a repo directory without keeping generated code under source control.


mbarnes:
:+1: @blrm #4120

PR #4123, opened 2019-08-07
[stg] skip the terminating pods which have finalizer set

bmeng:
Skip the pods who have the `finalizer` set. (Assume the terminating here is intentional)


wanghaoran1988:
+1

PR #4128, opened 2019-08-07
oso-zabbix-server: Add ops-rpm.repo

mbarnes:
For installing `openshift-tools-scripts-monitoring-zabbix-heal`.

Builds occur on compute nodes and compute nodes do not have all the necessary yum repos in `/etc/pki/entitlement` to mount into containers' `/run/secrets/etc-pki-entitlements`.


mbarnes:
:+1: @blrm #4128

PR #4131, opened 2019-08-08
add high alert for logging ops-runner

BlueShells:


BlueShells:
@wanghaoran1988 has +1 #4130

PR #4135, opened 2019-08-08
Add snapshot tag check to rpm and crontab

wanghaoran1988:


ansible/roles/oso_host_monitoring/templates/monitoring-config.yml.j2
```diff
@@ -287,6 +287,10 @@ host_monitoring_cron:
 - name: run create app with build process
   minute: "*/30"
   job: "ops-runner -f -t 1300 -s 120 -n csca.openshift.master.app.build.create /usr/bin/cron-send-create-app -v --basename build --loopcount 180 --source https://github.com/openshift/nodejs-ex &>> /var/log/build_app.log"
+
+- name: check volumes snapshot tags
+  minute: "*/15"
```
drewandersonnz: Does it have to be 15 mins? Can it be longer? I'm thinking of the load on the zabbix infrastructure. 
In theory, these alerts should not affect the daily running of the cluster, and are just checking if we have backups correctly.

BlueShells:
👍

PR #4152, opened 2019-08-21
remove logging check from reg-aws for this cluster didn't have logging

BlueShells:

PR #4153, opened 2019-08-21
remove logging check from reg-aws for this cluster didn't have logging

BlueShells:


BlueShells:
@wanghaoran1988 has +1 #4152

PR #4180, opened 2019-08-27
Add support for monitoring operator or script

dak1n1:
Allow clusters to run either the dedicated-admin-operator or the dedicated-admin script.
This will be helpful during the migration phase, as some clusters will be running the old
script and others will be running the operator. Both can be monitored now.


yithian (approved):

mbarnes (approved):
Looks ok to me.

PR #4190, opened 2019-08-28
Enable monitoring container to see cluster tier

dak1n1:
This change passes an environment variable to the monitoring container,
which will allow us to configure monitoring based on the cluster tier.
For example, only monitor dedicated-admin service on dedicated tier.

Related PR: openshift/openshift-ansible-ops#5617


dak1n1:
Approved in PR #4189

PR #4194, opened 2019-08-29
Specify namespace for zagg service

dak1n1:
This should fix an error encountered when the service is not found.


dak1n1:
Approved in PR #4193

PR #4196, opened 2019-08-30
Separate out RUN commands for easier debugging

dak1n1:
Having too many &&'s chaining together commands made the docker build difficult to debug.
I separated some of the yum commands out, and removed pcp before trying to install a new version.
This has lead to more reliable builds of the monitoring container in my testing.

https://jira.coreos.com/browse/SREP-2026


BlueShells:
👍

PR #4202, opened 2019-09-05
added createvpcendpoint and describevpcendpoint to the byoc customer admin policy

mwoodson:


mwoodson:
@jewzaam approved #4201

PR #4203, opened 2019-09-09
[stg]skip the teminating pod check on free-online clusters

bmeng:
Move the terminating pod into the if condition where it runs on dedicated clusters only.


wanghaoran1988:
+1

PR #4204, opened 2019-09-09
[prod] remove the check for the un-used path/cert for internal ssl

bmeng:
prod branch change for https://github.com/openshift/openshift-tools/pull/4198


bmeng:
Merge this as https://github.com/openshift/openshift-tools/pull/4198 has been `+1`ed

PR #4212, opened 2019-09-19
added the openshift registry ip cert role

mwoodson:


mwoodson:
@yithian approved #4211

PR #4215, opened 2019-09-20
fixed the registry ip docker cert dir role

mwoodson:


mwoodson:
@robotmaxtron  approved #4214

PR #4220, opened 2019-09-23
added 2 more guardduty rules to the customer admin role

mwoodson:


mwoodson:
@jewzaam approved #4219

PR #4225, opened 2019-09-26
updated the ssl check to accept the hosts from being passed into the role

mwoodson:


mwoodson:
@yithian approved #4224

PR #4228, opened 2019-10-02
force sync attempt

drewandersonnz:
still record errors in zabbix item
sync even on error
we have problems today where teams can't use new clusters because sync
is failing


dak1n1 (approved):
:+1:

PR #4230, opened 2019-10-07
aws_api_key_manager.py: Catch botocore.exceptions.ClientError

mbarnes:
Temporarily works around an AWS account issue, but more robust error handling is a good thing regardless.

### DIFF
--- a/scripts/iam-tools/aws_api_key_manager.py
+++ b/scripts/iam-tools/aws_api_key_manager.py
@@ -266,6 +266,10 @@ class ManageKeys(object):
                         )
                     return client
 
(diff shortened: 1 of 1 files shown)


lisa (approved):
/lgtm

It would be nice to understand why these errors are happening and fix them, but being able to get through the list in the mean time is 👍

jharrington22:
@mbarnes @lisa Just a comment here but we've seen eventual consistency issues with the AWS API and credentials. Sometimes they are not ready if you call the API immediately after creating them (we actually saw it so much we added a wait). It doesn't look like thats where you are getting the error her though, looks like you are getting it on credential creation? Just thought i'd mention this.

PR #4231, opened 2019-10-07
aws_api_key_manager.py: Catch botocore.exceptions.ClientError

mbarnes:
Temporarily works around an AWS account issue, but more robust error handling is a good thing regardless.


mbarnes:
:+1: @lisa #4230

PR #4235, opened 2019-10-09
added the ops gpg signing key

mwoodson:


mwoodson:
@mbarnes approved #4234

PR #4241, opened 2019-10-10
[prod] add zabbix key and trigger for new check

bmeng:
prod PR for https://github.com/openshift/openshift-tools/pull/4240


bmeng:
merge this as #4240 approved

PR #4243, opened 2019-10-14
Update README.md

NautiluX:
use code blocks for example config


RiRa12621:
merge it, we'll fix the tests separately as they don't seem to be related @NautiluX

PR #4244, opened 2019-10-22
Minor changes to create-app check

dak1n1:
This commit contains the following changes:
* Test http response of route instead of service IP.
* Test project creation using the project-request template.

https://jira.coreos.com/browse/SREP-2277
https://jira.coreos.com/browse/SREP-1859

### DIFF
--- a/ansible/roles/os_zabbix/vars/template_openshift_master.yml
+++ b/ansible/roles/os_zabbix/vars/template_openshift_master.yml
@@ -130,6 +130,18 @@ g_template_openshift_master:
     applications:
     - Openshift Master
 
+  - key: openshift.master.app.create.service_http_failed
+    description: "Non-200 HTTP status received from service during app-create check"
+    value_type: int
+    applications:
+    - Openshift Master
+
+  - key: openshift.master.app.create.route_http_failed
+    description: "Non-200 HTTP status received from route during app-create check"
+    value_type: int
+    applications:
+    - Openshift Master
+
   - key: openshift.master.dnsmasq.curl.status
     description: "check status of dnsmasq use curl"
     value_type: int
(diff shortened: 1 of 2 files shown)


scripts/monitoring/cron-send-create-app.py
```diff
@@ -86,11 +86,13 @@ def send_metrics(build_ran, create_app, http_code, run_time):
 
     if build_ran == 1:
         ms.add_metric({'openshift.master.app.build.create': create_app})
-        ms.add_metric({'openshift.master.app.build.create.code': http_code})
+        ms.add_metric({'openshift.master.app.create.route_http_failed': route_http_failed})
```
drewandersonnz, line 87: these should be `openshift.master.app.build.` as this is in the build section. Though this whole if could be different, but further refactoring is not needed for this PR.

scripts/monitoring/cron-send-create-app.py
```diff
@@ -167,9 +169,12 @@ def setup(config):
     except Exception:
         pass # don't want exception if project not found
 
+    # Create a new project using 'oc' instead of 'oc adm'.
+    # This will test the project-request template as part of project creation.
```
drewandersonnz, line 171: This will only test `project-request template` on the first time creation, as it does not remove the project at the end.

ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -1086,6 +1112,22 @@ g_template_openshift_master:
     description: The application create loop has failed 4 or more times in the last hour
     priority: high
 
+  - name: "Route has returned non-200 HTTP code multiple times in the last hour on {HOST.NAME}"
+    expression: "{Template Openshift Master:openshift.master.app.create.route_http_failed.sum(1h)}>3"
+    url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/check_create_app.asciidoc"
+    dependencies:
```
drewandersonnz, line 1118: Is the dependency needed here?

ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -1086,6 +1112,22 @@ g_template_openshift_master:
...
+    - "Openshift Master process not running on {HOST.NAME}"
+    description: The app-create check has encountered 4 or more route curl failures in the last hour
+    priority: high
+
+  - name: "Service IP has returned non-200 HTTP code multiple times in the last hour on {HOST.NAME}"
+    expression: "{Template Openshift Master:openshift.master.app.create.service_http_failed.sum(1h)}>3"
+    url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/check_create_app.asciidoc"
+    dependencies:
```
drewandersonnz, line 1126: same dependency question as above

ansible/roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -1078,6 +1090,20 @@ g_template_openshift_master:
     - "Openshift Master process not running on {HOST.NAME}"
     priority: avg
 
+  - name: "Route returned non-200 HTTP status on {HOST.NAME}"
```
drewandersonnz, line 1093: Thinking this should be standardised messaging as per the above: 
"App Create returned non-200 HTTP status on ROUTE"

scripts/monitoring/cron-send-create-app.py
drewandersonnz: We cannot remove the project as it is possible multiple tests are ongoing (app create and app build) and removing the project will interfere with the other execution.

drewandersonnz (changes requested):
Please see comments below.

scripts/monitoring/cron-send-create-app.py
dak1n1: I had considered that, but I chose to keep a single Zabbix item for each of the http failure types. There are a couple reasons to keep a single item instead of separating them out into "http failure that happened during build" vs "http failure that happened during a normal app-create":

* We would have 2 counters going for the a single issue, so we wouldn't alert quickly enough. The builds only run every 30 minutes, and each run times out after 9 minutes if a URL is unreachable. So it would take 2 hours to notice route failures and alert us. (Similar to how we alert for build failures today). IMO, the router potentially being offline needs a quicker response than a build failure does.

* An http failure encountered during app-create wouldn't increment the build counter, because it has its own counter.

* We would receive two alerts for the same issue, if the issue continued on for 2 hours. Which would be unnecessary noise.

* Adding an additional two items for each cluster to keep track of means more load on the zabbix server. I would want to justify any additional load being added to the Zabbix server, since it already carries a heavy load.

So by making just one counter (zabbix item) for this, it gets incremented faster and we can tell when something has gone wrong with a router faster.

scripts/monitoring/cron-send-create-app.py
dak1n1: Good catch. I thought it would be ok to just test the app itself, but you're right, we need a new project for each app-create run. This seems doable. I can work on it.

ansible/roles/os_zabbix/vars/template_openshift_master.yml
dak1n1: I copied the dependency from the app-create check, because if the master API goes offline, that means app creation will fail. And without an app to route the http traffic to, all curl requests are going to fail.

ansible/roles/os_zabbix/vars/template_openshift_master.yml
dak1n1: I think it's needed, yes.

scripts/monitoring/cron-send-create-app.py
dak1n1: Done. It creates a project each time now.

dak1n1:
I tested unique project creation and project/pod name lengths. I don't think there's much room for naming collisions, given that the datestamp now includes seconds. And there is a unique ID for each namespace (in the event that multiple namespaces are created in the same second).

```
[root@ded-stage-aws-master-ip-172-31-2-40.ec2.internal ~]# oc get pods --all-namespaces |grep sre-app
sre-app-check-20191025-235425-hj   build-20191025-235425-hj-1-5kxh5                 1/1       Running     0          3m
sre-app-check-20191025-235745-uk   pull-20191025-235745-uk-1-vxt7x                  1/1       Running     0          6s
```

drewandersonnz (approved):
+1 from me based on our conversation. 
Future additions maybe? 
* add sre-app-check-* projects to dedicated-admins blacklist so customer cannot see it
* need to monitor or cleanup projects that are too old, in the case the script fails before cleaning up

PR #4245, opened 2019-10-22
Minor changes to create-app check

dak1n1:
This commit contains the following changes:
* Test http response of route instead of service IP.
* Test project creation using the project-request template.

https://jira.coreos.com/browse/SREP-2277
https://jira.coreos.com/browse/SREP-1859

### DIFF
--- a/ansible/roles/os_zabbix/vars/template_openshift_master.yml
+++ b/ansible/roles/os_zabbix/vars/template_openshift_master.yml
@@ -130,6 +130,18 @@ g_template_openshift_master:
     applications:
     - Openshift Master
 
+  - key: openshift.master.app.create.service_http_failed
+    description: "Non-200 HTTP status received from service during app-create check"
+    value_type: int
+    applications:
+    - Openshift Master
+
+  - key: openshift.master.app.create.route_http_failed
+    description: "Non-200 HTTP status received from route during app-create check"
+    value_type: int
+    applications:
+    - Openshift Master
+
   - key: openshift.master.dnsmasq.curl.status
     description: "check status of dnsmasq use curl"
     value_type: int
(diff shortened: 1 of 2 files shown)


drewandersonnz:
Is it possible to test both Service and Route in the same code? 
Double the curl step and check both service and route separately, both must return HTTP 200.

This is because testing a Route no longer uses the Service Network so we are losing a test in this script.

dak1n1:
Added service check back. It reports http status of both route and service now.

```
2019-10-22 18:43:55,024 -  17004 - DEBUG    - curl(http://pull-10221843z-p0-ops-health-monitoring.5c20.ded-stage-aws.openshiftapps.com:80 timeout=30s)
2019-10-22 18:43:55,053 -  17034 - DEBUG    - route http code 200
2019-10-22 18:43:55,053 -  17034 - DEBUG    - route and service curl completed in 0 tries
[...]
2019-10-22 18:43:55,054 -  17034 - INFO     - Test finished. Time to complete test only: 16.0433068275
2019-10-22 18:43:55,054 -  17034 - DEBUG    - send_metrics()
2019-10-22 18:43:55,067 -  17048 - INFO     - Send data to MetricSender
2019-10-22 18:43:55,069 -  17050 - DEBUG    - Starting new HTTP connection (1): 172.30.241.134
2019-10-22 18:43:55,075 -  17056 - DEBUG    - http://172.30.241.134:80 "POST /metric HTTP/1.1" 200 21
2019-10-22 18:43:55,076 -  17056 - INFO     - Data sent to Zagg in 0.0220420360565 seconds
2019-10-22 18:43:55,076 -  17056 - INFO     - Deploy State: Success
2019-10-22 18:43:55,076 -  17057 - INFO     - Service HTTP response code: 200
2019-10-22 18:43:55,076 -  17057 - INFO     - Route HTTP response code: 200
```

mwoodson:
This code looks great!  Thanks for adding this in.

I have a concern about the reporting. IF I am reading through the code right, I don't believe we are actually alerting if the service or the route are not working, only if the app create fails.  It looks like we are sending this data to Zabbix, but Zabbix doesn'thave an item setup.

It looks like the data we are collecting from zabbix is found here:

https://github.com/openshift/openshift-tools/blob/prod/ansible/roles/os_zabbix/vars/template_openshift_master.yml#L122-L131

and the trigger is here:

https://github.com/openshift/openshift-tools/blob/prod/ansible/roles/os_zabbix/vars/template_openshift_master.yml#L1066-L1083

this includes the app.create and the app.build.create.

I believe the file here is great.  

I think the next steps are to add the items to zabbix, and a check on those items.

We need to add the items for:

 openshift.master.app.build.create.route_http_code
openshift.master.app.build.create.service_http_code

Then add triggers that alert on these.
￼

dak1n1:
Ok, I've created data points and alerts in zabbix, and decoupled the http status alert from the app-create alert. I think this is the functionality you were looking for. Can I get another review, @mwoodson, @drewandersonnz or @luis-falcon ? Whoever has time. Thanks!

Here are test results from curl failures:

```
2019-10-24 20:33:27,528 - 144616 - INFO     - Deploy State: Success
2019-10-24 20:33:27,528 - 144616 - INFO     - Service HTTP response code: 403
2019-10-24 20:33:27,528 - 144616 - INFO     - Route HTTP response code: 0
2019-10-24 20:33:27,528 - 144616 - INFO     - teardown()
```

By having the http status check decoupled from the app-create check, it keeps us from getting both an app-create alert and an http status alert (getting alerted twice for the same issue). The app-create alert only goes off if the pod fails to enter Running state.

dak1n1:
I tested this out on a cluster with 2 routers. It turns out it works fine without the `route=external` label. (That was something I thought it needed, but turns out it's fine as is.) So this is ready for review.

dak1n1:
approved in PR #4244

PR #4246, opened 2019-10-31
fixed the route 200 zabbix trigger

mwoodson:

PR #4247, opened 2019-10-31
fixed the route 200 zabbix trigger

mwoodson:


mwoodson:
@rogbas approved #4246

PR #4253, opened 2019-11-05
Revert to previous verion until ready to release

dak1n1:
Due to the way we build rpms for the monitoring container,
these changes had to be merged to prod before a stage container could be built.
Now it needs to be reverted back to the previous state, so that the staged changes
aren't accidentally pushed to prod by another monitoring container build.

This places the app-create script at commit # 0e983f9fcfe9b40996c84e6ee975b52739ee4a11.


drewandersonnz (approved):
+1 due to revert

PR #4250, opened 2019-11-05
SREP-2260 monitoring script for node health behind each available ELB

dofinn:

PR #4255, opened 2019-11-06
SREP 2260 aws elb health

dofinn:


dofinn:
+1 from: https://github.com/openshift/openshift-tools/pull/4250

PR #4259, opened 2019-11-08
Shrink monitoring container size

dak1n1:
Our monitoring container size just hit 3.5G, which is larger than
our default docker dm.basesize of 3G. Our hosts won't be able to pull it
until either the dm.basesize is increased, or the container size is shrunk.
This commit shrinks the container size by having fewer layers generated
by the RUN commands.


drewandersonnz (approved):

drewandersonnz:
`This commit shrinks the container size by having fewer layers generated
by the RUN commands.`

This commit also runs `yum clean all` at the end of each `yum install` layer, so it doens't capture old RPM repo data in the yum cache.

PR #4265, opened 2019-11-12
Added new elb health script to rpm spec so its build with monitoring …

dofinn:
…container


dofinn:
+1 from: https://github.com/openshift/openshift-tools/pull/4262

PR #4268, opened 2019-11-14
Add script for fluentd queue depth check

ninabauer:


scripts/monitoring/cron-send-fluentd-depth.py
```diff
@@ -0,0 +1,122 @@
...
+    def check_fluentd_queues(self):
+        """ Check oldest buffer file in Fluentd pods """
+        # Get timestamps of files in /var/lib/fluentd from each pod
+        buffer_list = []
+        for pod in self.fluentd_pods:
+            pod_name = pod['metadata']['name']
+            find_ts = "exec " + pod_name + " -- find /var/lib/fluentd -type f -name \*.log ! -name '*output_ops_tag*' -printf '%T+\n'"
+            buffer_ts = self.oc.run_user_cmd(find_ts)
+            timestamps = buffer_ts.split("\n")
+            timestamps.pop()
```
yithian: Does this unconditionally remove the last timestamp in `timestamps`? Is there a reason for that? I would expect we would want all the values to be kept but I might be missing something.

yithian (approved):

PR #4272, opened 2019-11-14
oc_storageclass: Drop "beta" from storage class annotation

mbarnes:
To synchronize with https://github.com/openshift/openshift-ansible/commit/18b42ee


mbarnes:
:+1: @dak1n1 #4271

PR #4277, opened 2019-11-22
resolved path to aws creds, added cron for triggering

dofinn:


dofinn:
+1 as per: https://github.com/openshift/openshift-tools/pull/4267

dofinn:
lgtm as per: https://github.com/openshift/openshift-tools/pull/4267

PR #4285, opened 2019-12-12
added the openshift_clusterrole role

mwoodson:


mwoodson:
@jharrington22 approved #4284

PR #4287, opened 2020-01-06
-updated teh elb health check to check last 3 values instead of just one value

mwoodson:


ansible/roles/os_zabbix/vars/template_aws.yml
```diff
@@ -65,7 +65,7 @@ g_template_aws:
     url: "https://github.com/openshift/ops-sop/blob/master/v3/alerts/check_aws_instance.asciidoc"
     priority: average
   - name: "Cluster ELBs have unhealthy instance/s checked from {HOST.NAME}"
-    expression: "{Template AWS:openshift.aws.elb.health.sum(15m)}>0"
+    expression: "{Template AWS:openshift.aws.elb.health.min(#3)}>0"
```
yithian, line 68: My reading of this expression is "if the minimum value for openshift.aws.elb.health is > 0 in any of the last three checks". Wouldn't that be pretty much what we already have (granted a 5 minute interval)?

Would we want something like this instead?
```suggestion
    expression: "{Template AWS:openshift.aws.elb.health.sum(15m)}>=3"
```
or
```suggestion
    expression: "{Template AWS:openshift.aws.elb.health.min(#3)}>=3"
```
?

yithian: based on https://www.zabbix.com/documentation/3.0/manual/appendix/triggers/functions

mwoodson: From the check:

0 = good
1 = bad

min(#3) would take the lowest of the 3. 
 
examples:

min(0,1,0) = 0 (don't fire)
min (1,1,0) = 0 (don't fire)

min (1,1,1) = 1, fire alert

The sum would work as well, but we have used this max, min (depending on what value is good and what is bad) pattern a lot in our checks, so I was sticking with it.

If this logic doesn't seem right, let me know.

mwoodson: how i read the current check:

if anything in the last 15 minutes is > 0, fire.  So there is a difference

yithian: I see, I was applying the order wrong in my head. Yup, this looks good then.

yithian (approved):

PR #4288, opened 2020-01-06
-updated teh elb health check to check last 3 values instead of just one value

mwoodson:


mwoodson:
@yithian approved #4287

PR #4304, opened 2020-02-12
Increase threshold for fluentd alert to 10 min

ninabauer:


ninabauer:
+1 by @RiRa12621 #4303

PR #4306, opened 2020-02-18
add_etcd_volume: Fix etcd_delegate_to host name determination

mbarnes:
Node names within a cluster changed in 3.11, making it a little more complicated to map a node name back to the Ansible inventory.


pyates86:
+1 @mbarnes

PR #4310, opened 2020-02-21
Fixed terminating pod check to handle oc 3.11.154+ and pods that have…

superfunhappytime:
… failed termination due to deadline exceeding

### DIFF
--- a/scripts/monitoring/cron-send-terminating-pod-check.py
+++ b/scripts/monitoring/cron-send-terminating-pod-check.py
@@ -91,20 +91,33 @@ def check_terminating_pod(pods):
 
     # Get the pod's terminated timestamp and compare with the current time
     # if the time difference greater than 10 minutes
-    # we thought the pod is stuck in terminating status for too long time
+    # the pod is stuck in terminating status for too long
     time_now = datetime.datetime.now()
     long_time_terminate_pod = []
     if pods:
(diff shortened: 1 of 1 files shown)


superfunhappytime:
[test]

scripts/monitoring/cron-send-terminating-pod-check.py
```diff
@@ -91,20 +91,33 @@ def check_terminating_pod(pods):
...
-                if item['type'] == "Ready":
-                    time_terminated = item['lastTransitionTime']
+            if 'conditions' in yaml_result['status']:
+                status_condition = yaml_result['status']['conditions']
+                for item in status_condition:
+                    if item['type'] == "Ready":
+                        time_terminated = item['lastTransitionTime']
+            elif ('phase' in yaml_result['status'] and
+                  yaml_result['status']['phase'] == "Failed" and
+                  yaml_result['status']['reason'] == "DeadlineExceeded"):
```
mrbarge, line 111: If reason isn't in the dict it'll throw a keyerror. Looks like the rest of the new code is very thorough in checking dict keys, so may as well add a check for this one too?

superfunhappytime: Good catch, missed that one.

mrbarge (approved):

PR #4323, opened 2020-04-14
Fix cron-event-watcher monitor OSD-3225

mrbarge:
This is the prod branch equivalent of stg PR #4322


mrbarge:
@dofinn  :+1: #4322

PR #4328, opened 2020-05-11
created the role openshift_project

mwoodson:


ansible/roles/openshift_project/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+osproj_display_name: "{{ osproj_name }}"
+osproj_description: "{{ osproj_display_name }}"
```
csheremeta: should this be `osproj_description` instead?

Issue #4349, opened 2020-08-19
Node service restart caused by Sync Pod startup


jianlongzhou:
What happened:
In our environment, the master node failed due to insufficient system resources, and most of the sync pods unexpectedly restarted during this period, which caused the node service on the computing node to restart, which in turn caused the business Pod on the node to have a MatchNodeSelector phenomenon (as https://github.com/kubernetes/kubernetes/issues/52902 as described).

What you expected to happen:
We hope that sync pod will not cause the atomic-node service on the compute node to restart once when openshift-node/configmap does not change.

How to reproduce it (as minimally and precisely as possible):
Simply delete the sync pod and wait for the controller to regenerate.

Anything else we need to know?:

Environment:

ocp version: v3.11.43
OS (e.g: cat /etc/os-release): RHEL 7.8
Kernel (e.g. uname -a): 3.10.0-1127.13.1.el7.x86_64

/kind bug

jianlongzhou:
We found that when the oc annotate operation fails, it will also cause the sync pod to restart. The set -e command in the shell script has this effect.

jianlongzhou:
Modifications have been made in the sync.yaml file in the newer openshift-ansible-roles rpm package.

PR #4350, opened 2020-08-19
FIX: Only match MD5 sequence value to avoid Node service restart when Sync Pod starts

jianlongzhou:
#4349 As described in the issue


jianlongzhou:
Modifications have been made in the sync.yaml file in the newer openshift-ansible-roles rpm package.

PR #4354, opened 2021-02-19
Modify heartbeat.ping, dnsmasq, and Service IP/Route non-200 alerts

csheremeta:
to reduce v3 noise - see [OSD-5913](https://issues.redhat.com/browse/OSD-5913)

PR #4355, opened 2021-02-23
Modify heartbeat.ping, dnsmasq, and Service IP/Route non-200 alerts

csheremeta:
to reduce v3 noise - see [OSD-5913](https://issues.redhat.com/browse/OSD-5913)


csheremeta:
@mmazur gave :+1: on #4354

PR #4359, opened 2021-03-01
Revert "Re-add 5/15 min Heartbeat alerts and set to disabled"

csheremeta:
This reverts commit 359c28904cc1e9e37557a974642f526176bdd472.

Reverting this because re-adding these alerts and setting to disabled
did not actually disable the alerts.


csheremeta:
@mmazur 👍🏻 #4358

PR #4361, opened 2021-06-23
OSD-7588 detect flapping node NotReady status

ninabauer:
STG PR #4360 

Adds trigger to detect if a node is flapping NotReady status


ninabauer:
lgtm from @sam-nguyen7 #4360

PR #4369, opened 2021-11-22
adding username and password for basic authentitacion to the yum_repos task

mwoodson:
new mirror servers supporting basic auth, not gpg keys. this adds the ability to add username/password.


blrm:
👍🏻 lgtm!

PR #4371, opened 2022-03-22
decom of unused v2 zabbix srv OSD-5865

rhdedgar:
This removes the alert trigger that checks for the online status of the V2 Zabbix server. The V2 server and proxy instances are no longer in use, and have been stopped.

### DIFF
--- a/ansible/roles/os_zabbix/vars/template_app_zabbix_server.yml
+++ b/ansible/roles/os_zabbix/vars/template_app_zabbix_server.yml
@@ -349,12 +349,6 @@ g_template_app_zabbix_server:
        'no': 2
 
    ztriggers:
-   - description: "Zabbix v2 Web UI is not responding"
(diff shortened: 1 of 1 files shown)


ansible/roles/os_zabbix/vars/template_app_zabbix_server.yml
```diff
@@ -303,21 +303,6 @@ g_template_app_zabbix_server:
      zabbix_type: trapper
 
    zhttptests:
```
dustman9000, line 305: I think this needs to be a []

dustman9000: Nevermind, I thought this was the only element of the list being deleted. The newlines were throwing me off.

dustman9000:
Looks ok to me. I'm not sure why the Jenkins PR check is failing though.

rhdedgar:
When I checked the build output yesterday, it was failing to initialize a build due to a cert that expired in 2021.

